PR #7, opened 2020-08-09
Refactor imports

diptorupd:
reorganize imports using PyDev's organize imports feature.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,12 +21,15 @@
 ### \file
 ### This file builds the dppl and dppl.ocldrv extension modules.
 ##===----------------------------------------------------------------------===##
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -21,12 +21,15 @@
...
 ##===----------------------------------------------------------------------===##
-import sys
-import numpy as np
 import os
+import sys
 import versioneer
-from setuptools import setup, Extension
+
 from Cython.Build import cythonize
+from setuptools import setup, Extension
```
PokhodenkoSA, line 29: I did not catch this. `setuptools` should be imported before Cython.

PR #9, opened 2020-08-10
Make dppl rt threadlocal

diptorupd:
Stores the sycl device queues in a thread local storage. The DpplRuntime object is no longer exposed, but new global prebound functions exposes the same API.

### DIFF
--- a/dppl/oneapi_interface.pyx
+++ b/dppl/oneapi_interface.pyx
@@ -59,6 +59,7 @@ cdef extern from "dppl_oneapi_interface.hpp" namespace "dppl":
         int64_t activateQueue (void **Q, _device_type DTy,
                                size_t device_num) except -1
         int64_t deactivateCurrentQueue () except -1
+        int64_t number_of_activated_queues (size_t &num) except -1
         int64_t dump () except -1
         int64_t dump_queue (const void *Q) except -1
 
@@ -103,7 +104,7 @@ cdef class _DpplRuntime:
         ''' Prints information about the SYCL queue object.
         '''
         if PyCapsule_IsValid(queue_cap, NULL):
-            self.rt.dump_queue(PyCapsule_GetPointer(queue_cap, NULL))
+            return self.rt.dump_queue(PyCapsule_GetPointer(queue_cap, NULL))
         else:
             raise ValueError("Expected a PyCapsule encapsulating a SYCL queue")
 
@@ -130,13 +131,21 @@ cdef class _DpplRuntime:
             e = UnsupportedDeviceTypeError("Device can only be cpu or gpu")
             raise e
 
+    def is_in_dppl_ctxt (self):
+        cdef size_t num = 0
+        self.rt.number_of_activated_queues(num)
+        if num:
+            return True
+        else:
+            return False
+
+
 # thread-local storage
 from threading import local as threading_local
 
 # Initialize a thread local instance of _DpplRuntime
 _tls = threading_local()
-_tls._in_dppl_ctxt = False
-_tls._runtime      = _DpplRuntime()
+_tls._runtime = _DpplRuntime()
 
 
 ################################################################################
@@ -144,16 +153,12 @@ _tls._runtime      = _DpplRuntime()
 ################################################################################
 
 
-def is_in_dppl_ctxt ():
-    return _tls._in_dppl_ctxt
-
-
 dump              = _tls._runtime.dump
 dump_queue_info   = _tls._runtime.dump_queue_info
 get_current_queue = _tls._runtime.get_current_queue
 get_num_platforms = _tls._runtime.get_num_platforms
 set_default_queue = _tls._runtime.set_default_queue
-
+is_in_dppl_ctxt   = _tls._runtime.is_in_dppl_ctxt
 
 from contextlib import contextmanager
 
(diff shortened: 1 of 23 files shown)


diptorupd:
Also contains the changes from PR #7

setup.py
```diff
@@ -21,12 +21,15 @@
...
 ##===----------------------------------------------------------------------===##
-import sys
-import numpy as np
 import os
+import sys
 import versioneer
-from setuptools import setup, Extension
+
 from Cython.Build import cythonize
+from setuptools import setup, Extension
```
PokhodenkoSA, line 29: `setuptools` should be imported before Cython.
In https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#configuring-the-c-build:
> Note that when using setuptools, you should import it before Cython, otherwise, both might disagree about the class to use here.

diptorupd: Good catch. I will fix it.

diptorupd: Fixed.

dppl/oneapi_interface.pyx
```diff
@@ -101,37 +94,66 @@ cdef class DpplRuntime:
...
+        else:
+            e = UnsupportedDeviceTypeError("Device can only be cpu or gpu")
+            raise e
+
+# thread-local storage
+from threading import local as threading_local
+
+# Initialize a thread local instance of _DpplRuntime
+_tls = threading_local()
+_tls._in_dppl_ctxt = False
```
DrTodd13, line 138: Not sure we ever discussed this scenario.  However, I think I agree that in a new thread, you go back to the default of not being in a context.

dppl/oneapi_interface.pyx
```diff
@@ -146,12 +168,13 @@ def device_context (dev=device_type.gpu, device_num=0):
...
+        _tls._in_dppl_ctxt = True
+        ctxt = _tls._runtime._activate_queue(dev, device_num)
         yield ctxt
     finally:
         # Code to release resource
         if ctxt:
-            print("Removing the context from the deque of active contexts")
-            runtime._deactivate_current_queue()
+            _tls._runtime._deactivate_current_queue()
+            _tls._in_dppl_ctxt = False
```
DrTodd13, line 178: Shouldn't this be a virtual stack of contexts?  I would think that on "finally" that you pop the top off the stack and then only set in_dppl_ctxt to false if the stack is empty.  Even better, you don't have this in_dppl_ctxt at all and just check the size of the stack if you need to know if you're in a context or not.

diptorupd: The stack of contexts is maintained inside the _runtime.rt object. I am using a std::deque to store the list of currently activated queues. The first entry of the deque is always the global or default queue. Using an example:

```
default_q = dppl.get_current_queue()

with dppl.device_context(dppl.device_type.cpu, 0) as cpu_q:
    ...
    with dppl.device_context(dppl.device_type.gpu, 0) as gpu_0_q:
        ...
    with dppl.device_context(dppl.device_type.gpu, 1) as gpu_1_q:
        ...
```
Here the stack will be maintained inside dppl._tls._runtime as:

`default_q` (initial state)
`cpu_q --> default_q` (on entering first with context)
`gpu_0_q --> cpu_q --> default_q` (on entering the first nested with context)
`cpu_q --> default_q` (on exiting the first nested with context)
`gpu_1_q --> cpu_q --> default_q` (on entering the second nested with context)
`cpu_q --> default_q` (on exiting the second nested with context)
`default_q` (on exiting the top-level with context)

The `_deactivate_current_queue()` call does the pop from the `deque. dppl.get_current_queue` always returns the current top of stack queue.

About your second point about ` in_dppl_ctxt`, yes the way it is currently implemented it returns wrong result when contexts are nested. Let me fix it to return the size of the deque -1 (since the default queue is always going to be inside the deque).

dppl/oneapi_interface.pyx
diptorupd: I am removing the flag. Instead of the flag let me add a public getter that returns True/False based on if we are inside *any* context by checking the size of the stack of contexts.

dppl/oneapi_interface.pyx
diptorupd: Made the change in 8318d44

dppl/oneapi_interface.pyx
diptorupd: Made the change in 8318d44

dppl/oneapi_interface.pyx
```diff
@@ -101,37 +95,70 @@ cdef class DpplRuntime:
...
+#--------------------------------- Public API ---------------------------------#
+################################################################################
+
+
+dump              = _tls._runtime.dump
+dump_queue_info   = _tls._runtime.dump_queue_info
+get_current_queue = _tls._runtime.get_current_queue
+get_num_platforms = _tls._runtime.get_num_platforms
+set_default_queue = _tls._runtime.set_default_queue
+is_in_dppl_ctxt   = _tls._runtime.is_in_dppl_ctxt
```
fschlimb, line 161: Why is the entire runtime thread-local? The only thing which is thread-local is the stack of contexts. Making shared data thread-local is calling for trouble.
I think the runtime class should encapsulate the thread-local business and return the thread-local values in its member functions like get_current_queue. Maybe this needs to be done in C++ (e.g. your std::deque be thread-local).

PR #20, opened 2020-08-14
Refactor oneapi wrapper v2

diptorupd:
Refactored several things to drop the use of oneAPI in favour of sycl.


PokhodenkoSA:
@diptorupd If you make it not draft we could merge it.

PR #28, opened 2020-09-11
Correct use ERRORLEVEL in conda scripts for Windows

PokhodenkoSA:
When I build conda package on Windows in my local environment it fails on tests.
It happens because oneAPI env activation script (part related to FPGA) does `set ERRORLEVEL=1`.
As mentioned in this [article](https://devblogs.microsoft.com/oldnewthing/20080926-00/?p=20743) in cmd.exe scripts
`IF ERRORLEVEL 1 ...` checks the last exit value but `IF %ERROLEVEL% NEQ 0 ...` checks env variable. The env variable does not relate to last exit status. I think it is error in oneAPI and conda scripts to use env variable instead of exit value.

Current CI works without this modifications but my local env does not work without it.

Also, I have tested it with my last commit on branch beta/2021. When I test in master it crashes on python -c "import dppl". It does not relate to this modifications. Just there is an issue between beta/2021 and current master.

### DIFF
--- a/conda-recipe/bld.bat
+++ b/conda-recipe/bld.bat
@@ -1,4 +1,8 @@
 call "%ONEAPI_ROOT%compiler\latest\env\vars.bat"
+IF ERRORLEVEL 1 exit 1
+REM conda uses %ERRORLEVEL% but FPGA scripts can set it. So it should be reseted.
+set ERRORLEVEL=
+
 set "CC=dpcpp-cl.exe"
 set "CXX=dpcpp-cl.exe"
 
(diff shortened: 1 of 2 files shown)


PokhodenkoSA:
I have finished reading the article mentioned. If env var %ERRORLEVEL% is not set then access to it will fallback to exit value. So setting ERRORLEVEL in oneAPI scripts is error, but using it in conda scripts is not an error.

diptorupd:
I am fine with this change. Maybe open a ticket for oneAPI, but not a very high priority.

diptorupd (approved):

PokhodenkoSA:
I have returned old checks style `IF %ERRORLEVEL% NEQ 0 ...` because the main problem is in vars.bat setting this variable. After resetting the variable is could be used and it will behave as expected - fallback to exit value of previous command.

PR #29, opened 2020-09-14
Change static variables to static pointers

PokhodenkoSA:
On Windows static variables construction and destruction cause crashes. I think it is because it should be called in specific order. With static variables the order is undefined.

Now the static variables are constructed at the first access to this variables. The first access happens after all necessary initializations.
Also I think destruction should happen in correct order. As we have no explicit destruction function it is hard to control where static global variables will be deleted. So the variables are changed to pointers and we do not call destruction for them at all. As they should exist to the end of the application it is ok to not destruct them and allow OS to clean its memory.

### DIFF
--- a/backends/source/dppl_sycl_queue_interface.cpp
+++ b/backends/source/dppl_sycl_queue_interface.cpp
@@ -105,17 +105,20 @@ class QMgrHelper
 public:
     static std::vector<cl::sycl::queue>&   cpu_queues_()
     {
-        static std::vector<cl::sycl::queue>* cpu_queues = QMgrHelper::init_queues(info::device_type::cpu);
+        static std::vector<cl::sycl::queue>* cpu_queues =
+            QMgrHelper::init_queues(info::device_type::cpu);
         return *cpu_queues;
     }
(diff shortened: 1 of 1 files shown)


backends/source/dppl_sycl_queue_interface.cpp
```diff
@@ -103,9 +103,24 @@ void error_reporter (const std::string & msg)
 class QMgrHelper
 {
 public:
-    static std::vector<cl::sycl::queue>   cpu_queues;
-    static std::vector<cl::sycl::queue>   gpu_queues;
-    static thread_local std::vector<cl::sycl::queue> active_queues;
+    static std::vector<cl::sycl::queue>&   cpu_queues_()
```
diptorupd, line 106: Minor nitpick.  Can you please add a space before the open paren? I usually ffollow the convention of adding a space before the open paren in function declaration. In the long term let us create a common coding style and formatter for VS Code so that everyone can be consistent.

diptorupd (approved):
Except minor formatting nitpick, looks good. Thanks for fixing it (once again).

PR #39, opened 2020-09-18
Change default current device to None

1e-to:


PokhodenkoSA:
@1e-to could you give a description for the changes please?

1e-to:
This PR related to [adding semantics](https://github.com/IntelPython/numba/pull/45) to Numba-dppl.
The current device value changed to None when we are not inside any `dppl.device_context`, and is only set the first time we enter the with scope.

diptorupd (approved):

diptorupd:
@1e-to can you review the new `dppl` API `dppl/sycl_core.pyx`? `dppl.get_current_queue` always returns a queue, but I have a `is_in_dppl_ctxt` function that checks if we are inside a with context. Will that work for the new dispatcher in Numba?

PR #34, opened 2020-09-15
Implementing USM memory management

PokhodenkoSA:
Closes #27.
Cython module created. 
Need refactoring for matching new requirements.

Contains changes from #30.

PR #40, opened 2020-09-18
Dppl program interface

diptorupd:
Add an interface to build OpenCL interoperability kernels and submit them to a Sycl queue. This is mostly experimental stuff that is needed only to support Numba and it might be moved to Numba proper at a later date.


diptorupd:
Closing this PR. Will open a refactored one rebased on top of #34.

PR #50, opened 2020-09-21
Method _usm_type of Memory class should accept sycl_context

oleksandr-pavlyk:
@PokhodenkoSA @diptorupd

### DIFF
--- a/dppl/_memory.pyx
+++ b/dppl/_memory.pyx
@@ -8,14 +8,14 @@ from cpython cimport Py_buffer
 cdef class Memory:
     cdef DPPLSyclUSMRef memory_ptr
     cdef Py_ssize_t nbytes
-    cdef SyclContext context
+    cdef SyclQueue queue
 
     cdef _cinit(self, Py_ssize_t nbytes, ptr_type, SyclQueue queue):
         cdef DPPLSyclUSMRef p
 
         self.memory_ptr = NULL
         self.nbytes = 0
-        self.context = None
+        self.queue = None
 
         if (nbytes > 0):
             if queue is None:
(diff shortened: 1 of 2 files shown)


PokhodenkoSA (approved):
Looks good for me. It would be better if there was a test.
@oleksandr-pavlyk could you please also add a test for this method?

Tests for USM: [dppl/tests/dppl_tests/test_sycl_usm.py](https://github.com/IntelPython/pydppl/blob/master/dppl/tests/dppl_tests/test_sycl_usm.py)

dppl/_memory.pyx
```diff
@@ -68,16 +68,36 @@ cdef class Memory:
...
+        cdef SyclQueue q
+        if context is None:
+            ctx = self._context
+            kind = DPPLUSM_GetPointerType(self.memory_ptr,
+                                          ctx.get_context_ref())
+        elif isinstance(context, SyclContext):
+            ctx = <SyclContext>(context)
+            kind = DPPLUSM_GetPointerType(self.memory_ptr,
+                                          ctx.get_context_ref())
+        elif isinstance(context, SyclQueue):
```
PokhodenkoSA, line 93: What if name the variable `context_or_queue` ?

PokhodenkoSA (approved):
Suggested rename a variable. In other places LGTM.

dppl/tests/dppl_tests/test_sycl_usm.py
```diff
@@ -46,16 +46,30 @@ def test_memory_without_context (self):
     def test_memory_cpu_context (self):
         mobj = self._create_memory()
 
-        # CPU context
+        # CPU context
```
diptorupd, line 49: superfluous white space

diptorupd (commented):
Minor formatting nitpick aside LGTM. Thanks!

dppl/_memory.pyx
diptorupd: Yes, makes sense. Since the argument can be either context or queue. If we do not like long names, let us just call it `syclobj` or something other than context.

PR #57, opened 2020-09-22
Fix leftover PyDPPL use and add/update copyright notices.

diptorupd:


PokhodenkoSA (approved):

dpctl/__init__.py
```diff
@@ -1,43 +1,43 @@
...
+## \file
+## This top-level dpctl module.
+##
+##===----------------------------------------------------------------------===##
 '''
-    Python Data Parallel Processing Library (PyDPPL)
+    Data Parallel Control (dpctl)
 
-    PyDPPL provides a lightweight Python abstraction over DPC++/SYCL and
+    Dpctl provides a lightweight Python abstraction over DPC++/SYCL and
```
PokhodenkoSA, line 28:
```suggestion
    dpCtl provides a lightweight Python abstraction over DPC++/SYCL and
```

PokhodenkoSA (commented):
I think that `dpCtl` is better than `Dpctl`. I would prefer to name the product `dpCtl` because it is easier to read. But in Pyhton world modules and packages should be named `dpctl` because it should be case sensitive.

dpctl/__init__.py
diptorupd: Done. The module is `dpctl` and the project `dpCtl`

PR #68, opened 2020-09-24
Fixes the implementation of the oclrv.Runtime singleton.

### LINKED ISSUE #62: Running unit tests causes failure
On running `python -m unittest dpctl.tests` the test suite passes but then I get this failure.

```
Exception ignored in: <function Runtime.__del__ at 0x7f0ba8c7cb00>
  File "/home/diptorupd/Desktop/devel/dpctl/dpctl/ocldrv.py", line 543, in __del__
AttributeError: 'NoneType' object has no attribute '_runtime'
```
The issue does not happen if I go into the dpctl directory and then do `python -m unittest tests`

diptorupd:
References to the ffi and lib modules are now stored inside Runtime. By doing that we ensure that the objects stay around till Runtime is alive. Without the fix when running unit tests I was encountering a crash stating NoneType has no attribute `_runtime`.

Fixes #62

### DIFF
--- a/dpctl/ocldrv.py
+++ b/dpctl/ocldrv.py
@@ -503,17 +503,18 @@ class Runtime():
     OpenCL devices found on the system.
     '''
 
-    _runtime     = None
-    _cpu_device  = None
-    _gpu_device  = None
-    _curr_device = None
+    _singleton    = None
 
     def __new__(cls):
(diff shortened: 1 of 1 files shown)


PokhodenkoSA (approved):
LGTM

PR #70, opened 2020-09-26
Add wrapper for SYCL queue::memcpy().

PokhodenkoSA:


backends/include/dppl_sycl_queue_interface.h
```diff
@@ -63,4 +63,13 @@ DPPLQueue_GetDevice (__dppl_keep const DPPLSyclQueueRef QRef);
 DPPL_API
 void DPPLQueue_Delete (__dppl_take DPPLSyclQueueRef QRef);
 
+/*!
+ * @brief C-API wrapper for sycl::queue::memcpy. It waits an event.
+ *
+ * @param    QRef           A DPPLSyclQueueRef pointer that gets deleted.
```
oleksandr-pavlyk, line 69: Is this an appropriate comment? Looks like it had been copy-pasted from the declaration above.

dpctl/_sycl_core.pxd
```diff
@@ -63,3 +63,4 @@ cdef class SyclQueue:
     cpdef SyclContext get_sycl_context (self)
     cpdef SyclDevice get_sycl_device (self)
     cdef DPPLSyclQueueRef get_queue_ref (self)
+    cpdef memcpy (self, dest, src, int count)
```
oleksandr-pavlyk, line 66: Did you mean `cdef` ?

oleksandr-pavlyk: For `c` part of `cpdef` to be efficient, `dest` and `src` need have type declarations. Looks like they can only be `Memory` objects. 

The method does not use `self` at all, so should it be a static method of the class?

PokhodenkoSA: The method uses `self` in code `DPPLQueue_memcpy(self.queue_ptr, c_dest, c_src, count)`.

oleksandr-pavlyk: Perhaps it would be better to not expose `memcpy` as a stand-alone symbol, but rather make it two method `copyto` and `copyfrom` of `Memory` object. That way it could copy to/from another Memory object, or host memory of any object that supports a buffer protocol.

backends/include/dppl_sycl_queue_interface.h
PokhodenkoSA: Fixed.

backends/include/dppl_sycl_queue_interface.h
```diff
@@ -63,4 +63,16 @@ DPPLQueue_GetDevice (__dppl_keep const DPPLSyclQueueRef QRef);
...
+/*!
+ * @brief C-API wrapper for sycl::queue::memcpy. It waits an event.
+ *
+ * @param    QRef           An opaque pointer to the sycl queue.
+ * @param    Dest           An USM pointer to the destination memory.
+ * @param    Src            An USM pointer to the source memory.
+ * @param    Count          A number of bytes to copy.
+ */
+DPPL_API
+void DPPLQueue_memcpy (__dppl_keep const DPPLSyclQueueRef QRef,
```
diptorupd, line 75: Minor nitpick. Does it make sense to rename as DPPLQueue_Memcpy to be consistent with rest of the API?

PokhodenkoSA: Done.

dpctl/sycl_core.pyx
```diff
@@ -132,6 +133,21 @@ cdef class SyclQueue:
     cdef DPPLSyclQueueRef get_queue_ref (self):
         return self.queue_ptr
 
+    cpdef memcpy (self, dest, src, int count):
+        cdef void *c_dest, *c_src
```
oleksandr-pavlyk, line 137: Cython generates warning here. It requires declarations on separate lines:

```
cdef void * c_dest
cdef void * c_src
```

oleksandr-pavlyk:
The `q.memcpy` assumes pointers are either host pointers or USM pointers accessible in the context associated with the queue. 

It would make sense for `SyclQueue.memcpy` to raise an exception if this is not the case.

But we still need a way to copy data from one `Memory` instance to another even if their memory reside in different SYCL contexts. So we could have  `Memory.copyto(src)` and `Memory.to_memoryview()`. 

`copyto` would copy from `src` or type `Memory` using host as an intermediary if needed, or from `src` being Python's object exposing buffer interface. 

`Memory.to_memoryview()` will create host memory, copy content of Memory to host memory and return the result as Python's `memoryview` object created from the host memory pointer.

dpctl/sycl_core.pyx
PokhodenkoSA: Fixed.

PokhodenkoSA:
@oleksandr-pavlyk Additional features will be inpmelemted in another PR. The issue for it: https://github.com/IntelPython/dpctl/issues/76

diptorupd:
@oleksandr-pavlyk @PokhodenkoSA With the remaining enhancement being tracked in #76 can we merge this as it stands right now?

PokhodenkoSA:
Lets merge it.

PR #86, opened 2020-10-02
Feature/backend specific queues broken

diptorupd:


PokhodenkoSA:
Variables for manipulating default SYCL backend:
```bash
export SYCL_BE=PI_OPENCL
export SYCL_BE=PI_LEVEL0  
unset SYCL_BE 
```

PR #90, opened 2020-10-03
Improve build scripts

### LINKED ISSUE #98: Run C++ tests in conda recipe
There are GTest tests but they are not started during conda package building.

PokhodenkoSA:
Fixes #98 but it would be better to move it to separate PR.

### DIFF
--- a/backends/CMakeLists.txt
+++ b/backends/CMakeLists.txt
@@ -93,6 +93,8 @@ set(OpenCL_LIBRARY "${DPCPP_ROOT}/lib/libOpenCL.so")
 message(STATUS "OpenCL_INCLUDE_DIR: ${OpenCL_INCLUDE_DIR}")
 message(STATUS "OpenCL_LIBRARY: ${OpenCL_LIBRARY}")
 
+add_definitions(-DCL_TARGET_OPENCL_VERSION=220)
+
 add_library(
   DPPLSyclInterface
   SHARED
--- a/backends/CMakeLists.txt
+++ b/backends/CMakeLists.txt
@@ -79,6 +79,9 @@ cmake_print_variables(OpenCL_LIBRARY)
 
 add_definitions(-DCL_TARGET_OPENCL_VERSION=220)
 
+
+# DPPLSyclInterface
+
 add_library(
   DPPLSyclInterface
   SHARED
@@ -94,14 +97,6 @@ add_library(
   source/dppl_utils.cpp
 )
 
-# Install DPPLOpenCLInterface
-add_library(
-  DPPLOpenCLInterface
-  SHARED
-  source/dppl_opencl_interface.c
-)
-
-# Install DPPLSyclInterface
 target_include_directories(
   DPPLSyclInterface
   PRIVATE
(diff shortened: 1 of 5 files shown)


backends/CMakeLists.txt
```diff
@@ -38,28 +38,13 @@ function (check_for_dpcpp)
...
-        message(STATUS "PYTHON_INCLUDE_DIR: ${PYTHON_INCLUDE_DIR}")
-    endif()
-endfunction()
-
 # Check for dpcpp in the specified DPCPP_ROOT
 check_for_dpcpp()
-# Check if the locations of Numpy and Python headers were provided
-check_for_numpy_inc()
+
+find_package(Python REQUIRED COMPONENTS Development NumPy)
```
diptorupd, line 44: Do these macros search in the conda prefix?

backends/tests/CMakeLists.txt
```diff
@@ -1,49 +1,40 @@
...
-
-  link_directories(${GTEST_LIB_DIR})
-
-  set(PYDPPL_BACKEND_TEST_CASES
-    test_sycl_kernel_interface
-    test_sycl_platform_interface
-    test_sycl_program_interface
-    test_sycl_queue_interface
-    test_sycl_queue_manager
+find_package(GTest REQUIRED)
```
diptorupd, line 1: Same question. I was not sure if the macro is smart enough to look under conda prefix.

conda-recipe/meta.yaml
```diff
@@ -24,6 +24,7 @@ requirements:
         - make  # [unix]
         - ninja  # [win]
         - numpy >=1.17
+        - gtest
```
diptorupd, line 27: gtest 1.10 is needed as I am using GTEST_SKIP that only present in 1.10.

backends/CMakeLists.txt
PokhodenkoSA: Yes.

backends/tests/CMakeLists.txt
PokhodenkoSA: I have tested on Linux. It works fine.

conda-recipe/meta.yaml
PokhodenkoSA:
```suggestion
        - gtest >=1.10
```

backends/tests/CMakeLists.txt
```diff
@@ -1,49 +1,40 @@
...
+
+include_directories(${GTEST_INCLUDE_DIR})
+include_directories(../include)
+
+link_directories(${GTEST_LIB_DIR})
+
+set(PYDPPL_BACKEND_TEST_CASES
+  test_sycl_kernel_interface
+  test_sycl_platform_interface
+  # test_sycl_program_interface
```
PokhodenkoSA, line 20: Note: commented.

backends/tests/CMakeLists.txt
diptorupd: On my system I have `gtest` installed in my conda environment and `find_package` fails to locate Gtest. What do I need to do to fix? I can install `gtest` through `apt`, but that is not what we want right?

diptorupd: Ok. So to get it to work, I need `cmake` to come from conda as well. So it works. Nice!

diptorupd:
Closing the PR as most of it is now implemented and the changes here are bit rotten beyond repair.

PR #96, opened 2020-10-04
Fix/issue 83

### LINKED ISSUE #83: We need to support ctypes uint8 and ulong
We need to support ctypes ctypes.c_uint8 and ctypes.c_ulong in dpctl.

diptorupd:
Closes #83


PokhodenkoSA (approved):
For 1de7ff43781766ee54cdea26d7eb5e50b070e3f5 - LGTM.
Other commits as I understand are from #93.

PR #97, opened 2020-10-05
Source TBB from oneAPI for activating CPU device

PokhodenkoSA:
In this PR I have tried to activate compute runtime for CPU. For some reason it is activated from oneAPI if activate TBB environment.
Finally activated TBB environment does not solve the problem because CPU device is visible in clinfo but it fails when I try to use it.

### DIFF
--- a/conda-recipe/run_test.sh
+++ b/conda-recipe/run_test.sh
@@ -4,6 +4,7 @@ set -e
 
 # Suppress error b/c it could fail on Ubuntu 18.04
 source ${ONEAPI_ROOT}/compiler/latest/env/vars.sh || true
+source ${ONEAPI_ROOT}/tbb/latest/env/vars.sh || true
 
 ${PYTHON} -c "import dpctl"
 ${PYTHON} -c "import dpctl.ocldrv"


conda-recipe/run_test.sh
```diff
@@ -4,6 +4,7 @@ set -e
 
 # Suppress error b/c it could fail on Ubuntu 18.04
 source ${ONEAPI_ROOT}/compiler/latest/env/vars.sh || true
+source ${ONEAPI_ROOT}/tbb/latest/env/vars.sh || true
```
Alexander-Makaryev, line 7: @PokhodenkoSA 
You hiding the error in compiler activation script (about FPGA), I disagree with it but ok... I think that explicit fix of scripts or renaming of oclfpga directory is better.
But why you need `|| true` in tbb activation line?

oleksandr-pavlyk: Generally to avoid failing the script if sourcing the TBB's `vars.sh` returns with non-zero code.

PokhodenkoSA: It does not help anyway. Sourcing TBB does not make CPU device work.

oleksandr-pavlyk:
It would be useful to understand the problem this PR is solving. Empty description does not help.

PokhodenkoSA:
@oleksandr-pavlyk Sorry. I will add it soon.

conda-recipe/run_test.sh
Alexander-Makaryev: It happens because someone changed "latest" compiler with beta10 version.
TBB has beta09 version, so it is incompatible.
source another version of compiler and it will work.

Alexander-Makaryev: I mean, device itself will be available in clinfo for example, but error -2 still occurs...

oleksandr-pavlyk:
Has this PR gone stale? If so, please close.

PR #103, opened 2020-10-05
onaAPI beta10 compiler

PokhodenkoSA:
With oneAPI beta10 compiler it is not working to use `CC=dpcpp.exe`. Use `CC=clang-cl.exe` instead.


diptorupd (approved):

diptorupd:
Thanks!

PR #87, opened 2020-10-02
Add functions to get address of ref

reazulhoque:
This PR will allow consumers to access the address of the refs as value.

PR #105, opened 2020-10-05
Feature/get ref address

diptorupd:
The PR reworks #87 to use the same name `addressof_ref` for all classes that wrap DPPLSyclxxxx objects. Added docstring.

The change is needed because inside Numba we are directly calling dpctl's C API during LLVM lowering for kernel submit call. Having these convenience functions lets us get the address of the DPPLSyclKernelRef from a SyclKernel Python object.

To standardize the API the same `addressof_ref` call was added for all classes.


diptorupd:
@PokhodenkoSA  can be merged once CI completes.

@reazulhoque Please note the name change for the functions.

PokhodenkoSA (approved):
LGTM. Will merge it.

dpctl/sycl_core.pyx
```diff
@@ -250,6 +287,16 @@ cdef class SyclProgram:
...
 
+    def addressof_ref (self):
+        """Returns the address of the C API DPPLSyclProgramRef pointer
+        as a long.
+
+        Returns:
+            The address of the DPPLSyclProgramRef object used to create this
+            SyclProgram cast to a long.
+        """
+        return int(<long>self._program_ref)
```
oleksandr-pavlyk, line 298: @PokhodenkoSA @diptorupd 

Long is not good here. It is a 32-bit integer on Windows. We should be using `<size_t>` instead.

Issue #110, opened 2020-10-06
Change address casts to use size_t instead of long


diptorupd:
@PokhodenkoSA @diptorupd 

Long is not good here. It is a 32-bit integer on Windows. We should be using `<size_t>` instead.

_Originally posted by @oleksandr-pavlyk in https://github.com/IntelPython/dpctl/pull/105#discussion_r500279366_

diptorupd:
Done in #111

PR #111, opened 2020-10-06
Change cast from long to size_t

diptorupd:
Addresses  #110

The addressof_ref methods of the SyclXXX classes should use size_t instead of long when casting the address to an Python int.


oleksandr-pavlyk (approved):
LGTM.

PR #116, opened 2020-10-07
Fix failing platform test case.

diptorupd:
Fixes for the scenarios where no sycl device except host exists.


diptorupd:
Closes #119

PokhodenkoSA (approved):
LGTM. Checked up to SDC on Linux.

PR #117, opened 2020-10-08
Fixed typo in the code for set_default_queue

oleksandr-pavlyk:
Allowing both strings or enums as selectors for
backend and device.


oleksandr-pavlyk:
@diptorupd @PokhodenkoSA

PokhodenkoSA (approved):

Issue #106, opened 2020-10-05
Remove dppl_opencl_interface.{h|c}, ocldrv.py, opencl_core.py


diptorupd:
The Legacy OpenCL interface is no longer needed and should be dropped.

Remove 
- [x] backends/include/dppl_opencl_interface.h
- [x] backends/source/dppl_opencl_interface.c
- [x] backends/include/error_check_macros.h
- [x] dpctl/ocldrv.py
- [x] dpctl/opencl_core.py

Other changes will be needed inside setup.py, CmakeLists.txt etc.

PR #138, opened 2020-10-15
Remove OpenCL interface

1e-to:
This PR closes https://github.com/IntelPython/dpctl/issues/106


diptorupd (commented):
cffi is no longer needed once ocldrv is removed, so meta.yaml should also be changed. Apart from that looks fine. Please make sure CI passes on both Windows and Linux.

PokhodenkoSA (approved):
As CI passed last time it LGTM.
I expect some things are not cleared but I can not list all it now.
I will merge if CI pass now.

PokhodenkoSA:
I would like CHANGE_LOG to be updated too.

diptorupd:
I tested it locally and things looked ok. Tested both `$scripts/build_for_develop.sh` and `conda build conda-recipe`.

PR #150, opened 2020-10-21
Change funcs to cpdef

### LINKED ISSUE #112: get_current_queue and other functions should be exposed as cpdef functions
dpctl functions `get_current_queue`, `get_current_device`, `get_current_backend` are not exposed as C functions. Other Cython modules that want to use dpctl cannot do a cimport to access these functions.

We should redefine these functions as cpdef and expose them via _sycl_core.pxd.

1e-to:
This PR closes #112 

It makes `get_current_backend` and `get_current_device_type` cimportable in Cython.


oleksandr-pavlyk:
Force push of the master introduced and additional unintended commits into this and my PR.

The last two commits of this PR must be cherry-picked on top of the new tip of the master 6684c7b, or perhaps @diptorupd should merge pull/#121, and then these two commits should be cherry-picked on top of that merge commit. 

The same comment applies to pull/#152

PR #152, opened 2020-10-21
Feature/memory zero copy

### LINKED ISSUE #43: Expose USM aligned_alloc* functions

### LINKED ISSUE #45: Expose the USM fine-grained control functions `mem_advise` and `prefetch` in the C API
Can we expose the USM fine-grained control functions `mem_advise` and `prefetch` in the C API? We might want to use them in Numba.

_Originally posted by @diptorupd in https://github.com/IntelPython/pydppl/issues/27#issuecomment-694904742_

### LINKED ISSUE #76: Make methods of Memory class for copy to and from another objects
Perhaps it would be better to not expose `memcpy` as a stand-alone symbol, but rather make it two method `copyto` and `copyfrom` of `Memory` object. That way it could copy to/from another Memory object, or host memory of any object that supports a buffer protocol.

_Originally posted by @oleksandr-pavlyk in https://github.com/IntelPython/dpctl/pull/70#discussion_r495948029_

The q.memcpy assumes pointers are either host pointers or USM pointers accessible in the context associated with the queue.

It would make sense for SyclQueue.memcpy to raise an exception if this is not the case.

But we still need a way to copy data from one Memory instance to another even if their memory reside in different SYCL contexts. So we could have Memory.copyto(src) and Memory.to_memoryview().

copyto would copy from src or type Memory using host as an intermediary if needed, or from src being Python's object exposing buffer interface.

Memory.to_memoryview() will create host memory, copy content of Memory to host memory and return the result as Python's memoryview object created from the host memory pointer.

_Originally posted by @oleksandr-pavlyk in https://github.com/IntelPython/dpctl/pull/70#issuecomment-700746734_

### LINKED ISSUE #120: Move _memory submodule directly under dpctl
We currently need to do `import dpctl._memory`. Instead we should be able to directly access `_memory` members by importing `dpctl`.

oleksandr-pavlyk:
Closes #43, closes #76, closes #45, closes #120.

Work on `Memory` class and its derivatives. 

Classes `MemoryUSMShared`, `MemoryUSMDevice`, and `MemoryUSMHost` constructors allow  to specify the number of bytes to allocate, and optionally the `queue`, or give a Python object expected to implement `__sycl_usm_array_interface__` attribute to create the USM memory object pointing the same USM memory with zero copy.

Classes also implement several new features. 

 - Method `copy_from_host(pyobj)` to copy content of Python object buffer to USM memory
 - Method `copy_to_hist(pyobj=None)` to copy content of USM memory to pre-allocated Python object buffer. If object does not implement the buffer protocol, `bytearray` of the appropriate size is allocated and populated. The method returns the populated object.
- Method `copy_from_device(sycl_usm_obj)` to copy content of USM memory behind Python object that is expected to implement ``__sycl_usm_array_interface__`` to the USM memory of the instance.

Classes are now pickleable, with it being implemented via copying to host and picking that. 

@Alexander-Makaryev

### DIFF
--- a/backends/include/dppl_sycl_queue_interface.h
+++ b/backends/include/dppl_sycl_queue_interface.h
@@ -209,7 +209,7 @@ void DPPLQueue_Memcpy (__dppl_keep const DPPLSyclQueueRef QRef,
  */
 DPPL_API
 void DPPLQueue_Prefetch (__dppl_keep DPPLSyclQueueRef QRef,
-			 const void *Ptr, size_t Count);
+                         const void *Ptr, size_t Count);
 
 /*!
  * @brief C-API wrapper for sycl::queue::mem_advise, the function waits on an event
@@ -223,6 +223,6 @@ void DPPLQueue_Prefetch (__dppl_keep DPPLSyclQueueRef QRef,
  */
 DPPL_API
 void DPPLQueue_MemAdvise (__dppl_keep DPPLSyclQueueRef QRef,
-			  const void *Ptr, size_t Count, int Advice);
+                          const void *Ptr, size_t Count, int Advice);
 
 DPPL_C_EXTERN_C_END
--- a/backends/include/dppl_sycl_queue_interface.h
+++ b/backends/include/dppl_sycl_queue_interface.h
@@ -199,4 +199,30 @@ DPPL_API
 void DPPLQueue_Memcpy (__dppl_keep const DPPLSyclQueueRef QRef,
                        void *Dest, const void *Src, size_t Count);
 
+/*!
+ * @brief C-API wrapper for sycl::queue::prefetch, the function waits on an event
+ * till the prefetch operation completes.
+ *
+ * @param    QRef           An opaque pointer to the sycl queue.
+ * @param    Ptr            An USM pointer to memory.
+ * @param    Count          A number of bytes to prefetch.
+ */
+DPPL_API
+void DPPLQueue_Prefetch (__dppl_keep DPPLSyclQueueRef QRef,
+			 const void *Ptr, size_t Count);
+
+/*!
+ * @brief C-API wrapper for sycl::queue::mem_advise, the function waits on an event
+ * till the operation completes.
+ *
+ * @param    QRef           An opaque pointer to the sycl queue.
+ * @param    Ptr            An USM pointer to memory.
+ * @param    Count          A number of bytes to prefetch.
+ * @param    Advice         Device-defined advice for the specified allocation. 
+ *                           A value of 0 reverts the advice for Ptr to the default behavior.
+ */
+DPPL_API
+void DPPLQueue_MemAdvise (__dppl_keep DPPLSyclQueueRef QRef,
+			  const void *Ptr, size_t Count, int Advice);
+
 DPPL_C_EXTERN_C_END
--- a/backends/include/dppl_sycl_queue_manager.h
+++ b/backends/include/dppl_sycl_queue_manager.h
@@ -167,6 +167,11 @@ void DPPLQueueMgr_PopQueue ();
  * ownership of the queue reference and should deallocate it using
  * DPPLQueue_Delete.
  *
+ * @param    CRef           Sycl context reference
+ * @param    DRef           Sycl device reference
+ *
+ * @return A copy of the sycl::queue created from given context and device
+ * references.
  */
 DPPL_API
 __dppl_give DPPLSyclQueueRef
--- a/backends/include/dppl_sycl_queue_manager.h
+++ b/backends/include/dppl_sycl_queue_manager.h
@@ -171,7 +171,7 @@ void DPPLQueueMgr_PopQueue ();
 DPPL_API
 __dppl_give DPPLSyclQueueRef
 DPPLQueueMgr_GetQueueFromContextAndDevice(__dppl_keep DPPLSyclContextRef CRef,
-					  __dppl_keep DPPLSyclDeviceRef DRef);
+                                          __dppl_keep DPPLSyclDeviceRef DRef);
 
 
 DPPL_C_EXTERN_C_END
--- a/backends/include/dppl_sycl_queue_manager.h
+++ b/backends/include/dppl_sycl_queue_manager.h
@@ -158,4 +158,20 @@ DPPLQueueMgr_PushQueue (DPPLSyclBackendType BETy,
 DPPL_API
 void DPPLQueueMgr_PopQueue ();
 
+
+/*!
+ * @brief Creates a new instance of SYCL queue from SYCL context and
+ * SYCL device.
+ *
+ * The instance is not placed into queue manager. The user assumes
+ * ownership of the queue reference and should deallocate it using
+ * DPPLQueue_Delete.
+ *
+ */
+DPPL_API
+__dppl_give DPPLSyclQueueRef
+DPPLQueueMgr_GetQueueFromContextAndDevice(__dppl_keep DPPLSyclContextRef CRef,
+					  __dppl_keep DPPLSyclDeviceRef DRef);
+
+
 DPPL_C_EXTERN_C_END
--- a/backends/include/dppl_sycl_usm_interface.h
+++ b/backends/include/dppl_sycl_usm_interface.h
@@ -36,6 +36,9 @@ DPPL_C_EXTERN_C_BEGIN
 /*!
  * @brief Crete USM shared memory.
  *
+ * @param    size     Number of bytes to allocate
+ * @param    QRef     Sycl queue reference to use in allocation
+ *
  * @return The pointer to USM shared memory.
  */
 DPPL_API
@@ -45,7 +48,11 @@ DPPLmalloc_shared (size_t size, __dppl_keep const DPPLSyclQueueRef QRef);
 /*!
  * @brief Crete USM shared memory.
  *
- * @return The pointer to USM shared memory with requested alignment.
+ * @param  alignment   Allocation's byte alignment
+ * @param  size        Number of bytes to allocate
+ * @param  QRef        Sycl queue reference to use in allocation
+ *
+ * @return The pointer to USM shared memory with the requested alignment.
  */
 DPPL_API
 __dppl_give DPPLSyclUSMRef
@@ -54,6 +61,9 @@ DPPLaligned_alloc_shared (size_t alignment, size_t size,
 
 /*!
  * @brief Crete USM host memory.
+ *
+ * @param    size     Number of bytes to allocate
+ * @param    QRef     Sycl queue reference to use in allocation
  *
  * @return The pointer to USM host memory.
  */
@@ -64,7 +74,11 @@ DPPLmalloc_host (size_t size, __dppl_keep const DPPLSyclQueueRef QRef);
 /*!
  * @brief Crete USM host memory.
  *
- * @return The pointer to USM host memory with requested alignment.
+ * @param  alignment   Allocation's byte alignment
+ * @param  size        Number of bytes to allocate
+ * @param  QRef        Sycl queue reference to use in allocation
+ *
+ * @return The pointer to USM host memory with the requested alignment.
  */
 DPPL_API
 __dppl_give DPPLSyclUSMRef
@@ -74,6 +88,9 @@ DPPLaligned_alloc_host (size_t alignment, size_t size,
 /*!
  * @brief Crete USM device memory.
  *
+ * @param    size     Number of bytes to allocate
+ * @param    QRef     Sycl queue reference to use in allocation
+ *
  * @return The pointer to USM device memory.
  */
 DPPL_API
@@ -82,6 +99,10 @@ DPPLmalloc_device (size_t size, __dppl_keep const DPPLSyclQueueRef QRef);
 
 /*!
  * @brief Crete USM device memory.
+ *
+ * @param    alignment   Allocation's byte alignment
+ * @param    size        Number of bytes to allocate
+ * @param    QRef        Sycl queue reference to use in allocation
  *
  * @return The pointer to USM device memory with requested alignment.
  */
@@ -93,6 +114,11 @@ DPPLaligned_alloc_device (size_t alignment, size_t size,
 /*!
  * @brief Free USM memory.
  *
+ * @param   MRef      USM pointer to free
+ * @param   QRef      Sycl queue reference to use.
+ *
+ * USM pointer must have been allocated using the same context as the one
+ * used to construct the queue.
  */
 DPPL_API
 void DPPLfree_with_queue (__dppl_take DPPLSyclUSMRef MRef,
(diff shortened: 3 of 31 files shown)


oleksandr-pavlyk:
By the way, reconstruction of `queue` from a pointer and context could proceed as follows:

```
   if isinstance(syclobj, SyclContext):
        ctx = <SyclContext> syclobj
        dev = DPPLget_pointer_device(data_ptr, ctx)
        new_queue = DPPLqueue_from_device_and_conext(dev, ctx)
```

oleksandr-pavlyk:
Implement support for aligned memory allocation, exposing `DPPL_aligned_alloc_shared` and friends.

Changed constructor of `MemoryUSMShared` to require and new keyword `alignment` and old one `queue` be keyword arguments only.

Non-positive alignment values are ignored and `malloc_*` is silently used instead.

backends/tests/test_sycl_queue_manager.cpp
```diff
@@ -244,6 +244,22 @@ TEST_F (TestDPPLSyclQueueManager, CheckIsCurrentQueue2)
     DPPLQueueMgr_PopQueue();
 }
 
+TEST_F (TestDPPLSyclQueueManager, CreateQueueFromDeviceAndContext)
+{
+    auto Q = DPPLQueueMgr_GetCurrentQueue();
+    auto D = DPPLQueue_GetDevice(Q);
```
diptorupd, line 250: Not important here, but all the `DPPL*_Get*()` calls should be followed by a corresponding `Delete` to avoid leaks.

backends/tests/test_sycl_usm_interface.cpp
```diff
@@ -0,0 +1,181 @@
...
+            ret = true;
+            break;
+        }
+    }
+    return ret;
+}
+
+void common_test_body(size_t nbytes, const DPPLSyclUSMRef Ptr, const DPPLSyclQueueRef Q, const char *expected) {
+
+    auto Ctx = DPPLQueue_GetContext(Q);
```
diptorupd, line 61: Should be deleted after use.

backends/tests/test_sycl_usm_interface.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    auto kind = DPPLUSM_GetPointerType(Ptr, Ctx);
+    EXPECT_TRUE(0 == std::strncmp(kind, expected, 4));
+
+    auto Dev = DPPLUSM_GetPointerDevice(Ptr, Ctx);
```
diptorupd, line 66: The device pointers should be deleted.

backends/tests/test_sycl_usm_interface.cpp
```diff
@@ -0,0 +1,181 @@
...
+    ~TestDPPLSyclUSMInterface ()
+    {  }
+};
+
+TEST_F(TestDPPLSyclUSMInterface, MallocShared)
+{
+    if (!has_devices())
+	GTEST_SKIP_("Skipping: No Sycl Devices.\n");
+
+    auto Q = DPPLQueueMgr_GetCurrentQueue();
```
diptorupd, line 90: Q should be freed as well. (Same for the rest of the functions)

dpctl/__init__.py
```diff
@@ -47,6 +47,7 @@
 __author__ = "Intel Corp."
 
 from ._sycl_core import *
+from ._memory import MemoryUSMShared, MemoryUSMDevice, MemoryUSMHost
```
diptorupd, line 50: Do we want this or something like `dpctl.memory.MemoryUSMShared` etc.?

dpctl/_backend.pxd
```diff
@@ -186,6 +186,10 @@ cdef extern from "dppl_sycl_queue_interface.h":
     cdef void DPPLQueue_Wait (const DPPLSyclQueueRef QRef)
     cdef void DPPLQueue_Memcpy (const DPPLSyclQueueRef Q,
                                 void *Dest, const void *Src, size_t Count)
+    cdef void DPPLQueue_Prefetch (const DPPLSyclQueueRef Q,
+                                const void *Src, size_t Count)
```
diptorupd, line 190: nitpick

dpctl/_memory.pyx
```diff
@@ -30,33 +30,171 @@
...
+import numpy as np
 
-cdef class Memory:
+cdef _throw_sycl_usm_ary_iface():
+    raise ValueError("__sycl_usm_array_interface__ is malformed")
 
-    cdef _cinit(self, Py_ssize_t nbytes, ptr_type, SyclQueue queue):
-        cdef DPPLSyclUSMRef p
 
+cdef void copy_via_host(void *dest_ptr, SyclQueue dest_queue,
```
diptorupd, line 45: If the memcpy calls are changed to non-blocking, this function should also be made non-blocking?

dpctl/_memory.pyx
```diff
@@ -30,33 +30,171 @@
...
+        arr_data_ptr = <Py_ssize_t>ary_data_tuple[0]
+        buf.p = <DPPLSyclUSMRef>(<void*>arr_data_ptr)
+        buf.writeable = 1 if ary_data_tuple[1] else 0
+        buf.itemsize = <Py_ssize_t>(dt.itemsize)
+        buf.nbytes = (<Py_ssize_t>ary_shape[0]) * buf.itemsize
+
+        if isinstance(ary_syclobj, dpctl.SyclQueue):
+            buf.queue = <SyclQueue>ary_syclobj
+        else:
+            # FIXME: need a way to construct a queue from
```
diptorupd, line 126: FIXME resolved?

diptorupd (commented):
Overall looks really good. Thanks for getting this done.

Mostly nitpicks, except the question about if we should return an event from the copy functions.

I want to sit with you and understand the Python bits a bit more.

backends/include/dppl_sycl_queue_interface.h
```diff
@@ -199,4 +199,30 @@ DPPL_API
...
+			 const void *Ptr, size_t Count);
+
+/*!
+ * @brief C-API wrapper for sycl::queue::mem_advise, the function waits on an event
+ * till the operation completes.
+ *
+ * @param    QRef           An opaque pointer to the sycl queue.
+ * @param    Ptr            An USM pointer to memory.
+ * @param    Count          A number of bytes to prefetch.
+ * @param    Advice         Device-defined advice for the specified allocation.
```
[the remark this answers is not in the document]
oleksandr-pavlyk: Sycl does not.  Standard says ``event mem_advise(const void *addr, size_t num_bytes, int advice);``

backends/include/dppl_sycl_queue_manager.h
```diff
@@ -158,4 +158,20 @@ DPPLQueueMgr_PushQueue (DPPLSyclBackendType BETy,
 DPPL_API
 void DPPLQueueMgr_PopQueue ();
 
+
+/*!
+ * @brief Creates a new instance of SYCL queue from SYCL context and
+ * SYCL device.
+ *
```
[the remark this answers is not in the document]
oleksandr-pavlyk: Added.

backends/include/dppl_sycl_usm_interface.h
```diff
@@ -60,6 +80,16 @@ DPPL_API
 __dppl_give DPPLSyclUSMRef
 DPPLmalloc_device (size_t size, __dppl_keep const DPPLSyclQueueRef QRef);
 
+/*!
+ * @brief Crete USM device memory.
+ *
```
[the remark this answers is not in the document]
oleksandr-pavlyk: Added. Not really sure why this is not marked as outdated.

backends/source/dppl_sycl_queue_interface.cpp
```diff
@@ -290,10 +290,28 @@ DPPLQueue_Wait (__dppl_keep DPPLSyclQueueRef QRef)
     SyclQueue->wait();
 }
 
-void DPPLQueue_Memcpy (__dppl_take const DPPLSyclQueueRef QRef,
+void DPPLQueue_Memcpy (__dppl_keep const DPPLSyclQueueRef QRef,
                        void *Dest, const void *Src, size_t Count)
 {
     auto Q = unwrap(QRef);
     auto event = Q->memcpy(Dest, Src, Count);
     event.wait();
```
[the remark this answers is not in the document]
oleksandr-pavlyk: I would do it in a separate PR. If we do it, we should do the same for `prefetch` and `mem_advise`, as well as add `memset`.

backends/source/dppl_sycl_queue_interface.cpp
```diff
@@ -290,10 +290,28 @@ DPPLQueue_Wait (__dppl_keep DPPLSyclQueueRef QRef)
...
 }
+
+void
+DPPLQueue_Prefetch (__dppl_keep DPPLSyclQueueRef QRef,
+		    const void *Ptr, size_t Count)
+{
+    auto Q = unwrap(QRef);
+    auto event = Q->prefetch(Ptr, Count);
+    event.wait();
```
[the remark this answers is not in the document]
oleksandr-pavlyk: Agreed.

backends/tests/test_sycl_queue_manager.cpp
```diff
@@ -244,6 +244,22 @@ TEST_F (TestDPPLSyclQueueManager, CheckIsCurrentQueue2)
...
+    auto C = DPPLQueue_GetContext(Q);
+
+    auto Q2 = DPPLQueueMgr_GetQueueFromContextAndDevice(C, D);
+    auto D2 = DPPLQueue_GetDevice(Q2);
+    auto C2 = DPPLQueue_GetContext(Q2);
+
+    EXPECT_TRUE(DPPLDevice_AreEq(D, D2));
+    EXPECT_TRUE(DPPLContext_AreEq(C, C2));
+
```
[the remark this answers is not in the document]
oleksandr-pavlyk: Queues are going to be pointer-different, but devices and contexts should be the same.

dpctl/__init__.py
oleksandr-pavlyk: I am undecided. On one hand, there is an appeal of simplicity of having it in `dpctl`, on another hand, having it there forces import of `numpy`, which currently presents some problems for `dpnp`.

dpctl/_memory.pyx
oleksandr-pavlyk: Yes, removed.

dpctl/_memory.pyx
oleksandr-pavlyk: Yes. It could return an event, but copying from host to device depends on the copying from device to host to complete, so we will need to somehow reflect that. 

Perhaps `copy_via_host` belongs in C++, in SYCLinterface to make this easier to implement.

backends/include/dppl_sycl_usm_interface.h
```diff
@@ -86,4 +116,13 @@ const char *
 DPPLUSM_GetPointerType (__dppl_keep const DPPLSyclUSMRef MRef,
                         __dppl_keep const DPPLSyclContextRef CRef);
 
+/*!
+ * @brief Get the device associated with USM pointer.
+ *
```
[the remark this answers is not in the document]
oleksandr-pavlyk: Done.

backends/source/dppl_sycl_queue_manager.cpp
```diff
@@ -534,3 +536,18 @@ void DPPLQueueMgr_PopQueue ()
...
+ * The function constructs a new SYCL queue instance from SYCL conext and
+ * SYCL device.
+ */
+DPPLSyclQueueRef
+DPPLQueueMgr_GetQueueFromContextAndDevice(__dppl_keep DPPLSyclContextRef CRef,
+					  __dppl_keep DPPLSyclDeviceRef DRef)
+{
+    auto dev = unwrap(DRef);
+    auto ctx = unwrap(CRef);
+    auto q = queue(*ctx, *dev);
```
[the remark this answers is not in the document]
oleksandr-pavlyk: Yes, and this should be accompanied by proper use of `except +` and `except *` in Cython.

backends/tests/test_sycl_usm_interface.cpp
```diff
@@ -0,0 +1,181 @@
...
+            continue;
+        if(!p.get_devices().empty()) {
+            ret = true;
+            break;
+        }
+    }
+    return ret;
+}
+
+void common_test_body(size_t nbytes, const DPPLSyclUSMRef Ptr, const DPPLSyclQueueRef Q, const char *expected) {
```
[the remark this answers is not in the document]
fschlimb: Welcome to the 90ies...Come on, in the 2020s nobody should be limited by a 80 char terminal...
Allowing 100 or 120 makes code much more readable.

dpctl/__init__.py
diptorupd: my personal preference is to have a memory sub-module under `dpctl`.

oleksandr-pavlyk: Ok. We need to close #120 then.

oleksandr-pavlyk: What about `cimport` ? Should it also be `cimport dpctl.memory`? If so, perhaps the easiest is to rename `_memory.pxd` and `_memory.pyx` to remove the underscore.

oleksandr-pavlyk:
@diptorupd The refactoring is complete. Lapse in logic on L109 of `_memory.pyx` fixed, `Memory` help class renamed `_Memory`, and `dpctl._memory` moved to `dpctl.memory._memory` with user-facing `dpctl.memory` module introduced.

backends/source/dppl_sycl_queue_manager.cpp
```diff
@@ -534,3 +536,17 @@ void DPPLQueueMgr_PopQueue ()
 {
     QMgrHelper::popSyclQueue();
 }
+
+/*!
+ * The function constructs a new SYCL queue instance from SYCL conext and
+ * SYCL device.
+ */
+DPPLSyclQueueRef
+DPPLQueueMgr_GetQueueFromContextAndDevice(__dppl_keep DPPLSyclContextRef CRef,
```
diptorupd: space before open paren.

diptorupd (approved):
Please take care of the minor formatting stuff and then merge.

oleksandr-pavlyk:
Windows build failure in CI is not related to changes in this PR at all - test environment fails to activate. 

I was able to build the package on Windows manually, and test suite run was clean.

backends/source/dppl_sycl_platform_interface.cpp
```diff
@@ -41,7 +41,7 @@ get_set_of_non_hostbackends ()
 {
     std::set<DPPLSyclBackendType> be_set;
     for (auto p : platform::get_platforms()) {
-		if(p.is_host())
```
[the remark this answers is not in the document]
oleksandr-pavlyk: Courtesy of Emacs's untabify. I edited and pushed changes to restore formatting.

PR #155, opened 2020-10-28
Fix warnings in building libDPPLSyclInterface

### LINKED ISSUE #144: Fix all warnings
It is hard to see problems if there are many warnings during compilation of C++ and Cython modules.

diptorupd:
Fixes #144


oleksandr-pavlyk:
Did you mean to push `` backends/warning.out``?

diptorupd:
@oleksandr-pavlyk Currently we are getting a flood of warning about the `tp_print` deprecation when we build our extensions. I looked up and these changes relate to [PEP 0590](https://www.python.org/dev/peps/pep-0590/) (also refer to https://bugs.python.org/issue37250, https://github.com/cython/cython/issues/2976, https://github.com/cython/cython/commit/f10a0a391edef10bd37095af87f521808cb362f7).

Based on what I see Cython should only include the `tp_print` member for when `#if PY_VERSION_HEX < 0x030800B1` is satisfied. I am using Python 3.8.5 and still see this warning. 

So, the question is how is the `PY_VERSION_HEX` set by Cython?

There are ways to work around the warning, e.g., https://github.com/openvinotoolkit/openvino/pull/2684/files that we can try. Any thoughts?

oleksandr-pavlyk:
Cython does not set `PY_VERSION_HEX`. It comes from `Python.h`. 

Cython may be generating different C/C++ sources based on that value, but Cython does change it AFAIK.

diptorupd:
@oleksandr-pavlyk I think this can be merged. It fixes all the warnings in the C++ code.

oleksandr-pavlyk (approved):

diptorupd:
Merging after testing locally on Windows.

PR #158, opened 2020-10-29
Changes to build_for_develop.bat

oleksandr-pavlyk:
1. Only source oneAPI compiler activation script if dpcpp.exe --version
   was not successful to avoid polluting env. variables
2. Script is responsive to USE_GTEST env. variable
   If it is undefined, GTEST portion will be skipped.

To build without GTEST
```
set USE_GTEST=
scripts\build_for_develop.bat
```

to build with GTEST:

```
set USE_GTEST=1
scripts\build_for_develop.bat
```

### DIFF
--- a/scripts/build_for_develop.bat
+++ b/scripts/build_for_develop.bat
@@ -1,5 +1,10 @@
-call "%ONEAPI_ROOT%\compiler\latest\env\vars.bat"
-IF ERRORLEVEL 1 exit /b 1
+REM check if oneAPI has been activated, only try activating if not
+dpcpp.exe --version >nul 2>&1
+IF %ERRORLEVEL% NEQ 0 (
+    set ERRORLEVEL=
+    call "%ONEAPI_ROOT%\compiler\latest\env\vars.bat"
+    IF ERRORLEVEL 1 exit /b 1
+)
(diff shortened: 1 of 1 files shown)


scripts/build_for_develop.bat
```diff
@@ -19,25 +24,34 @@ for /f "delims=" %%a in ('%CONDA_PREFIX%\python.exe -c "import numpy; print(nump
...
+if defined USE_GTEST (
+    set "_GTEST_INCLUDE_DIR=%CONDA_PREFIX%\Library\include"
+    set "_GTEST_LIB_DIR=%CONDA_PREFIX%\Library\lib"
+) else (
+    set "_GTEST_INCLUDE_DIR="
+    set "_GTEST_LIB_DIR="
+)
 cmake -G Ninja ^
-    -DCMAKE_BUILD_TYPE=Release ^
+    -DCMAKE_BUILD_TYPE=Debug ^
```
diptorupd, line 35: Let's keep this as Release and add a comment that it may not work if changed to Debug because of C++ std lib problems.

diptorupd (approved):

scripts/build_for_develop.bat
oleksandr-pavlyk: Sure. Will push.

PR #163, opened 2020-10-31
CMake and Gtest improvements

diptorupd:
Various clean ups and improvements to our gtest infrastructure:

- Use `find_package(GTest)` instead of the home grown gtest locator.
- Remove NumPy and Python includes as these are not needed and if required can be included using `finda_package`.
- Now building a single test runner executable for all tests.
- Do not print verbose `ctest` output by default and run tests in parallel.

### DIFF
--- a/backends/CMakeLists.txt
+++ b/backends/CMakeLists.txt
@@ -123,8 +123,17 @@ foreach(HEADER ${HEADERS})
   install(FILES "${HEADER}" DESTINATION include/Support)
 endforeach()
 
+option(
+    BUILD_CAPI_TESTS
+    "Build dpctl C API google tests"
+    OFF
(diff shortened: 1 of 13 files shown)


diptorupd:
@PokhodenkoSA I am now creating a single executable for each test and using a common test runner. Is this what you were suggesting?

What I do not like about it is, I now cannot build and run a single test. Such a feature was useful when debugging a single test failure. Is there a way to do such a thing?

diptorupd:
> What I do not like about it is, I now cannot build and run a single test. Such a feature was useful when debugging a single test failure. Is there a way to do such a thing?

Actually, it is extremely simple by using `ctest`. OK I like this change now :)

PR #175, opened 2020-11-12
Use DPC++ runtime from package `dpcpp_cpp_rt`

PokhodenkoSA:


diptorupd:
Thank you for getting to it :). I wanted to do this eventually.

README.md
```diff
@@ -20,12 +20,14 @@ Build and Install Conda Package
 conda create -n build-env conda-build
 conda activate build-env
 ```
-2. Build conda package
+2. Activate oneAPI DPC++ compiler environmnet and build conda package
 ```bash
+source ${ONEAPI_ROOT}/compiler/latest/env/vars.sh
```
diptorupd, line 25: I encountered few things when building this locally that we should capture in the README:

- We need IntelPython to get the `dpcpp_cpp_rt` package
- `ONEAPI_ROOT` should be defined separately if `setvars.sh` is not called.

diptorupd (approved):
Apart from the changes to README LGTM. Thanks!

README.md
oleksandr-pavlyk: A comment about the assumptions that oneAPI has been activated just above that line would be very appropriate in `README.md`.

oleksandr-pavlyk (approved):

PR #186, opened 2020-11-30
Fix problems in dpctl identified with Klockwork static code analysis

vlad-perevezentsev:
https://jira.devtools.intel.com/browse/SAT-3619

An up - to-date analysis of the klocwork master branch showed 3 warnings. 2 match the previous analysis and have been corrected. Warning memory leak is unchanged because it was created deliberately(Sergey Pokhodenko).

The memory leak warning is still displayed in the klocwork report. I don't know what to do with it. klocwork does not support comments in the code for skipping analysis. You can add a key to build the report table so that the warning is not displayed in the table.

### DIFF
--- a/backends/source/dppl_sycl_platform_interface.cpp
+++ b/backends/source/dppl_sycl_platform_interface.cpp
@@ -148,7 +148,12 @@ size_t DPPLPlatform_GetNumNonHostPlatforms ()
 
 size_t DPPLPlatform_GetNumNonHostBackends ()
 {
-    return get_set_of_non_hostbackends().size();
+    auto be_set = get_set_of_non_hostbackends();
+
+    if (be_set.empty())
+        return 0;
+
+    return be_set.size();
 }
 
 __dppl_give DPPLSyclBackendType *DPPLPlatform_GetListOfNonHostBackends ()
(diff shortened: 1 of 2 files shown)


backends/source/dppl_sycl_platform_interface.cpp
```diff
@@ -148,7 +148,12 @@ size_t DPPLPlatform_GetNumNonHostPlatforms ()
 
 size_t DPPLPlatform_GetNumNonHostBackends ()
 {
-    return get_set_of_non_hostbackends().size();
+    auto be_set = get_set_of_non_hostbackends();
```
PokhodenkoSA, line 151: What was this problem?

vlad-perevezentsev: Non-void function does not return value

PokhodenkoSA: `return get_set_of_non_hostbackends().size();` does not return value?

vlad-perevezentsev: The value should be returned, but at this point klocwork shows a warning

diptorupd: Weird that klockwork requires this change, but if it fixes the warning then I am OK with this change.

diptorupd (commented):
Are there only these two changes? If so, we can merge this and silence the memory leak warning using @vlad-perevezentsev's idea.

PokhodenkoSA:
@diptorupd we need to get new Klockwork results from CI.

PokhodenkoSA (approved):
LGTM. Will see next Klokwork results after merge.

PR #202, opened 2020-12-05
Update gold 2021 to 0.5.0rc2

diptorupd:
0.5.0rc2 has all the renaming changes from dppl to dpctl.


dpctl/__init__.pxd
```diff
@@ -1,4 +1,4 @@
-#===------------- __init__.pxd - dpctl module --------*- Cython -*----------===#
+# ===------------ __init__.pxd - dpctl module --------*- Cython -*----------===#
```
PokhodenkoSA, line 1: It is not necessary to fix number of #.
Possibly @diptorupd you are using old version of black. Use `20.8b1` as pointed in [CONTRIBUTING.md](https://github.com/IntelPython/dpctl/blob/gold/2021/CONTRIBUTING.md). Conda provides version `19.10b0`. Pip provides `20.8b1`.

dpctl/__init__.py
```diff
@@ -1,27 +1,27 @@
...
-## See the License for the specific language governing permissions and
-## limitations under the License.
-##
-##===----------------------------------------------------------------------===##
-##
-## \file
-## This top-level dpctl module.
-##
-##===----------------------------------------------------------------------===##
+#===----------------- __init__.py - dpctl module -------*- Cython -*--------===#
```
PokhodenkoSA: I do not like this line. It duplicates file name and also it lies about Cython. Conclusion: unnecessary support costs.
Lets keep only License at the header of the file and docstring with description of the content.

PokhodenkoSA (approved):

dpctl/__init__.pxd
diptorupd: Thanks!

dpctl/__init__.py
diptorupd: Cython was an obvious mistake, should be Python. I was following the convention LLVM uses, but we can make the change.

PR #214, opened 2020-12-11
Add callback mechanism so that array_finalize can see if obj is Numba meminfo.

DrTodd13:
I was not able to monkeypatch __sycl__usm_array_interface__ into Numba's MemInfoObj as that object appears not to be monkeypatchable.  Instead, I added a callback mechanism so that numba_dppy can register the previous code we had in array_finalize to check for USM MemInfo's.  This way, we still have the same code but without Numba occurring explicitly in dpctl.


PokhodenkoSA (approved):
LGTM. As I understand we need code on side of numba-dppy (or maybe numba) to add checker to the list of checkers via `add_external_usm_checker()`.
Tested locally on Linux - works fine.

PokhodenkoSA:
No tests for `add_external_usm_checker()`.

PR #217, opened 2020-12-11
Update examples

oleksandr-pavlyk:
Updated examples to build with oneAPI and the present state of `dpctl`.

Perhaps a script to build all examples in place and execute/test them could be added as well.


diptorupd (approved):

PokhodenkoSA:
Will merge as examples are not tested in CI.

PR #218, opened 2020-12-11
Use public keyword in _sycl_core.pxd on SyclContext/SyclQueue

oleksandr-pavlyk:
This causes Cython to convert `_sycl_core.pxd`, `_sycl_core.pyx`
into `_sycl_core.h` and `_sycl_core.cpp`.

The header file can be used by external C++/C projects to be aware
of SyclContext and SyclQueue python objects defined by dpctl.

These are PySyclContextObject, PySyclQueueObject, which are instances
of PyObject, and `PySyclContextType`, `PySyclQueueType` which can be used
in `PyObject_TypeCheck` calls, validating that a Python object is
a valid `syclobj` entry in `__sycl_usm_array_interface__`.

### DIFF
--- a/dpctl/_sycl_event.pxd
+++ b/dpctl/_sycl_event.pxd
@@ -23,7 +23,7 @@
 from ._backend cimport DPCTLSyclEventRef
 
 
-cdef class SyclEvent:
+cdef public class SyclEvent [object PySyclEventObject, type PySyclEventType]:
     ''' Wrapper class for a Sycl Event
     '''
     cdef  DPCTLSyclEventRef _event_ref
--- a/dpctl/memory/_memory.pxd
+++ b/dpctl/memory/_memory.pxd
@@ -28,7 +28,7 @@ from .._sycl_device cimport SyclDevice
 from .._sycl_queue cimport SyclQueue
 
 
-cdef class _Memory:
+cdef public class _Memory [object Py_MemoryObject, type Py_MemoryType]:
     cdef DPCTLSyclUSMRef memory_ptr
     cdef Py_ssize_t nbytes
     cdef SyclQueue queue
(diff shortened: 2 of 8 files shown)


oleksandr-pavlyk:
Further decoration of `cdef` functions that allow to retrieve native backend entities as `public` may be needed as well.

dpctl/_sycl_event.pxd
```diff
@@ -23,7 +23,7 @@
 from ._backend cimport DPCTLSyclEventRef
 
 
-cdef class SyclEvent:
+cdef public class SyclEvent [object PySyclEventObject, type PySuclEventType]:
```
diptorupd, line 26: typo `s/PySuclEventType/PySyclEventType/g`

dpctl/memory/_memory.pxd
```diff
@@ -28,7 +28,7 @@ from .._sycl_device cimport SyclDevice
 from .._sycl_queue cimport SyclQueue
 
 
-cdef class _Memory:
+cdef public class _Memory [object Py_MemoryObject, type Py_MemoryType]:
```
diptorupd, line 31: Do we need to expose the `_Memory` class?

diptorupd (commented):
LGTM barring the two comments I left on the PR.

dpctl/memory/_memory.pxd
oleksandr-pavlyk: I thought yes. Mostly for type detection. Testing that an object `obj` is a USM memory buffer can be as simpler as using `PyObject_TypeCheck` with `Py_MemoryType`.

oleksandr-pavlyk:
I thought this change should make it into 0.7.0

PR #224, opened 2020-12-16
Doc improvements

diptorupd:
Polished up the documentation a bit to match changes done as part of the upcoming 0.5.0 release.

### DIFF
--- a/dpctl/__init__.py
+++ b/dpctl/__init__.py
@@ -1,5 +1,3 @@
-# ===---------------- __init__.py - dpctl module -------*- Cython -*--------===#
-#
 #                      Data Parallel Control (dpCtl)
 #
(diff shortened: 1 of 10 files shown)


diptorupd:
@PokhodenkoSA this should be merged prior to tagging the final 0.5.0 release.

dpctl/__init__.py
```diff
@@ -1,5 +1,4 @@
-# ===---------------- __init__.py - dpctl module -------*- Cython -*--------===#
-#
+# ===-----------------------------------------------------------------------===#
 #                      Data Parallel Control (dpCtl)
```
PokhodenkoSA, line 2: This line is also need to support in the future.
I would like to have only license at the header and minimum decorations.

PR #234, opened 2020-12-21
Fix incorrect import in the cython example.

vlad-perevezentsev:


diptorupd:
LGTM! merging as it closes a bug in master.

PR #238, opened 2021-01-11
Add python and cython file coverage

vlad-perevezentsev:
1. Install the `coverage` package (temporary solutions).
```bash
conda install coverage
```
2. Install `dpctl`.
3. Enter the command in the console.
```bash
coverage run -m unittest dpctl.tests
```
4. To show a report on the coverage results.
```bash
coverage report -m
```
5. To create an annotated HTML listings with coverage results.
```bash
coverage html
```
6. The `htmlcov` folder will appear in the root folder of the project.
7. You need to open the file `htmlcov/index.html` to view the report in HTML format.
```bash
coverage erase  # Erase previously collected coverage data.
rm -rf htmlcov  # Delete the report folder.
```


diptorupd:
@vlad-perevezentsev Can you give me the full command of how you include the source files? I tried your steps by modifying Step 3 to `coverage run --source=dpctl -m unittest dpctl.tests`, but that does not show me the Cython coverage. Also, I want to exclude the test cases from the coverage report.

vlad-perevezentsev:
@diptorupd 
To cover cython files:
Enable the Cython plugin in the **.coveragerc** config file of your project:

[run]
plugins = Cython.Coverage

And compile your Cython modules with line tracing support. This can be done by putting the following two comment lines at the top of the modules that you want to trace:
#cython: linetrace=True
#distutils: define_macros=CYTHON_TRACE=1

That's a double opt-in. The first line instructs Cython to generate verbose line tracing code (and thus increase the size of the resulting C/C++ file), and the second line enables this code at C compile time, which will most likely slow down your program. You can also configure both settings globally in your setup.py script.
http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html

All the flags for the coverage command can be added to the **.coveragerc** file.
Add the **--omit=dpctl/tests/*** flag to the **coverage run** command to exclude test cases from coverage reports.

.coveragerc
```diff
@@ -0,0 +1,5 @@
+[run]
+plugins = Cython.Coverage
+branch = True
+source = dpctl
+omit = dpctl/tests/*, *__init__.py
```
PokhodenkoSA, line 5:
```suggestion
omit = dpctl/tests/*, *__init__.py

```

PR #241, opened 2021-01-16
Level zero codegen

diptorupd:
The PR adds an optional feature to enable Level Zero program creation from SPIR-V in dpctl.

- The feature is enabled for Linux for the time being.
- When using this feature `level-zero-dev` headers should be installed.
- Windows support is not enabled at this point.


diptorupd:
@PokhodenkoSA The PR is complete from a code perspective. I changed a function name, but Numba-dppy is the only one impacted. I will post a PR to Numba-dppy after this is merged.

There are two infra things I need help with:

1. How do I add the level zero library as a runtime dependency through setup.py? 
2. The Linux CI server needs the level-zero headers (level-zero-dev) to be installed. 

Once these two changes are done, we should be good to go.

oleksandr-pavlyk:
Level-zero libraries and level-zero include headers must be made available to Python as conda package, if `ze_loaded` is directly linked to `DPCTLSyclInterface`.  oneDAL does not link directly, but likely uses `dlopen` :

```
(base) [13:31:57 ansatnuc04 ~]$ grep ze_loader /localdisk/intel/oneapi/dal/latest/lib/intel64/*.so
Binary file /localdisk/intel/oneapi/dal/latest/lib/intel64/libonedal_dpc.so matches

(base) [13:40:02 ansatnuc04 ~]$ ldd /localdisk/intel/oneapi/dal/latest/lib/intel64/*.so | grep 'ze_loader' | wc
      0       0       0
```

oleksandr-pavlyk:
oneDAL's solution: https://github.com/oneapi-src/oneDAL/blob/1e6f13f10d14a117e235aaf6217f3c4a65469d4a/cpp/daal/include/services/internal/sycl/level_zero_module_sycl.h#L82

diptorupd:
> oneDAL's solution: https://github.com/oneapi-src/oneDAL/blob/1e6f13f10d14a117e235aaf6217f3c4a65469d4a/cpp/daal/include/services/internal/sycl/level_zero_module_sycl.h#L82

The reason this works is that they include the headers almost verbatim inside oneDAL. https://github.com/oneapi-src/oneDAL/blob/1e6f13f10d14a117e235aaf6217f3c4a65469d4a/cpp/daal/include/services/internal/sycl/level_zero_common.h.

Is this the route we want to go?

diptorupd:
Merging since CI passes and no objections :)

PR #261, opened 2021-01-29
Adding __array_function__ so that Numpy calls with dparrays work.

### LINKED ISSUE #206: Implement `__array_function__` for `dpctl.dptensor.numpy_usm_shared.ndarray`
The current `dpctl.dptensor.numpy_usm_shared.ndarray` does not provide an implementation of `__array_function__` that is needed for  `dpctl.dptensor.numpy_usm_shared.ndarray` to be passable to a Numpy function that can in turn invoke our implementation. The initial implementation for the `__array_function__` should most probably replicate the logic currently in the module that includes all `numpy` functions inside the module (https://github.com/IntelPython/dpctl/blob/df60e6a711f74621388d8ef34e6a4aa23079c6be/dpctl/dptensor/numpy_usm_shared.py#L287).

DrTodd13:
Calling numpy.sum with a dparray should work now.  __array_function__ is now implemented.  I changed the way that we determine if a class or function is defined in this file from using eval to getting the module from sys.modules[__name__] and then using hasattr on the module.

Resolves #206

### DIFF
--- a/dpctl/dptensor/numpy_usm_shared.py
+++ b/dpctl/dptensor/numpy_usm_shared.py
@@ -29,6 +29,7 @@ import sys
 import inspect
 import dpctl
 from dpctl.memory import MemoryUSMShared
+import builtins
 
 debug = False
 
@@ -84,7 +85,9 @@ class ndarray(np.ndarray):
             nelems = np.prod(shape)
             dt = np.dtype(dtype)
             isz = dt.itemsize
-            nbytes = int(isz * max(1, nelems))
+            # Have to use builtins.max explicitly since this module will
+            # import numpy's max function.
+            nbytes = int(isz * builtins.max(1, nelems))
             buf = MemoryUSMShared(nbytes)
             new_obj = np.ndarray.__new__(
                 subtype,
@@ -218,7 +221,7 @@ class ndarray(np.ndarray):
 
     # Convert to a NumPy ndarray.
     def as_ndarray(self):
-        return np.copy(self)
+        return np.copy(np.ndarray(self.shape, self.dtype, self))
 
     def __array__(self):
         return self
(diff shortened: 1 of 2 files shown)


dpctl/dptensor/numpy_usm_shared.py
```diff
@@ -271,13 +274,23 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         else:
             return NotImplemented
 
+    def __array_function__(self, func, types, args, kwargs):
+        fname = func.__name__
+        atypes = [type(x) for x in args]
+        has_func = isdef(fname)
+        dprint("__array_function__:", func, fname, type(func), types, atypes, has_func)
+        fargs = [x if not isinstance(x, ndarray) else np.ndarray(x.shape, x.dtype, x) for x in args]
```
oleksandr-pavlyk, line 282: ``np.ndarray(x.shape, x.dtype, x)`` is equivalent to ``x.view(np.ndarray)``, FYI.

```
In [1]: import dpctl

In [2]: import dpctl.dptensor.numpy_usm_shared as nps

In [9]: X = nps.ones(10)

In [10]: Y = X.view(np.ndarray)

In [11]: (type(X), type(Y))

Out[11]: (dpctl.dptensor.numpy_usm_shared.ndarray, numpy.ndarray)
```

oleksandr-pavlyk:
I broke something, as 

```
In [1]: import dpctl, dpctl.dptensor.numpy_usm_shared as nps

In [2]: import numpy as np

In [3]: hasattr(nps, 'empty')
Out[3]: False

In [4]: hasattr(np, 'empty')
Out[4]: True
```

dpctl/dptensor/numpy_usm_shared.py
DrTodd13: Thanks.  Your change of this in the code is fine.

oleksandr-pavlyk (approved):

oleksandr-pavlyk:
@PokhodenkoSA @diptorupd Ok to merge?

diptorupd:
> @PokhodenkoSA @diptorupd Ok to merge?

OK to go if you are satisfied with the changes :)

diptorupd:
@DrTodd13 "Darkened numpy_with_usm_shared.py with black" ... :D haha

PR #268, opened 2021-02-01
Update package wheel

Vyacheslav-Smirnov:
- Fix wheel version
 - Add long description
 - Add wheel install description


Vyacheslav-Smirnov:
Wheel tests are OK
Wheel package format is OK also

PR #276, opened 2021-02-17
change run_test scripts to use pytest.

diptorupd:
Changes the `run_test` scripts to use `pytest` and adds `pytest` as a run dependency to the `meta.yaml`.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -24,11 +24,15 @@ requirements:
         - make  # [unix]
         - ninja  # [win]
         - numpy >=1.17
+        - pytest
+        - pytest-cov
         - wheel
     run:
         - python
         - numpy >=1.17
         - dpcpp_cpp_rt
+        - pytest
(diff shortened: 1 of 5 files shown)


PokhodenkoSA (changes requested):
Packages required only for testing should be listed in
test:
  requires:
    [...]
See conda meta.yaml docs - https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#test-requirements

conda-recipe/meta.yaml
```diff
@@ -24,12 +24,17 @@ requirements:
         - make  # [unix]
         - ninja  # [win]
         - numpy >=1.17
+        - pytest
```
oleksandr-pavlyk, line 27: Like @PokhodenkoSA says, remove these two lines. Having packages listed as test-time dependencies is sufficient.

oleksandr-pavlyk (approved):

PR #315, opened 2021-03-11
Device descriptor image2d3d functions

1e-to:
info::device::image2d_max_width (size_t)
 info::device::image2d_max_height (size_t)
 info::device::image3d_max_width (size_t)
 info::device::image3d_max_height (size_t)
 info::device::image3d_max_depth (size_t)


dpctl-capi/tests/test_sycl_device_interface.cpp
```diff
@@ -275,6 +275,86 @@ TEST_P(TestDPCTLSyclDeviceInterface, Chk_GetPreferredVectorWidthHalf)
     }
 }
 
+TEST_P(TestDPCTLSyclDeviceInterface, Chk_GetImage2dMaxWidth)
+{
+    DPCTLSyclDeviceRef DRef = nullptr;
+    size_t image_2d_max_width = 0;
+    EXPECT_NO_FATAL_FAILURE(DRef = DPCTLDevice_CreateFromSelector(DSRef));
+    if (!DRef)
```
diptorupd, line 283: I changed the test cases in master. I moved the `DPCTLDevice_CreateFromSelector` into the fixture constructor and the `GTEST_SKIP` to the `SetUp()`. Reduces the redundant code in each test case. Please update these test cases in the same way.

dpctl-capi/source/dpctl_sycl_device_interface.cpp
```diff
@@ -388,6 +388,27 @@ bool DPCTLDevice_HasAspect(__dpctl_keep const DPCTLSyclDeviceRef DRef,
     return hasAspect;
 }
 
+#define declmethod(FUNC, NAME)                                                 \
```
diptorupd, line 391: :+1:

diptorupd (commented):
Rebase is needed with master and the C API test cases need to be slightly changed based on what I did in #337. Otherwise good to go.

oleksandr-pavlyk (approved):
Need to be squashed.

PR #318, opened 2021-03-11
Add descriptors: max_read_image_args, max_write_image_args

### LINKED ISSUE #317: Add device descriptors: max_read_image_args, max_write_image_args
Part of the task #286

- [ ] info::device::max_read_image_args (uint32_t)
- [ ] info::device::max_write_image_args (uint32_t)

1e-to:
Closes #317


oleksandr-pavlyk:
I think my comment on another PR applies here as well. 

On the Python side there is no need to ask for these descriptors at the time of `SyclDevice` construction. The query should be made on as-needed bases, i.e. when user queries the instance property.

This will avoid need to special-handle descriptors in case they are not supported for certain devices, like `max_sub_group_size` may not be supported on host device, and SYCL throws an exception.

dpctl-capi/source/dpctl_sycl_device_interface.cpp
```diff
@@ -388,6 +388,24 @@ bool DPCTLDevice_HasAspect(__dpctl_keep const DPCTLSyclDeviceRef DRef,
     return hasAspect;
 }
 
+#define declmethod(FUNC, NAME)                                                 \
```
diptorupd, line 391: You are really liking this style :+1:

dpctl-capi/tests/test_sycl_device_interface.cpp
```diff
@@ -275,6 +275,38 @@ TEST_P(TestDPCTLSyclDeviceInterface, Chk_GetPreferredVectorWidthHalf)
     }
 }
 
+TEST_P(TestDPCTLSyclDeviceInterface, Chk_GetMaxReadImageArgs)
+{
+    DPCTLSyclDeviceRef DRef = nullptr;
+    size_t max_read_image_args = 0;
+    EXPECT_NO_FATAL_FAILURE(DRef = DPCTLDevice_CreateFromSelector(DSRef));
+    if (!DRef)
```
diptorupd, line 283: See my comments in #315. Same apply here.

diptorupd (changes requested):
Rebase with master and update C API tests.

oleksandr-pavlyk (approved):

diptorupd:
@1e-to can you fix the merge conflicts. The macros in #315 and #318 should have different names.

oleksandr-pavlyk (approved):

diptorupd (approved):

PR #323, opened 2021-03-13
Feature/queue manager v3

diptorupd:
Contains several improvements to queue creation and management. Filter strings are now properly supported, in-order queues may be created, SUCL async errors are handled, and a whole bunch of related changes to both C API and Python API.

This PR is a subset and a cleaned up version of #310 and supersedes that PR.


PokhodenkoSA:
@diptorupd please squash PRs next time.

oleksandr-pavlyk:
@PokhodenkoSA I strongly disfavor the practice of enforcing squashing of commits in the PR. Doing so rewrites history (philosophical objection, granted, but development history is lost, commit messages loose their context and relevance, etc.), makes it difficult to share common commits which are not yet in master between branches, and makes it impossible to precision revert one of the faulty commits in the PR later on.

PokhodenkoSA:
@oleksandr-pavlyk I would like to recommend to make PRs smaller. It is nightmare to review PRs with hundreds lines changed and without proper description.
I like the idea of well organized commits targeted to particular goals. But usually commits in PRs are small, sometimes repetitive and have ugly commit messages. So I thing it is easier to achieve well organized commits with squashed PRs.

PokhodenkoSA:
I think forcing merging PRs is not right way too. If keeping original commits is important for some reason it is ok.

PR #324, opened 2021-03-13
Refactor/device properties

diptorupd:
SyclDevice no longer stores attributes related to SYCL device_info types (except for vendor_name, driver_version, device_name, and _max_work_item_sizes). All device_info values are converted to properties. 

Contains everything in #323 and should be merged after #323 is merged.

### DIFF
--- a/dpctl/_sycl_device.pyx
+++ b/dpctl/_sycl_device.pyx
@@ -55,6 +55,7 @@ from ._backend cimport (
     DPCTLSyclDeviceType,
 )
 from . import backend_type, device_type
+from libc.stdint cimport uint32_t
 import warnings
 
 __all__ = [
@@ -91,7 +92,7 @@ cdef class SyclDevice(_SyclDevice):
                 # Create a SyclDevice with an explicit filter string, in
                 # this case the first level_zero gpu device.
                 level_zero_gpu = dpctl.SyclDevice("level_zero:gpu:0"):
-                level_zero_gpu.dump_device_info()
+                level_zero_gpu.print_device_info()
 
         - by calling one of the device selector helper functions:
 
@@ -110,7 +111,7 @@ cdef class SyclDevice(_SyclDevice):
                 # Create a SyclDevice of type GPU based on whatever is returned
                 # by the SYCL `gpu_selector` device selector class.
                 gpu = dpctl.select_gpu_device():
-                gpu.dump_device_info()
+                gpu.print_device_info()
 
     """
     @staticmethod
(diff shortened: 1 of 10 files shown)


diptorupd:
@PokhodenkoSA @1e-to @oleksandr-pavlyk should the attributes of SyclDevice be defined using `property` or `cached_property`

oleksandr-pavlyk:
I was not aware of cached_property, but if it exists, cached property works too.

diptorupd:
> I was not aware of cached_property, but if it exists, cached property works too.

Was referring to https://pypi.org/project/cached-property/. Not sure if it worth the trouble.

PokhodenkoSA:
@diptorupd please don't use 3rd party package for cached properties.
Standard `functools` module already contains cached properties (i.e. [`@lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) new in version 3.2 and [`@cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) new in version 3.8).

I think cached properties are not necessary because our properties don't contain heavy executions.

dpctl/_sycl_device.pyx
```diff
@@ -473,11 +360,125 @@ cdef class SyclDevice(_SyclDevice):
...
+        """ Returns the maximum number of work-items
+            that are permitted in each dimension of the
+            work-group of the nd_range. The minimum
+            value is (1; 1; 1) for devices that are not of
+            device type ``info::device_type::custom``.
+        """
+        max_work_item_sizes = []
+        for n in range(3):
+            max_work_item_sizes.append(self._max_work_item_sizes[n])
+        return tuple(max_work_item_sizes)
```
oleksandr-pavlyk, line 424: Cython will generate more efficient code from doing 
```python
    return tuple(
        self._max_work_item_sizes[0],
        self._max_work_item_sizes[1],
        self._max_work_item_sizes[2]
    )
```

This avoids needless list temporary.

oleksandr-pavlyk (approved):
Other than my prior suggestion, this is good to go, I would like to urge to not put off merging it for longer than absolutely necessary :)

dpctl/_sycl_device.pyx
diptorupd: I made the changes, but am now getting this error

```
  File "/home/diptorupd/Desktop/devel/dpctl/dpctl/tests/test_sycl_queue.py", line 326, in test_valid_filter_selectors
    check(device)
  File "/home/diptorupd/Desktop/devel/dpctl/dpctl/tests/test_sycl_queue.py", line 66, in check_get_max_work_item_sizes
    max_work_item_sizes = device.max_work_item_sizes
  File "dpctl/_sycl_device.pyx", line 421, in dpctl._sycl_device.SyclDevice.max_work_item_sizes.__get__
    return tuple(
TypeError: tuple expected at most 1 argument, got 3

```

oleksandr-pavlyk (approved):

PR #327, opened 2021-03-18
Merge wheel changes from gold/2021

Vyacheslav-Smirnov:
* Update packages wheel deps

* Make wheel version public-like

* Fix package version

* Remove install_requires

* Append cflags and ldflags from env to build

* Fix issues

* Fix issue with external flags

* Add wheel install description to readme

* Fix formatting

* Fix readme format

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,25 @@ For development:
 python setup.py develop
 ```
 
+Install Wheel Package from Pypi
+==================================
+1. Install dpCtl
+```cmd
+python -m pip install --index-url https://pypi.anaconda.org/intel/simple -extra-index-url https://pypi.org/simple dpctl
+```
+Note: dpCtl wheel package is placed on Pypi, but some of its dependencies (like Intel numpy) are in Anaconda Cloud.
+That is why install command requires additional intel Pypi channel from Anaconda Cloud.
+
+2. Set path to Performance Libraries in case of using venv or system Python:
+On Linux:
+```cmd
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk:
Looks good overall. My only question is what necessitated the need to read `CFLAGS` and `LDFLAGS` from the environment?

Vyacheslav-Smirnov:
@oleksandr-pavlyk, this is required because we have global conda-build config for all built packages. It defines there variables "CFLAGS and LDFLAGS" + adds RPATH to Performance Libraries thus it is good to reuse them rather define these variables in all packages separately.

Vyacheslav-Smirnov:
@PokhodenkoSA , are you OK with the changes?

README.md
```diff
@@ -65,6 +65,25 @@ For development:
...
 ```
 
+Install Wheel Package from Pypi
+==================================
+1. Install dpCtl
+```cmd
+python -m pip install --index-url https://pypi.anaconda.org/intel/simple -extra-index-url https://pypi.org/simple dpctl
+```
+Note: dpCtl wheel package is placed on Pypi, but some of its dependencies (like Intel numpy) are in Anaconda Cloud.
+That is why install command requires additional intel Pypi channel from Anaconda Cloud.
```
PokhodenkoSA, line 75:
```suggestion
That is why install command requires additional `intel` Pypi channel from Anaconda Cloud.
```

PokhodenkoSA (approved):
Merge as `gold/2021` should be more stable than `master`.

PR #338, opened 2021-03-28
Remove workarounds to make device equality work.

diptorupd:
- dpcpp 2021.2 fixed device equality in the dpcpp runtime and we can
    now remove several workarounds in dpctl:

    - DPCTLDeviceMgr_GetDeviceAndContextPair is now renamed as
      DPCTLDeviceMgr_GetCachedContext and only returns a
      DPCTLSyclContextRef instead of a pair of DPCTLSyclContextRef
      and DPCTLSyclDeviceRef.
    - Remove the DPCTLDeviceMgr_AreEq function and related internal
      helper functions for hashing SYCL devices.
    - Remove DeviceWrapper class inside dpctl_sycl_device_manager.cpp.
    - Remove DPCTL_DeviceAndContextPair type form C API.

  - Make equivalent changes to the Python API.

### DIFF
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -25,12 +25,12 @@ from ._backend cimport (
     _arg_data_type,
     _backend_type,
     _queue_property_type,
-    DPCTL_DeviceAndContextPair,
     DPCTLContext_Delete,
     DPCTLDefaultSelector_Create,
     DPCTLDevice_CreateFromSelector,
-    DPCTLDeviceMgr_GetDeviceAndContextPair,
+    DPCTLDeviceMgr_GetCachedContext,
     DPCTLDeviceSelector_Delete,
+    DPCTLDevice_Copy,
     DPCTLDevice_Delete,
     DPCTLFilterSelector_Create,
     DPCTLQueue_AreEq,
@@ -189,7 +189,7 @@ cdef class SyclQueue:
                 status = self._init_queue_from_filter_string(
                     filter_c_str, props)
             elif isinstance(arg, SyclDevice):
-                status = self._init_queue_from_device(<SyclDevice> arg, props)
+                status = self._init_queue_from_device(<SyclDevice>arg, props)
             else:
                 raise TypeError(
                     "Positional argument {} is not a filter string or a "
@@ -259,13 +259,11 @@ cdef class SyclQueue:
         """
         cdef DPCTLSyclContextRef CRef
         cdef DPCTLSyclQueueRef QRef
-        cdef DPCTL_DeviceAndContextPair dev_ctx
 
-        dev_ctx = DPCTLDeviceMgr_GetDeviceAndContextPair(DRef)
-        if (dev_ctx.CRef is NULL) or (dev_ctx.DRef is NULL):
+        CRef = DPCTLDeviceMgr_GetCachedContext(DRef)
+        if (CRef is NULL):
+            DPCTLDevice_Delete(DRef)
             return -3
-        DRef = dev_ctx.DRef
-        CRef = dev_ctx.CRef
         QRef = DPCTLQueue_Create(
             CRef,
             DRef,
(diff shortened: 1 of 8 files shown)


diptorupd:
@oleksandr-pavlyk dpcpp 2021.2 has the device equality fixes that makes my workaround redundant. This PR removed the `DPCTL_DeviceAndContextPair`. However, now we have a hard dependency on dpcpp version 2021.2 as anything lower will cause things to break. I have another PR coming up where I add a way to specify the dpcpp version requirement in the CMakeList.txt.

dpctl/_sycl_queue.pyx
```diff
@@ -259,22 +259,20 @@ cdef class SyclQueue:
         """
         cdef DPCTLSyclContextRef CRef
         cdef DPCTLSyclQueueRef QRef
-        cdef DPCTL_DeviceAndContextPair dev_ctx
 
-        dev_ctx = DPCTLDeviceMgr_GetDeviceAndContextPair(DRef)
-        if (dev_ctx.CRef is NULL) or (dev_ctx.DRef is NULL):
+        CRef = DPCTLDeviceMgr_GetCachedContext(DRef)
+        if (CRef is NULL):
+            DPCTLDevice_Delete(DRef)
```
oleksandr-pavlyk, line 265: `DRef` is the input argument. We should not be deleting it.

oleksandr-pavlyk: If the argument device is a sub-device the look-up will fail. I thought we discussed that it would be better to create a new one instead.

oleksandr-pavlyk (changes requested):
The change in memory management in Cython code is not good, and best be reverted.

dpctl/_sycl_queue.pyx
diptorupd: `_init_queue_from_DPCTLSyclDeviceRef` is returning an error code if the context look up fails. I assume the logic to handle creation of a context for sub-devices is not yet implemented? Do your upcoming changes to `SyclContext` have that change?

As for freeing `DRef`, in the current code it does not cause an issue. Does it? Are you saying from a design style perspective we should let caller free the object? 

Since, `_init_queue_from_DPCTLSyclDeviceRef` is an internal helper that is invoked only by other internal helpers: `_init_queue_from_filter_string`, `_init_queue_from_device`, and `_init_queue_default`, freeing the `DRef` in `_init_queue_from_DPCTLSyclDeviceRef` when we know queue creation failed is fine IMO. But, I can change the code to free it in the callers on returning from  `_init_queue_from_DPCTLSyclDeviceRef` at the cost of adding a check on the return code. Do note that I have a copy in `_init_queue_from_device`, the only place where deleting the `DRef` can have an external side-effect.

diptorupd: > As for freeing `DRef`, in the current code it does not cause an issue. Does it? Are you saying from a design style perspective we should let caller free the object?

I see the issue now. `DRef` will leak when queue creation succeeded. I will fix that.

diptorupd: > `_init_queue_from_DPCTLSyclDeviceRef` is returning an error code if the context look up fails. I assume the logic to handle creation of a context for sub-devices is not yet implemented? Do your upcoming changes to `SyclContext` have that change?

Ok #334 has the changes.

diptorupd: @oleksandr-pavlyk  in `_init_queue_from_device` we are taking the `DRef` from the `SyclDevice` argument and storing it in the `SyclQueue` if queue creation succeeded. I think we need a `DPCTLDevice_Copy` as I have done here. What happens when the `SyclDevice` is deleted thus deallocating the `DPCTLSyclDeviceRef`?

oleksandr-pavlyk (approved):

oleksandr-pavlyk:
> The change in memory management in Cython code is not good, and best be reverted.

The changes implement convention that most `_init_*` functions consuming an opaque reference would delete it. Callers need to call pass a copy if they intend to keep the object alive.

PR #319, opened 2021-03-11
Changes to build scripts to enable building dpctl with a custom DPC++.

diptorupd:

PR #339, opened 2021-03-28
Add a version requirement for dpcpp inside CMakeList.txt.

diptorupd:


diptorupd:
@oleksandr-pavlyk There are two more things that we need to properly address before this is ready to be merged:

1. The `dpcpp_cpp_rt` package used at runtime and the dpcpp compiler used at runtime should be made to match.
2. #319 (using a custom dpcpp) will not work with this PR as open source dpcpp does not use same version scheme as oneAPI dpcpp.

oleksandr-pavlyk:
The cmake 3.20 supports oneAPI compiler as IntelLLVM, see [New In 3.20#Compilers](https://cmake.org/cmake/help/latest/release/3.20.html#compilers).

Perhaps this could be leveraged to support both custom build of open-source LLVM with SYCL support and dpcpp.

We can also institute a shallow dpctl init-time check that the DCPPP run-time meets functional requirements, i.e. can ascertain that two devices created with the same selector are equal, and raise RuntimeError otherwise.

oleksandr-pavlyk (approved):

PR #347, opened 2021-03-31
Fixed signature of SyclDevice._init_helper

oleksandr-pavlyk:
From #334, changes to _sycl_device.pyx and .pxd.

### DIFF
--- a/dpctl/_sycl_device.pyx
+++ b/dpctl/_sycl_device.pyx
@@ -123,7 +123,7 @@ cdef class SyclDevice(_SyclDevice):
 
     """
     @staticmethod
-    cdef void _init_helper(SyclDevice device, DPCTLSyclDeviceRef DRef):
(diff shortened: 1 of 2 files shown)


dpctl/_sycl_device.pyx
```diff
@@ -221,7 +222,7 @@ cdef class SyclDevice(_SyclDevice):
             int: The address of the DPCTLSyclDeviceRef object used to create
             this SyclDevice cast to a size_t.
         """
-        return int(<size_t>self._device_ref)
+        return <size_t>self._device_ref
```
diptorupd, line 225: should the cdef signature be `size_t`?

diptorupd: That's what I thought.

diptorupd (approved):

PR #109, opened 2020-10-06
Run C++ tests during building

1e-to:
This PR for testing the running C++ tests in conda recipe

Closes https://github.com/IntelPython/dpctl/issues/98

PR #355, opened 2021-04-02
Run C API tests as part of setup.py on Linux

diptorupd:


scripts/build_backend.py
```diff
@@ -85,11 +85,13 @@
             "-DDPCPP_INSTALL_DIR=" + DPCPP_ROOT,
             "-DCMAKE_C_COMPILER:PATH=" + os.path.join(DPCPP_ROOT, "bin", "clang"),
             "-DCMAKE_CXX_COMPILER:PATH=" + os.path.join(DPCPP_ROOT, "bin", "dpcpp"),
+            "-DDPCTL_BUILD_CAPI_TESTS=ON",
             "-DDPCTL_ENABLE_LO_PROGRAM_CREATION=ON",
             backends,
         ]
         subprocess.check_call(cmake_args, stderr=subprocess.STDOUT, shell=False)
         subprocess.check_call(["make", "V=1", "-j", "4"])
+        subprocess.check_call(["make", "check"])
```
oleksandr-pavlyk, line 94: Just a note: `check_call` raises an exception if the subprocess returns with a non-zero return code, so any failures in C-API test suite will fail the Python build as desired.

oleksandr-pavlyk:
I think it is a good idea to add support for CLI option to `develop` command that could override building/running of C-API tests. Otherwise Python-API development becomes too time-consuming.

diptorupd:
Closing this PR as #109 and #90 are already trying to address the same issue.

PR #359, opened 2021-04-02
Fix tests which fail if GPU is not available (only CPU is available)

PokhodenkoSA:
This PR contains several fixes for tests which should be skipped if GPU is not available but they was not skipped.

1. `dpctl.SyclContext("gpu")` raises `ValueError`. It seems that catching `dpctl.SyclQueueCreationError` is copy/past error.
2. In `test_context_can_be_used_in_queue2` exception was not caught.
3. `TestGetCurrentDevice` uses context `opencl:gpu:0` in several tests so it should be skipped if GPU is not available.

### DIFF
--- a/dpctl/tests/test_sycl_context.py
+++ b/dpctl/tests/test_sycl_context.py
@@ -356,7 +356,7 @@ def test_context_equals():
     try:
         ctx1 = dpctl.SyclContext("gpu")
         ctx0 = dpctl.SyclContext("gpu")
-    except dpctl.SyclQueueCreationError:
+    except ValueError:
         pytest.skip()
     assert ctx0.equals(ctx1)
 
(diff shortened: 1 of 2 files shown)


dpctl/tests/test_sycl_context.py
```diff
@@ -356,7 +356,7 @@ def test_context_equals():
     try:
         ctx1 = dpctl.SyclContext("gpu")
         ctx0 = dpctl.SyclContext("gpu")
-    except dpctl.SyclQueueCreationError:
+    except ValueError:
```
oleksandr-pavlyk, line 359: This is good. I think we may as well introduce `SyclContextCreationError` and use that, for symmetry,

oleksandr-pavlyk (approved):

PokhodenkoSA:
There are also a lot of failed tests when CPU is not available too.

diptorupd:
@PokhodenkoSA @oleksandr-pavlyk We can use `SYCL_DEVICE_FILTER` (refer https://github.com/intel/llvm/blob/sycl/sycl/doc/EnvironmentVariables.md) to simulate testing with only a specific type of device. Can we setup multiple CI runs where we explicitly select a specific type of device and turn off other types using the environment variable?

PR #370, opened 2021-04-06
Feature/sycl queue in order

oleksandr-pavlyk:
```
Python 3.7.9 (default, Mar 10 2021, 05:18:00)
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import dpctl

In [2]: q = dpctl.SyclQueue("gpu", property=("enable_profiling", "in_order"))

In [3]: q
Out[3]: <dpctl.SyclQueue at 0x7f2084fc0500, property=in_order>

In [4]: q.sycl_device
Out[4]: <dpctl.SyclDevice [backend_type.level_zero, device_type.gpu,  Intel(R) UHD Graphics [0x9bca]] at 0x7f20858ece30>

In [5]: q.is_in_order
Out[5]: True

```

### DIFF
--- a/dpctl-capi/source/dpctl_sycl_queue_interface.cpp
+++ b/dpctl-capi/source/dpctl_sycl_queue_interface.cpp
@@ -119,14 +119,16 @@ bool set_kernel_arg(handler &cgh,
 std::unique_ptr<property_list> create_property_list(int properties)
 {
     std::unique_ptr<property_list> propList;
-    if (properties & (DPCTL_ENABLE_PROFILING | DPCTL_IN_ORDER)) {
-        propList = std::make_unique<property_list>(
-            sycl::property::queue::enable_profiling(),
-            sycl::property::queue::in_order());
-    }
-    else if (properties & DPCTL_ENABLE_PROFILING) {
-        propList = std::make_unique<property_list>(
-            sycl::property::queue::enable_profiling());
+    if (properties & DPCTL_ENABLE_PROFILING) {
+        if (properties & DPCTL_IN_ORDER) {
+            propList = std::make_unique<property_list>(
+                sycl::property::queue::enable_profiling(),
+                sycl::property::queue::in_order());
+        }
+        else {
+            propList = std::make_unique<property_list>(
+                sycl::property::queue::enable_profiling());
+        }
     }
     else if (properties & DPCTL_IN_ORDER) {
         propList =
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk:
This does not seem right:

```python
In [1]: import dpctl

In [2]: dpctl.SyclQueue("cpu", property="enable_profiling")
enable_profiling
2
Out[2]: <dpctl.SyclQueue at 0x7f70546b6b40, property=in_order>

In [3]: Out[2].is_in_order
Out[3]: True
```

Debugging prints are mine. It appears that there is an issue with logic in `create_property_list`. I have pushed a fix.

dpctl-capi/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -119,14 +119,16 @@ bool set_kernel_arg(handler &cgh,
 std::unique_ptr<property_list> create_property_list(int properties)
 {
     std::unique_ptr<property_list> propList;
-    if (properties & (DPCTL_ENABLE_PROFILING | DPCTL_IN_ORDER)) {
```
diptorupd, line 122: oops. Thanks for testing and catching it.

dpctl-capi/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -450,7 +452,8 @@ void DPCTLQueue_Wait(__dpctl_keep DPCTLSyclQueueRef QRef)
     // \todo what happens if the QRef is null or a pointer to a valid sycl
     // queue
     auto SyclQueue = unwrap(QRef);
-    SyclQueue->wait();
+    if (SyclQueue)
```
diptorupd, line 455: 👍

dpctl-capi/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -459,17 +462,21 @@ void DPCTLQueue_Memcpy(__dpctl_keep const DPCTLSyclQueueRef QRef,
                        size_t Count)
 {
     auto Q = unwrap(QRef);
-    auto event = Q->memcpy(Dest, Src, Count);
-    event.wait();
+    if (Q) {
```
diptorupd, line 465: 👍 

Should an error be recorded if `Q` is null?

diptorupd (approved):

dpctl-capi/source/dpctl_sycl_queue_interface.cpp
oleksandr-pavlyk: Maybe, but we return `void`, so I figured I'd just prevent the crash. 

This can be addressed due to the pass of error handling improvements.

PR #341, opened 2021-03-31
Doc improvement

diptorupd:
Various improvements to dpctl documentation website. 
 - We now use doxyrest to generate C API docs.
 - A new QuickStart guide
 - Reorganizing the docs directory
 - Change existing file names.

PR #372, opened 2021-04-07
Various doc string updates to C API files.

diptorupd:
Separated out doxygen fixes from #341


oleksandr-pavlyk:
This all looks fine @diptorupd, the rendered docs are needed though to see the effect of these changes. 

Since these modify C-API files, I think we should first merge my PRs which modify/add-to C-API functions, and then make sure that these changes are applies to the newly added functions too.

diptorupd:
@oleksandr-pavlyk I rebased it on top of master. We should `squash and merge` the PR.

PR #378, opened 2021-04-08
Fixes names

### LINKED ISSUE #284: Have consistent naming of functions for SYCL, C, and Python API functions
We have some inconsistency across the naming of functions in C API and Python. *E.g.*, 
  
| SYCL Function                                       | dpCtl C-API function                | Python API function | 
| --------------------------------------------------   | ---------------------------------------- | -------------------------- | 
| `get_info<info::device::driver_version>` | `DPCTLDevice_GetDriverInfo` | `get_driver_version` |

Each function in C API should be named as `DPCTLXXXX_YYYY()`, where `XXXX` is the SYCL class name in camel case and `YYYY` is a meaningful approximation of the SYCL function name. The function name should be close to the SYCL function name with changes as needed to handle templates. For example, the `device` class `get_info` functions should be converted into `DPCTLDevice_GetZZZZ`, where `ZZZZ` is the descriptor information type. For the above example of `get_info<info::device::driver_version` the correct C API name should be `DPCTLDevice_GetDriverVersion`.

1e-to:
Closes #284

### DIFF
--- a/dpctl-capi/include/dpctl_sycl_device_interface.h
+++ b/dpctl-capi/include/dpctl_sycl_device_interface.h
@@ -126,7 +126,7 @@ bool DPCTLDevice_IsHost(__dpctl_keep const DPCTLSyclDeviceRef DRef);
  */
 DPCTL_API
 DPCTLSyclBackendType
-DPCTLDevice_GetBackendVersion(__dpctl_keep const DPCTLSyclDeviceRef DRef);
+DPCTLDevice_GetBackend(__dpctl_keep const DPCTLSyclDeviceRef DRef);
 
 /*!
  * @brief Returns the DPCTLSyclDeviceType enum value for the DPCTLSyclDeviceRef
--- a/dpctl-capi/include/dpctl_sycl_device_interface.h
+++ b/dpctl-capi/include/dpctl_sycl_device_interface.h
@@ -126,7 +126,7 @@ bool DPCTLDevice_IsHost(__dpctl_keep const DPCTLSyclDeviceRef DRef);
  */
 DPCTL_API
 DPCTLSyclBackendType
-DPCTLDevice_GetBackend(__dpctl_keep const DPCTLSyclDeviceRef DRef);
+DPCTLDevice_GetBackendVersion(__dpctl_keep const DPCTLSyclDeviceRef DRef);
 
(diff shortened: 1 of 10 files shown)


dpctl-capi/include/dpctl_sycl_device_interface.h
```diff
@@ -126,7 +126,7 @@ bool DPCTLDevice_IsHost(__dpctl_keep const DPCTLSyclDeviceRef DRef);
  */
 DPCTL_API
 DPCTLSyclBackendType
-DPCTLDevice_GetBackend(__dpctl_keep const DPCTLSyclDeviceRef DRef);
+DPCTLDevice_GetBackendVersion(__dpctl_keep const DPCTLSyclDeviceRef DRef);
```
diptorupd, line 129: This function does not need renaming. The SYCL function is called `get_backend` too.

oleksandr-pavlyk (changes requested):
@1e-to Please revert the change to `DPCTLDevice_GetBackend` -> `DPCTLDevice_GetBackendVersion` per @diptorupd  comment.

examples/cython/sycl_buffer/bench.py
```diff
@@ -27,15 +27,15 @@
 dpctl.set_global_queue("opencl:cpu")
 print(
     "SYCL({}) result: {}".format(
-        dpctl.get_current_queue().get_sycl_device().get_device_name(),
+        dpctl.get_current_queue().get_sycl_device().name,
```
oleksandr-pavlyk, line 30: Perhaps it should be `dpctl.get_current_queue().sycl_device.name` to simplify?

oleksandr-pavlyk (dismissed):
This is looking good. Thank you @1e-to

oleksandr-pavlyk (approved):

dpctl-capi/include/dpctl_sycl_device_interface.h
diptorupd: @1e-to can make this change? I will like to have this change in 0.7.0.

diptorupd: Oops missed your change. All set.

diptorupd (approved):

PR #385, opened 2021-04-14
Docstring fixes

diptorupd:
Final cleanups and improvements to docstrings before releasing 0.7.0.


diptorupd:
@oleksandr-pavlyk I polished up the docstrings of `SyclContext`. The rendered docs are published at https://intelpython.github.io/dpctl/latest/docfiles/dpctl_pyapi/SyclContext.html. Would be great if we could polish up more of the classes. 

Also, you can generate the Sphinx docs just for the Python API omitting the C API and avoiding the issues you were facing with doxyrest. Here's the "how-to": https://github.com/diptorupd/dpctl/blob/doc-improvement/docs/README.md#build-only-sphinx-for-python-api.

oleksandr-pavlyk (approved):
This is great @diptorupd 

I will try to do the same to `SyclQueue`.

PR #387, opened 2021-04-14
Feature/dcpp compiler directive support

AndresGuzman-Ballen:
Add support to using DPC++ compiler conda packages with the conda compiler directives ({{ compiler('dpcpp') }})

### DIFF
--- a/conda-recipe/build.sh
+++ b/conda-recipe/build.sh
@@ -1,14 +1,5 @@
 #!/bin/bash
 
-# We need dpcpp to compile dpctl_sycl_interface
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk:
dpctl uses different compiler to build native Python extensions (should use the same compiler the CPython was built with) than the compiler used to build the backend (should be DPC++ toolchain, either coming from oneAPI or built from sources at github.com/intel/llvm).

I suppose this means conda ought be able to install compilers into environment, but only activate one. 

Path to SYC compiler should be indicated with `--sycl-compiler-prefix`.

PR #390, opened 2021-04-14
Improve examples

oleksandr-pavlyk:
This change modified Cython examples, as well as Python examples.

It add two files, `examples/python/device_selection.py` and `examples/python/subdevices.py`.

Please look them over, and suggest ideas for further examples.


examples/python/device_selection.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Device created can be influenced by environment variable
+    SYCL_DEVICE_FILTER, which determines SYCL devices seen by the
+    SYCL runtime.
+    """
+    d = dpctl.SyclDevice("gpu,cpu")
+    print("Selected " + ("GPU" if d.is_gpu else "CPU") + " device")
+
+
+def custom_select_device():
```
diptorupd, line 78: very nice.

examples/python/device_selection.py
```diff
@@ -0,0 +1,97 @@
...
+    # choose the device with highest default_selector score
+    max_score = 0
+    selected_dev = None
+    for d in devs:
+        if d.default_selector_score > max_score:
+            max_score = d.default_selector_score
+            selected_dev = d
+    if selected_dev:
+        print_device(selected_dev)
+    return selected_dev
```
diptorupd, line 97: We need a `if __name__ == "__main__":` to actually execute these functions.

diptorupd (changes requested):
The device selection example functions are not being run anywhere in the example. Please fix that. Except for that we are golden.

examples/python/device_selection.py
oleksandr-pavlyk: done.

diptorupd (approved):

PR #392, opened 2021-04-16
Fix issue 189

### LINKED ISSUE #189: dpctl.memory.MemoryUSMShared(np.int64(16)) should work
```
In [4]: dpctl.memory.MemoryUSMShared(np.int64(16384))
<ipython-input-4-abef76d3355f> in <module>()
----> 1 dpctl.memory.MemoryUSMShared(np.int64(16384))

/localdisk/work/opavlyk/repos/dpctl/dpctl/memory/_memory.pyx in dpctl.memory._memory.MemoryUSMShared.__cinit__()

/localdisk/work/opavlyk/repos/dpctl/dpctl/memory/_memory.pyx in dpctl.memory._memory._Memory._cinit_other()

ValueError: Argument 16384 does not expose `__sycl_usm_array_interface__`.
```

```
In [5]: dpctl.memory.MemoryUSMShared(16384)
Out[5]: <Intel(R) USM allocated memory block of 16384 bytes at 0x55f2202e3000>

```

oleksandr-pavlyk:
Closes #189 

When type-checking input, check for the input to be instance
of abtract class `numbers.Integral`, rather than class `int`

Modified test suite to not expect the test from the ticket
to fail.

### DIFF
--- a/dpctl/memory/_memory.pyx
+++ b/dpctl/memory/_memory.pyx
@@ -34,6 +34,7 @@ from cpython.bytes cimport PyBytes_AS_STRING, PyBytes_FromStringAndSize
 from cpython cimport pycapsule
 
 import numpy as np
+import numbers
 
 __all__ = [
     "MemoryUSMShared",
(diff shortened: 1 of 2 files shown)


dpctl/memory/_memory.pyx
```diff
@@ -34,6 +34,7 @@ from cpython.bytes cimport PyBytes_AS_STRING, PyBytes_FromStringAndSize
 from cpython cimport pycapsule
 
 import numpy as np
```
diptorupd, line 36: Should this be `cimport numpy as cnp`?

oleksandr-pavlyk: No, this file only uses Python API of NumPy.

diptorupd: makes sense.

PR #393, opened 2021-04-16
Edit to _SyclDevice.__doc__

oleksandr-pavlyk:
Use double quotes in honor of everything `black .`

Use the same verbiage as in .pyx file

### DIFF
--- a/dpctl/_sycl_device.pxd
+++ b/dpctl/_sycl_device.pxd
@@ -29,8 +29,8 @@ from libcpp cimport bool as cpp_bool
 
 
 cdef class _SyclDevice:
-    ''' Wrapper data owner class for a Sycl Device
-    '''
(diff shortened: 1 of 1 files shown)


dpctl/_sycl_device.pxd
```diff
@@ -29,8 +29,8 @@ from libcpp cimport bool as cpp_bool
 
 
 cdef class _SyclDevice:
-    ''' Wrapper data owner class for a Sycl Device
-    '''
+    """ A helper data-owner class to abstract a cl::sycl::device instance.
```
diptorupd, line 32: Although this docstring is not added to Sphinx generated docs, it is a good practice to add "``cl::sycl::device``". That helps render it properly in SPhinx.

oleksandr-pavlyk: Done.

PR #400, opened 2021-04-20
Implemented support for constructing MemoryUSM* from object with __sycl_usm_array_interface__ when array-info is not contiguous

### LINKED ISSUE #396: MemoryUSM* constructors should be able to handle objects with `__sycl_usm_array_interface__` and non-trivial strides.
For Python object `suai_obj` exposing `__sycl_usm_array_interface__` the call `dpctl.memory.MemoryUSMShared(suai_obj)` currently requires that `strides` entry in the interface dictionary be [empty](https://github.com/IntelPython/dpctl/blob/master/dpctl/memory/_memory.pyx#L166) (implying C-contiguous data). 

This issue is to request that support for arbitrary `'strides'` and `'offset'` settings be added.

The `dpctl.memory` objects will then assume ownership of memory with 
   - `pointer = char_ptr + itemsize * minimum(offset + dot(element_strides, indexes), indexes for shape)` 
   - `nbytes = itemsize*(1 + maximum(offset + dot(element_strides, indexes), indexes for shape) - minimum(offset + dot(element_strides, indexes), indexes for shape)`.

For example, for `sua_obj`  which is a (2,2) view of (4, 5) array corresponding to starred elements of the table below:

|  | 0 | 1 | 2 | 3 | 4
| --- | --- | --- | --- | --- | --- |
|  0  |       |      |      |       |      |
| 1   |       |  *   |      |  *  |      |
| 2   |       |      |      |     |      |
| 3   |       |  *   |      |  *  |      |

The interface dictionary will have `offset = 6` and `strides=(10, 2)` and `data=(address_of(ar[0,0]), WRITEABLE)`.

The `MemoryUSMShared(sual_obj)` will map memory from `a[1,1]` to `a[3, 3]` including 13 elements, namely `{a[1,1], a[1,2], a[1, 3], a[1,4], a[2, 0], a[2, 1], a[2,2], a[2, 3], a[2,4], a[3,0], a[3, 1], a[3,2], a[3,3]}`.

oleksandr-pavlyk:
Closes #396 

This PR permits constructor calls like `dpctl.memory.MemoryUSMShared( obj_with_suai )` to work with Python object whose `__sycl_usm_array_interface__` dictionary specifies non-contiguous strides, and non-byte-size type-string, used to address elements of a contiguous  block of memory. 

I.e. index tuple `(i0, i1, ...)` would address element `offset + strides[0]*i0 + strides[1]*i1 + ...`  in C-array of dimension `shape[0]*shape[1]*...`.

In this case type-less USM memory buffer created addresses memory from the minimal index to maxima index (inclusive) generated by enumeration of all  displacements  `offset + strides[0]*i0 + strides[1]*i1 + ...`  when iterating over all possible tuple of indexes in determined by the `shape`.

Example:

```python
import numpy as np
import dpctl
import dpctl.memory as dpmem

class View:
    def __init__(self, buffer, shape, strides, offset):
        self.buffer=buffer
        self.shape=shape
        self.strides=strides
        self.offset = offset

    @property
    def __sycl_usm_array_interface__(self):
        sua_iface = self.buffer.__sycl_usm_array_interface__
        sua_iface['offset'] = self.offset
        sua_iface['shape'] = self.shape
        sua_iface['strides'] = self.strides
        return sua_iface

# Allocate 32 bytes, and fill them with canary numbers
buf = dpmem.MemoryUSMShared(32)
buf.copy_from_host(np.full((32,), 77, dtype='|u1'))

# View of 10 bytes, starting with byte 8, and going every other byte
v = View(buf, (10,), strides=(2,), offset=8)

# Create USMShared buffer from the view 
# spanning from the lowest byte, offset=8 to highest byte, offset=8 +(10-1)*2
buf2 = dpmem.MemoryUSMShared(v)
# That is total of 19 bytes, (counting the last one too)
assert buf2.nbytes == 19
# populate this buffer with different values
buf2.copy_from_host(np.full((19,), 1, dtype="|u1"))
# copy content of the parent buffer to host
res = buf.copy_to_host()
del buf
del buf2
print(res)
```

Executing this prints

```
[77 77 77 77 77 77 77 77  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
  1  1  1 77 77 77 77 77]
```

### DIFF
--- a/dpctl/memory/_sycl_usm_array_interface_utils.pxi
+++ b/dpctl/memory/_sycl_usm_array_interface_utils.pxi
@@ -1,0 +1,200 @@
+
+
+cdef bint _valid_usm_ptr_and_context(DPCTLSyclUSMRef ptr, SyclContext ctx):
+    usm_type = _Memory.get_pointer_type(ptr, ctx)
+    return usm_type in (b'shared', b'device', b'host')
+
+
+cdef DPCTLSyclQueueRef _queue_ref_copy_from_SyclQueue(
+    DPCTLSyclUSMRef ptr, SyclQueue q):
+    """ Check that USM ptr is consistent with SYCL context in the queue,
+    and return a copy of QueueRef if so, or NULL otherwise.
+    """
+    cdef SyclContext ctx = q.get_sycl_context()
+    if (_valid_usm_ptr_and_context(ptr, ctx)):
+        return DPCTLQueue_Copy(q.get_queue_ref())
+    else:
+        return NULL
+
+
+cdef DPCTLSyclQueueRef _queue_ref_copy_from_USMRef_and_SyclContext(
+    DPCTLSyclUSMRef ptr, SyclContext ctx):
+    """ Obtain device from pointer and sycl context, use
+        context and device to create a queue from which this memory
+        can be accessible.
+    """
+    cdef SyclDevice dev = _Memory.get_pointer_device(ptr, ctx)
+    cdef DPCTLSyclContextRef CRef = ctx.get_context_ref()
+    cdef DPCTLSyclDeviceRef DRef = dev.get_device_ref()
+    return DPCTLQueue_Create(CRef, DRef, NULL, 0)
+
+
+cdef DPCTLSyclQueueRef get_queue_ref_from_ptr_and_syclobj(
+    DPCTLSyclUSMRef ptr, object syclobj):
+    """ Constructs queue from pointer and syclobject from
+        __sycl_usm_array_interface__
+    """
+    cdef SyclContext ctx
+    if type(syclobj) is SyclQueue:
+        return _queue_ref_copy_from_SyclQueue(ptr, <SyclQueue> syclobj)
+    elif type(syclobj) is SyclContext:
+        ctx = <SyclContext>syclobj
+        return _queue_ref_copy_from_USMRef_and_SyclContext(ptr, ctx)
+    elif type(syclobj) is str:
+        q = SyclQueue(syclobj)
+        return _queue_ref_copy_from_SyclQueue(ptr, <SyclQueue> q)
+    elif pycapsule.PyCapsule_IsValid(syclobj, "SyclQueueRef"):
+        q = SyclQueue(syclobj)
+        return _queue_ref_copy_from_SyclQueue(ptr, <SyclQueue> q)
+    elif pycapsule.PyCapsule_IsValid(syclobj, "SyclContextRef"):
+        ctx = <SyclContext>SyclContext(syclobj)
+        return _queue_ref_copy_from_USMRef_and_SyclContext(ptr, ctx)
+    elif hasattr(syclobj, '_get_capsule'):
+        cap = syclobj._get_capsule()
+        if pycapsule.PyCapsule_IsValid(cap, "SyclQueueRef"):
+            q = SyclQueue(cap)
+            return _queue_ref_copy_from_SyclQueue(ptr, <SyclQueue> q)
+        elif pycapsule.PyCapsule_IsValid(cap, "SyclContexRef"):
+            ctx = <SyclContext>SyclContext(cap)
+            return _queue_ref_copy_from_USMRef_and_SyclContext(ptr, ctx)
+        else:
+            return NULL
+    else:
+        return NULL
+
+
+cdef object _pointers_from_shape_and_stride(
+    int nd, object ary_shape, Py_ssize_t itemsize, Py_ssize_t ary_offset,
+    object ary_strides):
+    """
+    Internal utility: for given array data about shape/layout/element
+    compute left-most displacement when enumerating all elements of the array
+    and the number of bytes of memory between the left-most and right-most
+    displacements.
+
+    Returns: tuple(min_disp, nbytes)
+    """
+    if (nd > 0):
+        if (ary_strides is None):
+            nelems = 1
+            for si in ary_shape:
+                sh_i = int(si)
+                if (sh_i <= 0):
+                    raise ValueError("Array shape elements need to be positive")
+                nelems = nelems * sh_i
+            return (ary_offset, nelems * itemsize)
+        else:
+            min_disp = ary_offset
+            max_disp = ary_offset
+            for i in range(nd):
+                str_i = int(ary_strides[i])
+                sh_i = int(ary_shape[i])
+                if (str_i > 0):
+                    max_disp += str_i * (sh_i - 1)
+                else:
+                    min_disp += str_i * (sh_i - 1);
+            return (min_disp, (max_disp - min_disp + 1) * itemsize)
+    elif (nd == 0):
+        return (ary_offset, itemsize)
+    else:
+        raise ValueError("Array dimensions can not be negative")
+
+
+cdef class _USMBufferData:
+    """
+    Internal data struct populated from parsing
+    `__sycl_usm_array_interface__` dictionary
+    """
+    cdef DPCTLSyclUSMRef p
+    cdef int writeable
+    cdef object dt
+    cdef Py_ssize_t itemsize
+    cdef Py_ssize_t nbytes
+    cdef SyclQueue queue
+
+    @staticmethod
+    cdef _USMBufferData from_sycl_usm_ary_iface(dict ary_iface):
+        cdef object ary_data_tuple = ary_iface.get('data', None)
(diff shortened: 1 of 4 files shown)


oleksandr-pavlyk:
@diptorupd the ci turned up green

dpctl/memory/_sycl_usm_array_interface_utils.pxi
```diff
@@ -0,0 +1,187 @@
...
+cdef DPCTLSyclQueueRef _queue_ref_copy_from_USMRef_and_SyclContext(
+    DPCTLSyclUSMRef ptr, SyclContext ctx):
+    """ Obtain device from pointer and sycl context, use
+        context and device to create a queue from which this memory
+        can be accessible.
+    """
+    cdef SyclDevice dev = _Memory.get_pointer_device(ptr, ctx)
+    cdef DPCTLSyclContextRef CRef = ctx.get_context_ref()
+    cdef DPCTLSyclDeviceRef DRef = dev.get_device_ref()
+    return DPCTLQueue_Create(CRef, DRef, NULL, 0)
```
diptorupd, line 29: The `CRef` and `DRef` attributes to `DPCTLQueue_Create` have `__dpctl_keep` ownership attribute and should be freed in this function.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
```diff
@@ -0,0 +1,187 @@
...
+    elif hasattr(syclobj, '_get_capsule'):
+        cap = syclobj._get_capsule()
+        if pycapsule.PyCapsule_IsValid(cap, "SyclQueueRef"):
+            q = SyclQueue(cap)
+            return _queue_ref_copy_from_SyclQueue(ptr, <SyclQueue> q)
+        elif pycapsule.PyCapsule_IsValid(cap, "SyclContexRef"):
+            ctx = <SyclContext>SyclContext(cap)
+            return _queue_ref_copy_from_USMRef_and_SyclContext(ptr, ctx)
+        else:
+            return QRef
```
diptorupd, line 62: Why no just return `NULL`? You are not using `QRef` any place else.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
```diff
@@ -0,0 +1,187 @@
...
+    """
+    if (nd > 0):
+        if (ary_strides is None):
+            nelems = 1
+            for si in ary_shape:
+                sh_i = int(si)
+                if (sh_i <= 0):
+                    raise ValueError("Array shape elements need to be positive")
+                nelems = nelems * sh_i
+            return (ary_offset, nelems * itemsize)
```
diptorupd, line 86: What is the expected value for `ary_offset` in the default case when strides are based on lexicographic traversal? Is it always going to be 0? If not zero, will be not have a potential out of bound access?

dpctl/memory/_sycl_usm_array_interface_utils.pxi
```diff
@@ -0,0 +1,187 @@
...
+        else:
+            min_disp = ary_offset
+            max_disp = ary_offset
+            for i in range(nd):
+                str_i = int(ary_strides[i])
+                sh_i = int(ary_shape[i])
+                if (str_i > 0):
+                    max_disp += str_i * (sh_i - 1)
```
diptorupd, line 94: Is this correct? should it not be `(str_i*sh_i)-1`?

dpctl/memory/_sycl_usm_array_interface_utils.pxi
```diff
@@ -0,0 +1,187 @@
...
+        if not ary_data_tuple or len(ary_data_tuple) != 2:
+            raise ValueError("__sycl_usm_array_interface__ is malformed:"
+                             " 'data' field is required, and must be a tuple"
+                             " (usm_pointer, is_writeable_boolean).")
+        arr_data_ptr = <Py_ssize_t>ary_data_tuple[0]
+        writeable = 1 if ary_data_tuple[1] else 0
+        # Check that memory and syclobj are consistent:
+        # (USM pointer is bound to this sycl context)
+        memRef = <DPCTLSyclUSMRef>arr_data_ptr
+        QRef = get_queue_ref_from_ptr_and_syclobj(memRef, ary_syclobj)
```
diptorupd, line 149: `QRef` leaks when you raise an exception between here and its use on L 185.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
```diff
@@ -0,0 +1,187 @@
...
+            if (dt.hasobject or not (np.issubdtype(dt.type, np.integer) or
+                                     np.issubdtype(dt.type, np.inexact))):
+                raise TypeError("Only integer types, floating and complex floating "
+                                "types are supported.")
+            itemsize = <Py_ssize_t> (dt.itemsize)
+        except TypeError as e:
+            raise ValueError("__sycl_usm_array_interface__ is malformed:"
+                             " dict('typestr': {}) is unexpected. ".format(ary_typestr)) from e
+
+        if (ary_strides is None or (isinstance(ary_strides, collections.abc.Sized) and
```
diptorupd, line 172: long line. Please reformat to <80 chars.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
oleksandr-pavlyk: But `ctx` and `dev` are python objects, and they will make sure `sycl` objects referenced by `DRef` and `CRef` are disposed of

dpctl/memory/_sycl_usm_array_interface_utils.pxi
oleksandr-pavlyk: No, the displacement is `max(str_i * i for i in range(0, sh_i))` which for positive `str_i` amounts to `str_i*(sh_i-1)`.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
```diff
@@ -0,0 +1,200 @@
...
+    """
+    cdef DPCTLSyclUSMRef p
+    cdef int writeable
+    cdef object dt
+    cdef Py_ssize_t itemsize
+    cdef Py_ssize_t nbytes
+    cdef SyclQueue queue
+
+    @staticmethod
+    cdef _USMBufferData from_sycl_usm_ary_iface(dict ary_iface):
```
diptorupd, line 116: can we expose this to Python iface? I see a reuse in Numba.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
diptorupd: Right. `get_context_ref` and `get_device_ref` will give the raw pointer with lifetime managed by the Python objects.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
oleksandr-pavlyk: Most common case of the offset is 0, most common non-zero offset is when negative strides are involved:

```
In [1]: import numpy as np

In [2]: X = np.empty((4,5))

In [5]: X[::-1, ::-1].ctypes.data
Out[5]: 93992765649928

In [6]: X[::-1, ::-1].ctypes.data - X.ctypes.data
Out[6]: 152
```

This corresponds to (4*5-1)*8, where 8 is the itemsize of the array element.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
oleksandr-pavlyk: Fixed that. Added calls to `DPCTLQueue_Delete` before raising the exception. Thanks for catching it.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
oleksandr-pavlyk: Fix this too, here and in other `dpctl/memory` Cython source files.

dpctl/memory/_sycl_usm_array_interface_utils.pxi
oleksandr-pavlyk: Yes, we can. Accessor functions for native members of the struct will need to be added though.

Should be in a separate PR though

diptorupd (approved):
Please squash and merge.

PR #407, opened 2021-04-27
rename dpctl.dptensor -> dpctl.tensor

oleksandr-pavlyk:
This will create conflict with #394, so it would be good to merge that in master, and I can rebase.


dpctl/tensor/__init__.py
```diff
@@ -17,7 +17,7 @@
 """
     **Data Parallel Tensor Collection**
 
-    `dpctl.dptensor` is an experimental collection of tensor implementations
+    `dpctl.tensor` is an experimental collection of tensor implementations
```
diptorupd, line 20: Do we need the word "experimental"?

diptorupd (dismissed):

dpctl/tensor/__init__.py
oleksandr-pavlyk: We will do it in another PR.

PR #413, opened 2021-04-28
Ignore all Cython generated headers.

diptorupd:
The cython generated header files are added to .gitignore.


diptorupd:
Merging as this is only a `.gitignore` change.

PR #416, opened 2021-04-30
Fix/code style

diptorupd:
@oleksandr-pavlyk I fixed most of the issues in our Cython files found using `flake` and applied `isort` and `black`. One issue I ran into was that the order of imports in `__init__.py` matters. With the sorted imports in the file the tests fail to execute. For this reason, I reverted the changes `isort` made to `__init__.py`.


diptorupd:
@oleksandr-pavlyk I fixed almost all the `flake8` issues. A lot of the issues need to be ignored as these are not applicable to Cython. I have rebased this PR on top of #412, so you can see the last few pending issues in the CI.

PR #386, opened 2021-04-14
Level-zero on Windows (WIP)

1e-to:
Closes #340

PR #426, opened 2021-05-04
Improvements to setup.py

### LINKED ISSUE #361: Optional Level Zero
LevelZero is not available yet in public CI.
Make LevelZero optional by default. Provide flag for forcing LevelZero.

diptorupd:
Closes #361 

- The `build_backend.py` is not invoked using a sub-process. We now use `importlib` to properly import the file as a module.
- `build_backend.py` now provides a function to run the build steps to compile `DPCTLSyclInteface`.
- Initial changes to `build_backend.py` to make it possible to pass in different build options. Presently, the only new option that was added is `l0_support`.
- Minor clean ups and fixes to `setup.py`.

### DIFF
--- a/scripts/build_backend.py
+++ b/scripts/build_backend.py
@@ -18,129 +18,148 @@
 """
 
 
-import glob
-import os
-import shutil
-import subprocess
-import sys
+def build_backend(l0_support=False):
+    import glob
+    import os
+    import shutil
+    import subprocess
+    import sys
 
-IS_WIN = False
-IS_LIN = False
+    IS_WIN = False
+    IS_LIN = False
 
-if "linux" in sys.platform:
-    IS_LIN = True
-elif sys.platform in ["win32", "cygwin"]:
-    IS_WIN = True
-else:
-    assert False, sys.platform + " not supported"
-
-ONEAPI_ROOT = os.environ.get("ONEAPI_ROOT")
-CODE_COVERAGE = os.environ.get("CODE_COVERAGE")
-
-if IS_LIN:
-    DPCPP_ROOT = os.path.join(ONEAPI_ROOT, r"compiler/latest/linux")
-if IS_WIN:
-    DPCPP_ROOT = os.path.join(ONEAPI_ROOT, r"compiler\latest\windows")
-
-dpctl_dir = os.getcwd()
-build_cmake_dir = os.path.join(dpctl_dir, "build_cmake")
-if os.path.exists(build_cmake_dir):
-    for f in os.listdir(build_cmake_dir):
-        f_path = os.path.join(build_cmake_dir, f)
-        if os.path.isdir(f_path):
-            if (f == "level-zero") and os.path.isdir(
-                os.path.join(f_path, ".git")
-            ):
-                # do not delete Git checkout of level zero headers
-                pass
-            else:
-                shutil.rmtree(f_path)
-        else:
-            os.remove(f_path)
-else:
-    os.mkdir(build_cmake_dir)
-os.chdir(build_cmake_dir)
-
-INSTALL_PREFIX = os.path.join(dpctl_dir, "install")
-if os.path.exists(INSTALL_PREFIX):
-    shutil.rmtree(INSTALL_PREFIX)
-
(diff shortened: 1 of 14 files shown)


scripts/build_backend.py
```diff
@@ -18,116 +18,132 @@
...
-            "-DCMAKE_PREFIX_PATH=" + INSTALL_PREFIX,
-            "-DDPCPP_INSTALL_DIR=" + DPCPP_ROOT,
             "-DCMAKE_C_COMPILER:PATH="
-            + os.path.join(DPCPP_ROOT, "bin", "clang"),
+            + os.path.join(DPCPP_ROOT, "bin", "clang-cl.exe"),
             "-DCMAKE_CXX_COMPILER:PATH="
-            + os.path.join(DPCPP_ROOT, "bin", "dpcpp"),
-            "-DDPCTL_ENABLE_LO_PROGRAM_CREATION=ON",
+            + os.path.join(DPCPP_ROOT, "bin", "dpcpp.exe"),
+            "-DCMAKE_PREFIX_PATH=" + INSTALL_PREFIX,
```
oleksandr-pavlyk, line 121: Should L0 be enabled to be enabled on Windows too?

setup.py
```diff
@@ -131,12 +126,12 @@ def extensions():
     elif IS_MAC:
         pass
     elif IS_WIN:
-        libs += ["DPCTLSyclInterface", "sycl"]
+        libs += ["DPCTLSyclInterface"]
 
     if IS_LIN:
         librarys = [dpctl_sycl_interface_lib]
```
oleksandr-pavlyk, line 132: "librarys" -> "libraries"

oleksandr-pavlyk:
The failure on Windows is due to `%ONEAPI_ROOT%\setvars.bat` setting `%DPCPP_ROOT%` environment variable, which does not play well with us using `DPCPP` module, so execution of `FindDPCPP.cmake` is entirely skipped, and variables `DPCPP_OPENCL_LIBRARY` and `DPCPP_SYCL_LIBRARY` end up being not set, causing linkage problems.

The suggested solution is to renamed to CMake module.

scripts/build_backend.py
diptorupd: #386 is specifically attempting to do so.

diptorupd:
> The suggested solution is to renamed to CMake module.

#319 has the change. Let us try once more after #319 and maybe #386 is merged.

oleksandr-pavlyk:
@diptorupd Try using these new keywords with custom installation of compiler if you have the time. I will do that in 1 hour, after a meeting

oleksandr-pavlyk:
The following worked file for me:

```bash
# here $LDH is path to local-disk-home
python setup.py develop --sycl-compiler-prefix=${LDH}/sycl_workspace/llvm/build
```

setup.py
```diff
@@ -266,14 +265,142 @@ def extensions():
...
+            None,
+            "Whether to generate coverage report "
+            "when building the backend library",
+        ),
+    ]
+
+    def initialize_options(self):
+        super().initialize_options()
+        self.level_zero_support = "True"
+        self.coverage = os.getenv("CODE_COVERAGE", "False")
```
diptorupd, line 293: let us just remove the ennvar to control generation of coverage data?

oleksandr-pavlyk: Environment variable  ``CODE_COVERAGE`` is also used to pass additional Cython extensions defines.

I think option ``--coverage`` only makes sense for `develop` command, not `install`.

oleksandr-pavlyk:
README and Quickstart documents were modified.

setup.py
diptorupd: I agree. The `--coverage` option forces a `DEBUG` build for the C API and we should not allow the option with `install`. How and when does `setuptools` call the `extension` function? Is it not possible to check the `--coverage` option that was provided to `setup.py`in `extension`?

oleksandr-pavlyk: The `extensions()` is called during call to `setup`, way before arguments are processed. It is possible to pass the define via `build_ext` command see https://medium.com/@dfdeshom/better-test-coverage-workflow-for-cython-modules-631615eb197a

I have to try if that works.

PR #429, opened 2021-05-05
Added two tests for filter_string property

oleksandr-pavlyk:
The property should be able to reconstruct the same parent device that produced it,
serving as a device identifier.

N.B.: in the setup we have currently, (additional driver in Python's `lib/`, we have two devices with the same filter selector string).

### DIFF
--- a/dpctl-capi/helper/source/dpctl_utils_helper.cpp
+++ b/dpctl-capi/helper/source/dpctl_utils_helper.cpp
@@ -433,13 +433,15 @@ int64_t DPCTL_GetRelativeDeviceId(const device &Device)
 {
     auto relid = -1;
     auto p = Device.get_platform();
+    auto be = p.get_backend();
     auto dt = Device.get_info<sycl::info::device::device_type>();
-    auto dev_vec = p.get_devices(dt);
+    auto dev_vec = device::get_devices(dt);
     int64_t id = 0;
     for (const auto &d_i : dev_vec) {
         if (Device == d_i)
             relid = id;
-        ++id;
+        if (d_i.get_platform().get_backend() == be)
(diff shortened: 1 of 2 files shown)


dpctl-capi/helper/source/dpctl_utils_helper.cpp
```diff
@@ -433,13 +433,15 @@ int64_t DPCTL_GetRelativeDeviceId(const device &Device)
...
+    auto be = p.get_backend();
     auto dt = Device.get_info<sycl::info::device::device_type>();
-    auto dev_vec = p.get_devices(dt);
+    auto dev_vec = device::get_devices(dt);
     int64_t id = 0;
     for (const auto &d_i : dev_vec) {
         if (Device == d_i)
             relid = id;
-        ++id;
+        if (d_i.get_platform().get_backend() == be)
```
oleksandr-pavlyk, line 443: The open source llvm/sycl already implements `device::get_backend()` so the temporary creation of `platform` class could soon be eliminated

diptorupd: This should work. 👍

diptorupd (approved):
LGTM! Thanks for catching the issue.

dpctl-capi/helper/source/dpctl_utils_helper.cpp
diptorupd: @oleksandr-pavlyk We needed  break inside the first `if` after the device equality check.

oleksandr-pavlyk: Right. I will add a separate PR to add it.

PR #430, opened 2021-05-05
Static method of _Memory to create from external allocation

oleksandr-pavlyk:
Method takes USMRef, number of bytes and queue reference.

If USMRef is bound to context in the queue and USM type comes
back not 'unknown', appropriate MemoryUSM* object is created
which assumes ownership of this memory.

Optional `memory_owner` keyword argument may be specified to set
the recorded reference object to something other than None (signaling
that _Memory is responsible for memory deallocation).

This can be used to create MemoryUSM* Python object from natively
allocated memory, populate it, but leave deallocation to the caller.
It could also be set to a PyCapsule object that carries a custom deleter.

### DIFF
--- a/dpctl-capi/source/dpctl_sycl_usm_interface.cpp
+++ b/dpctl-capi/source/dpctl_sycl_usm_interface.cpp
@@ -44,9 +44,18 @@ DEFINE_SIMPLE_CONVERSION_FUNCTIONS(void, DPCTLSyclUSMRef)
 __dpctl_give DPCTLSyclUSMRef
 DPCTLmalloc_shared(size_t size, __dpctl_keep const DPCTLSyclQueueRef QRef)
 {
-    auto Q = unwrap(QRef);
-    auto Ptr = malloc_shared(size, *Q);
-    return wrap(Ptr);
+    if (!QRef) {
+        std::cerr << "Input QRef is nullptr\n";
+        return nullptr;
+    }
+    try {
+        auto Q = unwrap(QRef);
+        auto Ptr = malloc_shared(size, *Q);
+        return wrap(Ptr);
+    } catch (feature_not_supported const &fns) {
+        std::cerr << fns.what() << '\n';
+        return nullptr;
+    }
 }
 
 __dpctl_give DPCTLSyclUSMRef
@@ -54,14 +63,29 @@ DPCTLaligned_alloc_shared(size_t alignment,
                           size_t size,
                           __dpctl_keep const DPCTLSyclQueueRef QRef)
 {
-    auto Q = unwrap(QRef);
-    auto Ptr = aligned_alloc_shared(alignment, size, *Q);
-    return wrap(Ptr);
+    if (!QRef) {
+        std::cerr << "Input QRef is nullptr\n";
+        return nullptr;
+    }
+    try {
+        auto Q = unwrap(QRef);
+        auto Ptr = aligned_alloc_shared(alignment, size, *Q);
+        return wrap(Ptr);
+    } catch (feature_not_supported const &fns) {
+        std::cerr << fns.what() << '\n';
+        return nullptr;
+    }
 }
 
 __dpctl_give DPCTLSyclUSMRef
 DPCTLmalloc_host(size_t size, __dpctl_keep const DPCTLSyclQueueRef QRef)
 {
+    if (!QRef) {
+        std::cerr << "Input QRef is nullptr\n";
+        return nullptr;
+    }
+    // SYCL 2020 spec: for devices without aspect::usm_host_allocations:
+    // undefined behavior
     auto Q = unwrap(QRef);
     auto Ptr = malloc_host(size, *Q);
     return wrap(Ptr);
@@ -72,6 +96,12 @@ DPCTLaligned_alloc_host(size_t alignment,
                         size_t size,
                         __dpctl_keep const DPCTLSyclQueueRef QRef)
 {
+    if (!QRef) {
+        std::cerr << "Input QRef is nullptr\n";
+        return nullptr;
+    }
+    // SYCL 2020 spec: for devices without aspect::usm_host_allocations:
+    // undefined behavior
     auto Q = unwrap(QRef);
     auto Ptr = aligned_alloc_host(alignment, size, *Q);
     return wrap(Ptr);
@@ -80,9 +110,18 @@ DPCTLaligned_alloc_host(size_t alignment,
 __dpctl_give DPCTLSyclUSMRef
 DPCTLmalloc_device(size_t size, __dpctl_keep const DPCTLSyclQueueRef QRef)
 {
-    auto Q = unwrap(QRef);
-    auto Ptr = malloc_device(size, *Q);
-    return wrap(Ptr);
+    if (!QRef) {
+        std::cerr << "Input QRef is nullptr\n";
+        return nullptr;
+    }
+    try {
+        auto Q = unwrap(QRef);
+        auto Ptr = malloc_device(size, *Q);
+        return wrap(Ptr);
+    } catch (feature_not_supported const &fns) {
+        std::cerr << fns.what() << '\n';
+        return nullptr;
+    }
 }
 
 __dpctl_give DPCTLSyclUSMRef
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -20,7 +20,7 @@ per-file-ignores =
     dpctl/_sycl_platform.pyx: E999, E225
     dpctl/_sycl_queue.pyx: E999, E225, E226, E227
     dpctl/_sycl_queue_manager.pyx: E999, E225
-    dpctl/memory/_memory.pyx: E999, E225, E226, E227
+    dpctl/memory/_memory.pyx: E999, E225, E226, E227, E125
```
diptorupd, line 23: Where is it complaining about E125?

dpctl/memory/_memory.pxd
```diff
@@ -55,6 +55,12 @@ cdef public class _Memory [object Py_MemoryObject, type Py_MemoryType]:
         DPCTLSyclUSMRef p, SyclContext ctx)
     @staticmethod
     cdef public bytes get_pointer_type(DPCTLSyclUSMRef p, SyclContext ctx)
+    @staticmethod
+    cdef public object create_from_usm_pointer_size_qref(
+        DPCTLSyclUSMRef USMRef,
+        Py_ssize_t nbytes,
+        DPCTLSyclQueueRef QRef,
+        object memory_owner=*)
```
diptorupd, line 63: Style - close parent should be on next line aligned with the `cdef`.

dpctl/memory/_memory.pyx
```diff
@@ -431,6 +433,53 @@ cdef class _Memory:
...
+            raise ValueError("Could not retrieve context from QRef")
+        usm_type = DPCTLUSM_GetPointerType(USMRef, CRef)
+        if usm_type == b"shared":
+            mem_ty = MemoryUSMShared
+        elif usm_type == b"device":
+            mem_ty = MemoryUSMDevice
+        elif usm_type == b"host":
+            mem_ty = MemoryUSMHost
+        else:
+            raise ValueError("Argument pointer is not bound to "
```
diptorupd, line 471: Style - The string should start with an indent on next line and the closing parenthesis should be de-indented on the next line, *i.e.* aligned with `ValueError`.

dpctl/memory/_memory.pyx
```diff
@@ -431,6 +433,53 @@ cdef class _Memory:
 
         return <bytes>usm_type
 
+    @staticmethod
+    cdef object create_from_usm_pointer_size_qref(
+        DPCTLSyclUSMRef USMRef, Py_ssize_t nbytes,
+        DPCTLSyclQueueRef QRef, object memory_owner=None):
+        r"""
+        Create appropriate MemoryUSM* object from pre-allocated
```
diptorupd, line 441: Style - double backtick arounf `MemoryUSM*`

.flake8
oleksandr-pavlyk: Alignment

dpctl/memory/_memory.pyx
```diff
@@ -431,6 +433,53 @@ cdef class _Memory:
...
+
+        Memory will be freed by MemoryUSM* object for default
+        value of memory_owner keyword. The non-default value should
+        be an object whose dealloc slot frees the memory.
+
+        The object may not be a no-op dummy Python object to
+        delay freeing the memory until later times.
+        """
+        cdef const char *usm_type
+        cdef DPCTLSyclContextRef CRef = NULL
```
diptorupd, line 452: leaked?

dpctl/memory/_memory.pyx
diptorupd, line 471: `CRef` needs to be freed here.

dpctl/memory/_memory.pyx
```diff
@@ -431,6 +433,53 @@ cdef class _Memory:
...
+            mem_ty = MemoryUSMHost
+        else:
+            raise ValueError("Argument pointer is not bound to "
+                             "context in the given queue")
+        res = _Memory.__new__(_Memory)
+        _mem = <_Memory> res
+        _mem._cinit_empty()
+        _mem.memory_ptr = USMRef
+        _mem.nbytes = nbytes
+        QRef_copy = DPCTLQueue_Copy(QRef)
```
diptorupd, line 478: Check if copy worked.

dpctl/memory/_memory.pyx
```diff
@@ -431,6 +433,53 @@ cdef class _Memory:
...
+        _mem.queue = SyclQueue._create(QRef_copy)
```
diptorupd, line 479: `SyclQueueCreationError` check?

dpctl/memory/_memory.pyx
diptorupd: or maybe not.

dpctl/memory/_memory.pyx
```diff
@@ -431,6 +434,55 @@ cdef class _Memory:
...
+        cdef const char *usm_type
+        cdef DPCTLSyclContextRef CRef = NULL
+        cdef DPCTLSyclQueueRef QRef_copy = NULL
+        cdef _Memory _mem
+        cdef object mem_ty
+        if nbytes <= 0:
+            raise ValueError("Number of bytes must must be positive")
+        if (QRef is NULL):
+            raise TypeError("Argument DPCTLSyclQueueRef is NULL")
+        CRef = DPCTLQueue_GetContext(QRef)
```
diptorupd, line 461: `DPCTLQueue_GetContext` has `__dpctl_give` so `CRef` needs to be freed.

dpctl/memory/_memory.pyx
```diff
@@ -431,6 +434,55 @@ cdef class _Memory:
...
+        if (CRef is NULL):
+            raise ValueError("Could not retrieve context from QRef")
+        usm_type = DPCTLUSM_GetPointerType(USMRef, CRef)
```
diptorupd, line 464: I think we should move to `isl` style where all functions args are by default `__dpctl_take` instead of `__dpctl_keep`. That makesit easy to keep track of pointer releases and if some are missed it will cause a crash immediately.

.flake8
diptorupd: Yes, but where. I am asking since perhaps we can fix the issue instead of suppressing ti.

oleksandr-pavlyk: in `_memory.pyx`, lines 438-440

diptorupd: Try moving the closing parenthesis on L439 to the next line aligned with the `cdef`.

oleksandr-pavlyk: That worked! All concerns have been addressed.

dpctl/memory/_memory.pyx
oleksandr-pavlyk: I just added `DPCTLContext_Delete(CRef)` right after line 464

dpctl/memory/_memory.pyx
oleksandr-pavlyk: Added delete. Thanks for catching this leak.

dpctl/memory/_memory.pyx
oleksandr-pavlyk: We could refactor in some not very distant future.

dpctl/memory/_memory.pyx
oleksandr-pavlyk: Added the check.

dpctl/memory/_memory.pyx
```diff
@@ -431,6 +435,65 @@ cdef class _Memory:
...
```
diptorupd, line 466: The DPCTLUSM_GetPointerType function actually does not check if the context and pointer are valid. How will the SYCL get_pointer_type behave if I pass in a nullptr or a null CRef? Can you fix that here?

diptorupd: We need to tackle the error handlinG ASAP. So much to do gulp! :(

oleksandr-pavlyk: Done.

dpctl-capi/source/dpctl_sycl_usm_interface.cpp
```diff
@@ -44,24 +44,46 @@ DEFINE_SIMPLE_CONVERSION_FUNCTIONS(void, DPCTLSyclUSMRef)
...
 
 __dpctl_give DPCTLSyclUSMRef
 DPCTLmalloc_host(size_t size, __dpctl_keep const DPCTLSyclQueueRef QRef)
 {
+    if (!QRef) {
+        std::cerr << "Input QRef is nullptr\n";
+        return nullptr;
+    }
     auto Q = unwrap(QRef);
     auto Ptr = malloc_host(size, *Q);
```
diptorupd, line 88: Should we not check for an exception here? At least a `bad_alloc` exception?

dpctl-capi/source/dpctl_sycl_usm_interface.cpp
```diff
@@ -44,24 +44,46 @@ DEFINE_SIMPLE_CONVERSION_FUNCTIONS(void, DPCTLSyclUSMRef)
...
-    return wrap(Ptr);
+    if (!QRef) {
+        std::cerr << "Input QRef is nullptr\n";
+        return nullptr;
+    }
+    try {
+        auto Q = unwrap(QRef);
+        auto Ptr = malloc_shared(size, *Q);
+        return wrap(Ptr);
+    } catch (feature_not_supported const &fns) {
```
diptorupd, line 55: Are you sure a `feature_not_supported` error is thrown?

oleksandr-pavlyk:
CI failure is network related, timing out when trying to retrieve NumPy for pip installation

```
pip._vendor.urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='binstar-cio-packages-prod.s3.amazonaws.com', port=443): Read timed out.
```

diptorupd (approved):

PR #440, opened 2021-05-11
Added Device class representing Data-API notion of device

oleksandr-pavlyk:
Data-API notion of device is higher level abstraction. The class
does allow for public constructor, has class method to create
instance instead.

It typesets as Device(filter_string) for parent devices and
Device(queue) for sub-devices. It can constructed from

   - filter selector string
   - SyclDevice corresponding to a root device
     (attempt at passing sub-device raises and error)
   - SyclQueue
   - Another instance of Device class

It implements sycl_queue, sycl_device, sycl_context attributes.

usm_ndarray adds .device property, and .to_device(device) method.

```
In [5]: X = dpt.usm_ndarray((4, 5), dtype='d')

In [6]: X.device
Out[6]: Device(level_zero:gpu:0)

In [7]: Y = X.to_device('cpu')

In [8]: Y.device
Out[8]: Device(opencl:cpu:0)
```

### DIFF
--- a/dpctl/tensor/_device.py
+++ b/dpctl/tensor/_device.py
@@ -18,15 +18,15 @@ import dpctl
 
 class Device:
     """
-    Class representing Data-API concept of Device
+    Class representing Data-API concept of device.
 
-    This is a wrapper around `dpctl.SyclQueue` with custom
+    This is a wrapper around :class:`dpctl.SyclQueue` with custom
     formatting. The class does not have public constructor,
     but a class method to construct it from device= keyword
     in Array-API functions.
 
-    Instance can be queries for `sycl_queue`, `sycl_context`,
-    or `sycl_device`
+    Instance can be queried for ``sycl_queue``, ``sycl_context``,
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_device.py
```diff
@@ -0,0 +1,53 @@
+import dpctl
```
diptorupd, line 1: license header is needed.

dpctl/tensor/_device.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class Device:
+    def __new__(clf, *args, **kwargs):
```
diptorupd, line 5: Please document the class and the functions.

diptorupd, line 5: `s/clf/cls/`

dpctl/tensor/_device.py
oleksandr-pavlyk: Added.

diptorupd (approved):

PR #446, opened 2021-05-18
Add action to build dpctl using oneAPI and generate docs

### LINKED ISSUE #205: Auto-generate API documentation with each commit to master using Github Actions
Currently, I have been updating the API docs for `dpctl` by manually generating and pushing the documents to the gh-pages branch. The steps that I use are:

**Step 1**
```
git checkout master
cd docs
mkdir build
cd build 
cmake -DDPCTL_DOCGEN_PREFIX=<Some location out of tree>
make Sphinx
```
**Step 2**
```
git clean -dfx
git checkout gh-pages
git rm -rf *
mv <Location where docs were generated in first step> .
git ci -m "Updated docs."
git push
```
We should be able to do these using Azure as part of CI.

oleksandr-pavlyk:
Closes #205

Attempt to add GH action to install oneAPI compiler, compiler dpctl with it locally, generate HTML version of docs.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -1,0 +1,94 @@
+name: Generate documentation
+on:
+  push:
+    branches:
+      - master
+      - doc-deployment-action
+
+jobs:
+  build-and-deploy:
+    name: Build and Deploy Documentation
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.6.0
+        with:
+          access_token: ${{ github.token }}
+      - name: Add Intel repository
+        run: |
+          wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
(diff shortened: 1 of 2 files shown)


diptorupd:
@oleksandr-pavlyk Thanks for looking into this :+1:. Your PR will actually also address or start to address the issue with a public CI. I fixed one missing step that was causing your jig to fail. We need to checkout the sources before calling `python setup.py develop`. However, we need to finish the changes in #319 and #426 before we can get any further.

oleksandr-pavlyk:
I managed to delete all content of gh-pages by running the action in this branch.

This was enough:

```
67     - name: Deploy docs
68        uses: peaceiris/actions-gh-pages@v3
69        with:
70          github_token: ${{ secrets.GITHUB_TOKEN }}
71          # publish_dir: doc/_build/html/
```

Sorry about that.  Since the damage has been done, I have uncommented the `publish_dir` line. Once it works for master, I can try to recover 0.6 and 0.7 versions too.

oleksandr-pavlyk:
The action republished some pages, but there is no information about Python submodules. Sigh.

oleksandr-pavlyk:
Rebased on top of master to pick up changes to Doxygen cmake. Merging.

PR #449, opened 2021-05-18
Added test for as_usm_memory

oleksandr-pavlyk:
This test checks for conformance to __sycl_usm_array_interface__ spect from Wiki


oleksandr-pavlyk:
CI is green, merging.

PR #450, opened 2021-05-18
Add a flag to make doxygen HTML generation optional.

diptorupd:
We use Doxyrest to generate the API docs for the dpctl C API sources. The changes in the PR make it possible to turn off Doxygen HTML generation, and by default the HTML generation is turned off.


diptorupd:
Merging as this is unrelated to any code changes.

PR #453, opened 2021-05-20
Feature/producing filter selector

oleksandr-pavlyk:
dpctl-capi/:
   * Added `DPCTLDeviceMgr_GetPositionInDevices(DRef, device_mask)` that returns ordinal of the given DRef in the filtered vector returned by ``sycl::device::get_devices()``, or -1 if not found (e.g. `DRef` points to a sub-device).
   * Added tests, and docstring

dpctl/:
   * Added `DPCTLDeviceMgr_GetPositionInDevices` to `_backend.pxd`
   * Implemented 4 cdef methods of `dpctl.SyclDevice`:
        * `get_overall_ordinal(self)` - index in unfiltered vector returned by ``get_devices()``.
        * `get_backend_ordinal(self)` - index in vector filtered to only contain devices with the backend of this device
        * `get_device_type_ordinal(self)` - index in vector filtered to only contain devices with device_type as in this device
        * `get_backend_and_device_type_ordinal(self)` - indexed in vector filtered to only contain devices with device_type and backend of this device
    * Added Python visible method `get_filter_string(self, include_backend=True, include_device_type=True)` which producer filter selector string selecting this device that would include or omit backend/device_type as requested.
    * Test added

Example:

```python
In [1]: import dpctl

In [2]: default_dev = dpctl.SyclDevice()

In [3]: default_dev.get_filter_string() # same as default_dev.filter_string
Out[3]: 'level_zero:gpu:0'

In [4]: default_dev.get_filter_string(include_backend=False)
Out[4]: 'gpu:1'

In [5]: default_dev == dpctl.SyclDevice(_)
Out[5]: True

In [6]: default_dev.get_filter_string(include_device_type=False)
Out[6]: 'level_zero:0'

In [7]: default_dev == dpctl.SyclDevice(_)
Out[7]: True

In [8]: default_dev.get_filter_string(include_device_type=False, include_backend=False)
Out[8]: '4'

In [9]: default_dev == dpctl.SyclDevice(_)
Out[9]: True
```

### DIFF
--- a/dpctl-capi/source/dpctl_sycl_device_manager.cpp
+++ b/dpctl-capi/source/dpctl_sycl_device_manager.cpp
@@ -164,6 +164,35 @@ DPCTLDeviceMgr_GetDevices(int device_identifier)
     return wrap(Devices);
 }
 
+int DPCTLDeviceMgr_GetPositionInDevices(__dpctl_keep DPCTLSyclDeviceRef DRef,
+                                        int device_identifier)
+{
+    constexpr int not_found = -1;
+    if (!DRef) {
+        return not_found;
+    }
+
+    const auto &root_devices = device::get_devices();
+    default_selector mRanker;
+    int index = not_found;
+    auto reference_device = *(unwrap(DRef));
+
+    for (const auto &root_device : root_devices) {
+        if (mRanker(root_device) < 0)
+            continue;
+        auto Bty(DPCTL_SyclBackendToDPCTLBackendType(
+            root_device.get_platform().get_backend()));
+        auto Dty(DPCTL_SyclDeviceTypeToDPCTLDeviceType(
+            root_device.get_info<info::device::device_type>()));
+        if ((device_identifier & Bty) && (device_identifier & Dty)) {
+            ++index;
+            if (root_device == reference_device)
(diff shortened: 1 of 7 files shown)


dpctl-capi/source/dpctl_sycl_device_manager.cpp
```diff
@@ -164,6 +164,35 @@ DPCTLDeviceMgr_GetDevices(int device_identifier)
...
+    if (!DRef) {
+        return not_found;
+    }
+
+    const auto &root_devices = device::get_devices();
+    default_selector mRanker;
+    int index = not_found;
+    auto reference_Device = *(unwrap(DRef));
+
+    for (const auto &root_Device : root_devices) {
```
diptorupd, line 180: Minor nitpick. Please change to proper snake case naming `s/root_Device/root_device`

diptorupd:
LGTM bar the minor naming nitpick.

dpctl-capi/source/dpctl_sycl_device_manager.cpp
oleksandr-pavlyk: Done :)

PR #455, opened 2021-05-21
Correctly copy symlinks to the dpctl dir.

diptorupd:
After introducing so-versioning for `libDPCTLSyclInterface` the symlinks are not being correctly copied from the build directory (`build_cmake`) to the install directory (`dpctl`). The PR fixes this by using the `follow_symlinks` argument to `shutils.copy`.

### DIFF
--- a/scripts/build_backend.py
+++ b/scripts/build_backend.py
@@ -143,7 +143,11 @@ def build_backend(
         for file in glob.glob(
             os.path.join(dpctl_dir, "install", "lib", "*.so*")
         ):
-            shutil.copy(file, os.path.join(dpctl_dir, "dpctl"))
+            shutil.copy(
+                src=file,
+                dst=os.path.join(dpctl_dir, "dpctl"),
+                follow_symlinks=False,
+            )
     elif IS_WIN:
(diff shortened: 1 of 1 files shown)


scripts/build_backend.py
```diff
@@ -143,7 +143,13 @@ def build_backend(
         for file in glob.glob(
             os.path.join(dpctl_dir, "install", "lib", "*.so*")
         ):
-            shutil.copy(file, os.path.join(dpctl_dir, "dpctl"))
+            print(file)
```
oleksandr-pavlyk, line 146:
```suggestion
            print(file)
            
```

scripts/build_backend.py
```diff
@@ -143,7 +143,13 @@ def build_backend(
...
+            print(dpctl_dir)
```
oleksandr-pavlyk, line 147:
```suggestion
```

scripts/build_backend.py
diptorupd: Sorry these are leftover changes. I am removing the prints.

oleksandr-pavlyk (approved):

PR #470, opened 2021-05-25
Turn off SO versioning for release 0.8

oleksandr-pavlyk:
The issue is that conda-build would turn symbolic links of version
library shared objects into hard copies.

Downstream project (e.g. numba-dppy) may end up loading different
hard copies of the library and thus maintain different global states
and encounter segfaults from discrepancies in these states


tomashek:
@oleksandr-pavlyk , conda-build definitely supports symlinks for Linux, and we are supposed to have so versioning. Where did you see this problem?

oleksandr-pavlyk:
@tomashek Building dpctl using conda-build does that (look at `lib/libDPCTLSyclInterface.so*` files in the conda tar ball)

PR #479, opened 2021-05-27
Fix sub devices

### LINKED ISSUE #475: Segfault in creating sub-devices
Passing an incorrect argument for `count` to `create_sub_devices` can lead to a segfault. The issue is that SYCL expects an `unsigned long` as the count and passing a zero as done in the below reproducer. We are missing an exception check in C API.

```
In [1]: import dpctl

In [2]:                 cpu_d = dpctl.SyclDevice("cpu")
   ...:                 cpu_count = cpu_d.max_compute_units
   ...:                 sub_devs = cpu_d.create_sub_devices(partition=cpu_count // 100)
   ...:                 for d in sub_devs:
   ...:                     d.print_device_info()
```

oleksandr-pavlyk:
Closes #475 

Issue alike crash of ``dpctl.SyclDevice("cpu").create_sub_devices(partition=0)`` should be fixed.

### DIFF
--- a/dpctl-capi/source/dpctl_sycl_device_interface.cpp
+++ b/dpctl-capi/source/dpctl_sycl_device_interface.cpp
@@ -29,6 +29,7 @@
 #include "Support/CBindingWrapping.h"
 #include "dpctl_sycl_device_manager.h"
 #include <CL/sycl.hpp> /* SYCL headers   */
+#include <algorithm>
 #include <cstring>
 
 using namespace cl::sycl;
@@ -577,8 +578,8 @@ DPCTLDevice_CreateSubDevicesEqually(__dpctl_keep const DPCTLSyclDeviceRef DRef,
                                     size_t count)
 {
     vector_class<DPCTLSyclDeviceRef> *Devices = nullptr;
-    auto D = unwrap(DRef);
-    if (D) {
+    if (DRef && count) {
+        auto D = unwrap(DRef);
         try {
             auto subDevices = D->create_sub_devices<
                 info::partition_property::partition_equally>(count);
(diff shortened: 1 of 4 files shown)


dpctl-capi/source/dpctl_sycl_device_interface.cpp
```diff
@@ -577,8 +578,8 @@ DPCTLDevice_CreateSubDevicesEqually(__dpctl_keep const DPCTLSyclDeviceRef DRef,
                                     size_t count)
 {
     vector_class<DPCTLSyclDeviceRef> *Devices = nullptr;
-    auto D = unwrap(DRef);
-    if (D) {
+    if (DRef && count) {
```
diptorupd, line 581: Please print a warning if count is 0.

dpctl-capi/source/dpctl_sycl_device_interface.cpp
```diff
@@ -612,11 +613,22 @@ DPCTLDevice_CreateSubDevicesByCounts(__dpctl_keep const DPCTLSyclDeviceRef DRef,
     vector_class<DPCTLSyclDeviceRef> *Devices = nullptr;
     std::vector<size_t> vcounts;
     vcounts.assign(counts, counts + ncounts);
```
diptorupd, line 615: Since you know the size, might as well reserve the memory saves a realloc.

dpctl-capi/source/dpctl_sycl_device_interface.cpp
```diff
@@ -612,11 +613,22 @@ DPCTLDevice_CreateSubDevicesByCounts(__dpctl_keep const DPCTLSyclDeviceRef DRef,
...
-    auto D = unwrap(DRef);
-    if (D) {
+    size_t min_elem = *std::min_element(vcounts.begin(), vcounts.end());
+    if (DRef && min_elem) {
```
diptorupd, line 617: Again some explanation to a user why they are getting back a nullptr will help.

dpctl/_sycl_device.pyx
```diff
@@ -706,7 +706,12 @@ cdef class SyclDevice(_SyclDevice):
             the sub-devices.
         """
         cdef DPCTLDeviceVectorRef DVRef = NULL
-        DVRef = DPCTLDevice_CreateSubDevicesEqually(self._device_ref, count)
+        if count > 0:
+            DVRef = DPCTLDevice_CreateSubDevicesEqually(self._device_ref, count)
+        else:
+            raise ValueError(
+                "Creating sub-devices with zero compute units is requested"
```
diptorupd, line 713: What will Cython do here? Will it raise the exception of silently cause havoc?

dpctl/_sycl_device.pyx
```diff
@@ -734,10 +740,17 @@ cdef class SyclDevice(_SyclDevice):
             )
         for i in range(ncounts):
             counts_buff[i] = counts[i]
-        DVRef = DPCTLDevice_CreateSubDevicesByCounts(
-            self._device_ref, counts_buff, ncounts
-        )
+            if counts_buff[i] == 0:
```
diptorupd, line 743: Why not do this before you `malloc` counts_buff to not even allocate the array if input is not valid.

dpctl/_sycl_device.pyx
oleksandr-pavlyk: the `cdef list func_name(...)` will correctly propagate the Python error raised.

dpctl-capi/source/dpctl_sycl_device_interface.cpp
oleksandr-pavlyk: done

dpctl-capi/source/dpctl_sycl_device_interface.cpp
oleksandr-pavlyk: done.

dpctl-capi/source/dpctl_sycl_device_interface.cpp
oleksandr-pavlyk: done.

diptorupd (approved):

PR #481, opened 2021-05-27
Implemented support for non-standard installation of oneAPI compiler by dpctl

### LINKED ISSUE #480: Building with oneAPI for non-standard compiler installation
For the default value of `--sycl-compiler-prefix` option to `develop`/`install` setup commands we require `ONEAPI_ROOT` and do not set `DPCTL_CUSTOM_DPCPP_INSTALL_DIR`.

The logic is all built assuming that setting ``DPCTL_CUSTOM_DPCPP_INSTALL_DIR`` signals use of custom-built DPC++ toolchain. 

When oneAPI compiler bits are installed into Python environment, we can not use `ONEAPI_ROOT`, since compiler's location is no longer `${ONEAPI_ROOT}/compiler/latest/linux` but `${CONDA_PREFIX}` instead, and must rely on `--sycl-compiler-prefix` which sets  `DPCTL_CUSTOM_DPCPP_INSTALL_DIR` and CMake script starts assuming that we build with OS DPC++ toolchain.

@AndresGuzman-Ballen

oleksandr-pavlyk:
Closes #480 

@AndresGuzman-Ballen Please check to see if this changes unblocks the use of DPC++ installation in Python environment.

### DIFF
--- a/dpctl-capi/CMakeLists.txt
+++ b/dpctl-capi/CMakeLists.txt
@@ -9,9 +9,13 @@ project(
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
 find_package(Git REQUIRED)
 
-option(DPCTL_CUSTOM_DPCPP_INSTALL_DIR
-    "Use a custom version of DPCPP installed at the provided location."
+option(DPCTL_DPCPP_HOME_DIR
+    "The installation home for the DPC++ toolchain compiler."
     OFF
+)
+option(DPCTL_DPCPP_FROM_ONEAPI
+    "Indicates whether DPCTL_DPCPP_HOME_DIR points to a oneAPI installation."
+    ON
 )
 # Option to turn on support for creating Level Zero interoperability programs
 # from a SPIR-V binary file.
@@ -36,10 +40,10 @@ option(DPCTL_BUILD_CAPI_TESTS
 )
 
 # Minimum version requirement only when oneAPI dpcpp is used.
-if(DPCTL_CUSTOM_DPCPP_INSTALL_DIR)
-    find_package(IntelSycl REQUIRED)
-else()
+if(DPCTL_DPCPP_FROM_ONEAPI)
     find_package(IntelSycl 2021.2.0 REQUIRED)
+else()
+    find_package(IntelSycl REQUIRED)
 endif()
 
 if(DPCTL_ENABLE_LO_PROGRAM_CREATION)
--- a/dpctl-capi/cmake/modules/FindIntelSycl.cmake
+++ b/dpctl-capi/cmake/modules/FindIntelSycl.cmake
@@ -36,10 +36,9 @@ include(FindPackageHandleStandardArgs)
 
 # Check if a specific DPC++ installation directory was provided then set
 # IntelSycl_ROOT to that path.
-if(DPCTL_CUSTOM_DPCPP_INSTALL_DIR)
-    set(IntelSycl_ROOT ${DPCTL_CUSTOM_DPCPP_INSTALL_DIR})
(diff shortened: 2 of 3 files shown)


diptorupd:
@oleksandr-pavlyk I do not like the approach of checking the compiler string. Just checking `dpcpp` is not a good practice. One can use oneAPI and use `-DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXXFLAGS=-fsycl` and not specify `dpcpp`. Please see my proposed solution in #480.

oleksandr-pavlyk:
> @oleksandr-pavlyk I do not like the approach of checking the compiler string. Just checking `dpcpp` is not a good practice. One can use oneAPI and use `-DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXXFLAGS=-fsycl` and not specify `dpcpp`. Please see my proposed solution in #480.

Ok, I will implement that, but the only difference would be that we would not insist on 2021.2 minimum version requirement.

oleksandr-pavlyk:
Confirmed that `python setup.py develop` works with default one API, and with custom-toolchain, using ``python setup.py develop --sycl-compiler-prefix=${DPCPP_HOME}/llvm/build``

dpctl-capi/CMakeLists.txt
```diff
@@ -9,10 +9,14 @@ project(
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
 find_package(Git REQUIRED)
 
-option(DPCTL_CUSTOM_DPCPP_INSTALL_DIR
+option(DPCTL_DPCPP_HOME_DIR
     "Use a custom version of DPCPP installed at the provided location."
```
diptorupd, line 13: The message should say "The installation home for the DPCPP compiler."

oleksandr-pavlyk: Good catch. Fixed.

dpctl-capi/cmake/modules/FindIntelSycl.cmake
```diff
@@ -49,38 +48,32 @@ elseif(DEFINED ENV{ONEAPI_ROOT})
...
 
 # We will extract the version information from the compiler
-if(USING_ONEAPI_DPCPP)
-    set(dpcpp_cmd "${IntelSycl_ROOT}/bin/dpcpp")
-    set(dpcpp_arg "--version")
-else()
-    set(dpcpp_cmd "${IntelSycl_ROOT}/bin/clang++")
-    set(dpcpp_arg "--version")
-endif()
+set(clangpp_cmd "${IntelSycl_ROOT}/bin/clang++")
```
diptorupd, line 61: `s\clangpp\clangcpp`

oleksandr-pavlyk: clangpp stood for clang++, clangcpp does not :)

diptorupd: Aha! then how about `clangxx`

oleksandr-pavlyk: Renamed to `clangxx`.

diptorupd:
LGTM. In @AndresGuzman-Ballen's case all he needs to do is pass in `--sycl_compiler_path=${BUILD_DIR}`. Right?

oleksandr-pavlyk:
Yes, and he already got this working on Linux, and Windows (a test fails here, but building process works)

diptorupd (approved):
Please squash and merge

diptorupd:
@oleksandr-pavlyk please update the docs in a follow up or as a separate commit in this PR.

diptorupd (approved):

PR #485, opened 2021-06-02
Remove imports from __future__. The future is here now.

diptorupd:
dpctl supports only Python 3, we do not need to import from `__future__`.


oleksandr-pavlyk (approved):
> The future is here now.

Hear, hear!

PR #487, opened 2021-06-02
Fix SDL check

oleksandr-pavlyk:
Follow DPCTLDevice_CreateFromSelector which passes the check.

This should fix an issue from SAT-4170: 

>  Possible memory leak. Dynamic memory stored in 'P' allocated through function 'new' at line 84 can be lost at line 88


diptorupd (approved):
Thanks!

PR #489, opened 2021-06-02
Defines __hash__ for all classes that define __eq__ method.

### LINKED ISSUE #483: All the Sycl* extension types should define a __hash__ function.
Currently, the `Sycl*` Cython extension types can not be hashed as we do not define the `__hash__` function. Due to the lack of the `__hash__` function dpctl types cannot be stored as dictionary keys.

diptorupd:
- SyclContext, SyclDevice, SyclQueue classes previously where
    not hashable. The PR defines __hash__ function based on the
    C API native pointer value stored in the classes.

Closes #483

### DIFF
--- a/dpctl/_sycl_context.pyx
+++ b/dpctl/_sycl_context.pyx
@@ -337,6 +337,14 @@ cdef class SyclContext(_SyclContext):
         else:
             return False
 
+    def __hash__(self):
+        """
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk:
I think simply hashing the address is not enough. We also need to use ``std::hash`` of the inderlying object.

diptorupd:
superseded by #491

PR #496, opened 2021-06-23
Added test class to test DRef=nullptr case

oleksandr-pavlyk:
In local build the coverage went from red to yellow.


diptorupd (approved):
👍

PR #497, opened 2021-06-23
Extend parameterized test in test_sycl_queue_interface

oleksandr-pavlyk:
Test with queues that use different properties, default/non-default error handler to improve coverage.

### DIFF
--- a/dpctl-capi/tests/test_sycl_queue_interface.cpp
+++ b/dpctl-capi/tests/test_sycl_queue_interface.cpp
@@ -41,20 +41,30 @@ namespace
 {
 DEFINE_SIMPLE_CONVERSION_FUNCTIONS(queue, DPCTLSyclQueueRef);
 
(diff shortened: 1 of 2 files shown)


dpctl-capi/tests/test_sycl_queue_interface.cpp
```diff
@@ -41,20 +41,31 @@ namespace
 {
 DEFINE_SIMPLE_CONVERSION_FUNCTIONS(queue, DPCTLSyclQueueRef);
 
+void error_handler_fn(int err)
```
diptorupd, line 44: Do not suppress the warning in this way. Instead use,
```
void error_handler_fn(int /**/)
```

oleksandr-pavlyk: Done.

diptorupd (approved):

PR #499, opened 2021-06-25
Added a test that retrieved DPCPP compiled kernel and submits them via DPCTLQueue_SubmitRange

oleksandr-pavlyk:
This does not presently work with coverage turned on due to LLVM version discrepancy between IGC and DPC++.

So the test is excluded for the build with coverage.

### DIFF
--- a/dpctl-capi/CMakeLists.txt
+++ b/dpctl-capi/CMakeLists.txt
@@ -86,6 +86,7 @@ elseif(UNIX)
         "-Wuninitialized "
         "-Wmissing-declarations "
         "-fdiagnostics-color=auto "
+        "-Wno-deprecated-declarations "
     )
     string(CONCAT SDL_FLAGS
         "-fstack-protector "
--- a/dpctl-capi/dbg_build.sh
+++ b/dpctl-capi/dbg_build.sh
@@ -13,7 +13,7 @@ cmake                                                       \
     -DCMAKE_CXX_COMPILER=dpcpp                              \
     -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX}                \
     -DCMAKE_PREFIX_PATH=${INSTALL_PREFIX}                   \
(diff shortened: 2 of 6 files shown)


oleksandr-pavlyk:
Even though, logically, the ``sycl::kernel`` produced by ``dpcpp_kernel::get_fill_kernel`` has been given all the arguments that command group function object lambda would capture by reference, removing those from `DPCTLQueue_SubmitRange` results in an incorrect result (test fails).

Thus, duplicating these captured values via `set_args` appears needed. Perhaps this is to be expected, as the captured values are captured on the host side and must still be sent over to the device.

oleksandr-pavlyk:
@diptorupd I have added tests for `DCPTLQueue_SubmitNDRange`. Expanded tests for `int`, `unsigned int`, `float` and `double`.

oleksandr-pavlyk:
To fix the "generate-coverage" workflow it sufficies to remove linking to `dpcpp_kernels` static library.

The error is

```
[100%] Linking CXX executable dpctl_c_api_tests
error: Linking COMDATs named '__profd_dpcpp_kernels.cpp__ZN7__spirvL16initGlobalOffsetILi1EN2cl4sycl2idILi1EEEEET0_v': noduplicates has been violated!
dpcpp: error: sycl-link command failed with exit code 1 (use -v to see invocation)
```
 
@diptorupd Is there another way to fix it?

oleksandr-pavlyk:
Failures on Linux are due to OpenCL runtime packages being unavailable on PyPI due to limitation on the size of published packages.

So this can be merged.

diptorupd:
Merging to master as TeamCity CI fails due to `dpcpp_cpp_rt` package not being there on `pypi`.

diptorupd (approved):
Merging to master as TeamCity CI fails due to dpcpp_cpp_rt package not being there on pypi.

PR #510, opened 2021-07-15
Add a new DPCTLEvent_GetWaitList function

vlad-perevezentsev:
This PR adds the 'DPCTLEvent_GetWaitList' function which returns the list of events that this event waits for in the dependence graph.

### DIFF
--- a/dpctl-capi/include/dpctl_sycl_event_interface.h
+++ b/dpctl-capi/include/dpctl_sycl_event_interface.h
@@ -31,12 +31,17 @@
 #include "dpctl_data_types.h"
 #include "dpctl_sycl_enum_types.h"
 #include "dpctl_sycl_types.h"
+#include "dpctl_vector.h"
 
 DPCTL_C_EXTERN_C_BEGIN
 
 /**
  * @defgroup EventInterface Event class C wrapper
  */
+
+// Declares a set of types and functions to deal with vectors of
+// DPCTLSyclEventRef. Refer dpctl_vector_macros.h
+DPCTL_DECLARE_VECTOR(Event)
 
 /*!
  * @brief A wrapper for ``sycl::event`` contructor to construct a new event.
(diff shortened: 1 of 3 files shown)


dpctl-capi/include/dpctl_sycl_event_interface.h
```diff
@@ -91,4 +96,15 @@ DPCTLEvent_Copy(__dpctl_keep const DPCTLSyclEventRef ERef);
 DPCTL_API
 DPCTLSyclBackendType DPCTLEvent_GetBackend(__dpctl_keep DPCTLSyclEventRef ERef);
 
+/*!
+ * @brief  Return the vector of events that this event waits for
```
oleksandr-pavlyk, line 100:
```suggestion
 * @brief  Returns the vector of events that this event still waits for
```

oleksandr-pavlyk: We should also say this is a wrapper around ``sycl::event::get_wait_list``.

dpctl-capi/include/dpctl_sycl_event_interface.h
```diff
@@ -31,13 +31,18 @@
...
 #include "dpctl_sycl_types.h"
+#include "dpctl_vector.h"
 
 DPCTL_C_EXTERN_C_BEGIN
 
 /**
  * @defgroup EventInterface Event class C wrapper
  */
 
+// Declares a set of types abd functions to deal with vectors of
```
diptorupd, line 42: Minor typo “and” instead of “abd”

vlad-perevezentsev:
@oleksandr-pavlyk I added a Sycl test for function.
It works on SYCL but  test fails with DPCTLSyclEvent.
```
/localdisk/work/vperevez/dp_dparray/dpctl_vlad/dpctl/dpctl-capi/tests/test_sycl_event_interface.cpp:158: Failure
Value of: DPCTLEventVector_Size(EVRef) > 0
  Actual: false
Expected: true
```

Dependencies passed via ' sycl:: handler::depends_on` are not included in the graph, but are tracked (Compiler LLVM team).
This is a bug. https://jira.devtools.intel.com/browse/CMPLRLLVM-30013

oleksandr-pavlyk (changes requested):

oleksandr-pavlyk (approved):
I have applied the requested changes and resolved conflicts from earlier merges

PR #513, opened 2021-07-27
Add a new DPCTLEvent_WaitAndThrow function 

vlad-perevezentsev:
This PR adds a function `DPCTLEvent_GetWaitList` which waits for the event to complete and calls the asynchronous error handler

### DIFF
--- a/dpctl-capi/include/dpctl_sycl_event_interface.h
+++ b/dpctl-capi/include/dpctl_sycl_event_interface.h
@@ -58,6 +58,15 @@ DPCTL_API
 void DPCTLEvent_Wait(__dpctl_keep DPCTLSyclEventRef ERef);
 
 /*!
+ * @brief C-API wrapper for 'sycl::event.wait_and_throw'.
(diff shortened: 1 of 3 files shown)


vlad-perevezentsev:
@diptorupd  I added a sycl test that checks the call of the asynchronous error handler in the last commit.
Is it correct for this function?
Do I need to rewrite it via dpctl interface?
Can you tell me how better to write a test for this function?

oleksandr-pavlyk:
@vlad-perevezentsev please use backquotes rather than plain single quotes to delineate sycl object names.

oleksandr-pavlyk:
I suppose these tests will do for now, but I thought of using SYC C++ to create an event from submitting a kernel which causes a raise of an asynchronous exception to a queue with a custom handler that a side effect that a test checks for.

oleksandr-pavlyk (changes requested):

oleksandr-pavlyk (approved):

PR #514, opened 2021-07-27
Use dpcpp compiler package for Linux

PokhodenkoSA:
Now dpcpp_linux-64 compiler is available on `-c intel`. It is not necessary to install oneAPI.
This PR uses dpcpp compiler package instead of using oneAPI.
Only for Linux.

### DIFF
--- a/conda-recipe/build.sh
+++ b/conda-recipe/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ${PYTHON} setup.py clean --all
-${PYTHON} setup.py install
+${PYTHON} setup.py install --sycl-compiler-prefix=$CONDA_PREFIX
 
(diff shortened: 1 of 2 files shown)


Vyacheslav-Smirnov (approved):
Changes are OK from my side

AndresGuzman-Ballen:
Aw man, didn't anticipate having to have different behavior for wheel packages :( Since wheel packages do not handle anything environment-related like conda, it's very unlikely that DPC++ will be provided as a wheel package. Is the workaround you included in this PR acceptable to you all? Seems annoying but I don't think there's a better solution for it right now :(

PokhodenkoSA:
> Is the workaround you included in this PR acceptable to you all?

We will see soon.

PR #517, opened 2021-07-30
Add matrix for python version

PokhodenkoSA:
Similar to https://github.com/IntelPython/numba-dppy/pull/511 but for dpctl.
GitHub Actions conda uses python 3.9 by default. So we need to control python version.
This PR also:
- [x] improves caching for conda packages
- [x] makes consistency with numba-dppy workflow


PokhodenkoSA:
Close due to #518

PR #522, opened 2021-08-11
Add a `execution_status` method for the SyclEventRaw class

vlad-perevezentsev:
This PR adds the  `execution_status`  method which returns the value of the 'event_status_type' enumeration.

### DIFF
--- a/dpctl/_sycl_event.pyx
+++ b/dpctl/_sycl_event.pyx
@@ -29,9 +29,13 @@ from ._backend cimport (  # noqa: E211
     DPCTLEvent_Copy,
     DPCTLEvent_Create,
     DPCTLEvent_Delete,
+    DPCTLEvent_GetCommandExecutionStatus,
     DPCTLEvent_Wait,
     DPCTLSyclEventRef,
+    _event_status_type,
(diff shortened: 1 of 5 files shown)


dpctl/_sycl_event.pyx
```diff
@@ -202,3 +206,18 @@ cdef class SyclEventRaw(_SyclEventRaw):
             "SyclEventRef",
             &_event_capsule_deleter
         )
+
+    def execution_status(self):
```
oleksandr-pavlyk, line 210: Perhaps this should be a read-only property. 

```python
In [2]: dpctl.SyclEventRaw().execution_status
Out[2]: <function SyclEventRaw.execution_status>

In [3]: dpctl.SyclEventRaw().execution_status()
Out[3]: <event_status_type.complete: 4>
```

So one would use `In[2]` rather than `In[3]`.

oleksandr-pavlyk:
```suggestion
    @property
    def execution_status(self):
```

oleksandr-pavlyk (approved):

PR #523, opened 2021-08-12
Add 3 methods of the SyclEventRaw class for profiling

vlad-perevezentsev:
This PR adds 3 methods (`profiling_info_submit`. `profiling_info_start`. `profiling_info_end`) for the `SyclEventRaw` class that return a value describing the time in nanoseconds.


oleksandr-pavlyk (approved):
This needs docstrings. I am going to merge and add these in a separate PR though.

PR #527, opened 2021-08-16
Add a `wait_list` method for the SyclEventRaw class

vlad-perevezentsev:
This PR will add a `wait_list` method for the `SyclEventRaw` class which returns a list of events that the event is still waits for.


dpctl/_sycl_event.pyx
```diff
@@ -222,3 +227,20 @@ cdef class SyclEventRaw(_SyclEventRaw):
             return event_status_type.complete
         else:
             raise ValueError("Unknown event status.")
+
+    def wait_list(self):
```
oleksandr-pavlyk, line 231: Did you mean to use `@property` decorator?

oleksandr-pavlyk (approved):

PR #528, opened 2021-08-16
Cleanup/find intel sycl

oleksandr-pavlyk:
These are the changes needed to make support for `--sycl-compiler-prefix` on Windows. 

Conda recipe is modified to use that. Local build was successful.

### DIFF
--- a/conda-recipe/bld.bat
+++ b/conda-recipe/bld.bat
@@ -1,11 +1,9 @@
-call "%ONEAPI_ROOT%\compiler\latest\env\vars.bat"
-if errorlevel 1 (
-    echo "oneAPI compiler activation failed"
-    exit /b 1
-)
+
+REM A workaround for activate-dpcpp.bat issue to be addressed in 2021.4
+set "LIB=%CONDA_PREFIX%\Library\lib;%CONDA_PREFIX%\compiler\lib;%LIB%"
 
 "%PYTHON%" setup.py clean --all
-"%PYTHON%" setup.py install
+"%PYTHON%" setup.py install --sycl-compiler-prefix=%CONDA_PREFIX%\Library
 if errorlevel 1 exit 1
 
 rem Build wheel package
--- a/conda-recipe/bld.bat
+++ b/conda-recipe/bld.bat
@@ -3,7 +3,7 @@ REM A workaround for activate-dpcpp.bat issue to be addressed in 2021.4
 set "LIB=%CONDA_PREFIX%\Library\lib;%CONDA_PREFIX%\compiler\lib;%LIB%"
 
 "%PYTHON%" setup.py clean --all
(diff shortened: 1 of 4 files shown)


conda-recipe/bld.bat
```diff
@@ -1,11 +1,9 @@
-call "%ONEAPI_ROOT%\compiler\latest\env\vars.bat"
-if errorlevel 1 (
-    echo "oneAPI compiler activation failed"
-    exit /b 1
-)
+
+REM A workaround for activate-dpcpp.bat issue to be addressed in 2021.4
+set "LIB=%CONDA_PREFIX%\Library\lib;%CONDA_PREFIX%\compiler\lib;%LIB%"
```
AndresGuzman-Ballen, line 3: I should add that the bug is only for trying to use dpcpp in a virtual environment. The line above is within the following if statement in activate-dpcpp.bat:
    IF NOT "%CONDA_BUILD%" == ""
This line will always be true when conda-build is building the recipe using bld.bat. However if you have done something like 'conda create -n dpcpp_test dpcpp_win-64 vs2017_win-64 -c intel -c defaults --override-channels' and activate the environment, this line will be needed to be executed in order to successfully compile source code using the dpcpp executable

conda-recipe/bld.bat
```diff
@@ -1,11 +1,9 @@
...
 
 "%PYTHON%" setup.py clean --all
-"%PYTHON%" setup.py install
+"%PYTHON%" setup.py install --sycl-compiler-prefix=%CONDA_PREFIX%\Library
```
1e-to, line 6: Why is CONDA_PREFIX used here and in the file above? Shouldn't it be BUILD_PREFIX?

oleksandr-pavlyk:
```suggestion
"%PYTHON%" setup.py install --sycl-compiler-prefix=%BUILD_PREFIX%\Library
```

1e-to:
I tested it on my github hosted windows, the build was successful (--no-test --python 3.8)

oleksandr-pavlyk:
All CI builds failed due to an assertion failure in team-city integration script. @AndresGuzman-Ballen

conda-recipe/bld.bat
AndresGuzman-Ballen: @1e-to good eye :) yes this can remain "CONDA_PREFIX" (they are the same)

AndresGuzman-Ballen:
@oleksandr-pavlyk It failed once, yes but it seems like that issue has gone away: https://teamcity-or.intel.com/viewType.html?buildTypeId=Sat_NewCi_IndividualPackages_Dpctl_Development_Python37_Linux_Building. Notice how pull/529 was able to build.

PR #529, opened 2021-08-17
Add a static method `wait` for the SyclEventRaw class

vlad-perevezentsev:
This PR adds a static method `wait` for the SyclEventRaw class with a call to `DPCTLEvent_WaitAndThrow`

### DIFF
--- a/dpctl/_sycl_event.pyx
+++ b/dpctl/_sycl_event.pyx
@@ -37,6 +37,7 @@ from ._backend cimport (  # noqa: E211
     DPCTLEvent_GetProfilingInfoSubmit,
     DPCTLEvent_GetWaitList,
     DPCTLEvent_Wait,
+    DPCTLEvent_WaitAndThrow,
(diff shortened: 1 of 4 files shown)


dpctl/_sycl_event.pyx
```diff
@@ -192,8 +193,20 @@ cdef class SyclEventRaw(_SyclEventRaw):
...
 
-    cpdef void wait(self):
-        DPCTLEvent_Wait(self._event_ref)
+    @staticmethod
+    cdef void _wait(SyclEventRaw event):
+        DPCTLEvent_WaitAndThrow(event._event_ref)
+
+    @staticmethod
+    def wait(event):
+        if isinstance(event, list):
```
oleksandr-pavlyk, line 202: I think this should work even if `event` is a tuple, or any iterable object of events (need not be a list).

oleksandr-pavlyk (approved):

PR #538, opened 2021-08-19
Conda build scripts should build wheels in the same setup invocation …

oleksandr-pavlyk:
…as install

Using

```
python setup.py cmd1 opts1 cmd2 opts ...
```


AndresGuzman-Ballen (approved):
LG2M!

oleksandr-pavlyk:
Linux CI builds passed, windows CI were hanging. Checking locally revealed issues caused by quoting at the wrong place. 

Once fixed, both wheel-output and regular install steps succeeded fine.

PR #539, opened 2021-08-19
Fixed dpctl_config, added dpctl_service.h, .cpp

oleksandr-pavlyk:
Fixed Config/dpctl_config.h.in so that `DPCTL_DPCPP_VERSION` preprocessor
variable is now non-empty.

Created `const char * DPCTLService_GetDPCPPVersion(void)` service function
to get the version stored in that preprocessor variable.

This would allow Python to record minimum version requirement for the
DPC++ toolchain runtime.

```
Python 3.7.10 (default, Jun  4 2021, 06:52:02)
IPython 7.25.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import dpctl, ctypes

In [2]: lib = ctypes.cdll.LoadLibrary('dpctl/libDPCTLSyclInterface.so')

In [3]: fn = lib.DPCTLService_GetDPCPPVersion

In [4]: fn.argtypes=[]

In [5]: fn.restype = ctypes.c_char_p

In [6]: fn()
Out[6]: b'2021.3.0'
```

### DIFF
--- a/dpctl-capi/include/dpctl_service.h
+++ b/dpctl-capi/include/dpctl_service.h
@@ -1,4 +1,4 @@
-//===--------- dpctl_config.h - Configured options for dpctl C API         ===//
+//===- dpctl_service.h - C API for service functions   -*-C++-*- ===//
 //
 //                      Data Parallel Control (dpctl)
 //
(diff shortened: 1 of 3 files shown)


AndresGuzman-Ballen (dismissed):
Pretty cool!

PR #543, opened 2021-08-23
Add the SyclTimer class to the examples

vlad-perevezentsev:
This PR adds a python `sycl_timer file` containing the `SyclTimer` class and the `dppy-kernel file` as an example for working with this class

### DIFF
--- a/examples/python/sycl_timer.py
+++ b/examples/python/sycl_timer.py
@@ -1,0 +1,62 @@
+#                      Data Parallel Control (dpctl)
+#
+# Copyright 2020-2021 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


examples/python/sycl_timer.py
```diff
@@ -0,0 +1,62 @@
...
+    def dt(self):
+        self.event_start.wait()
+        self.event_finish.wait()
+        return (
+            (self.host_finish - self.host_start) * self.time_scale,
+            (
+                self.event_finish.profiling_info_start
+                - self.event_start.profiling_info_end
+            )
+            / 1e9
```
oleksandr-pavlyk, line 60: I think I would right it as 

```
     ) * 1e-9 * self.time_scale,
)
```

but this a  nitpick.

oleksandr-pavlyk (approved):

examples/python/sycl_timer.py
```diff
@@ -0,0 +1,62 @@
...
+# limitations under the License.
+
+
+import timeit
+
+import dpctl
+
+
+class SyclTimer:
+    def __init__(self, host_time=timeit.default_timer, time_scale=1):
```
oleksandr-pavlyk, line 24: Doc string would be desired here, with a usage example.

oleksandr-pavlyk (approved):

PR #555, opened 2021-08-26
Moved sycl timer into dpctl.SyclTimer

oleksandr-pavlyk:
Added docstring, made dt into a property (rather than a method).

Removed stray "SyclEventRaw" from __all__ in dpctl/__init__.py


oleksandr-pavlyk:
@vlad-perevezentsev, ping

vlad-perevezentsev (approved):

PR #558, opened 2021-08-30
Add aspect selector

vlad-perevezentsev:
This PR adds the device selection function by aspects and tests to it.

Docstrings will be added in the next commit.

### DIFF
--- a/dpctl/_device_selection.py
+++ b/dpctl/_device_selection.py
@@ -1,0 +1,29 @@
+from . import SyclDevice, get_devices
+
+
+def select_device_with_aspects(aspect_list, deny_list=[]):
+    check_list = aspect_list + deny_list
+    for asp in check_list:
+        if type(asp) != str:
+            raise TypeError("The list objects must be of a string type")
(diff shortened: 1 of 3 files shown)


dpctl/_device_selection.py
```diff
@@ -0,0 +1,29 @@
...
+            raise TypeError("The list objects must be of a string type")
+        if not hasattr(SyclDevice, "has_aspect_" + asp):
+            raise ValueError(f"The {asp} aspect is not supported in dpctl")
+    devs = get_devices()
+    max_score = 0
+    selected_dev = None
+
+    for dev in devs:
+        aspect_status = True
+        for asp in aspect_list:
```
oleksandr-pavlyk, line 17: Instead of explicitly iterating and using `eval` it is better to do

```python
aspect_status = all((getattr(dev, "has_aspect_" + asp) is True for asp in aspect_list))
aspect_status = aspect_status and not(any((getattr(dev, "has_aspect_" + asp) is True for asp in deny_list))
```

oleksandr-pavlyk:
I think this is ready to go in.

PR #563, opened 2021-09-01
Improve coverage of dpctl.__init__ file and other service functions

oleksandr-pavlyk:
Added tests to check `DPCTLService_GetDPCPPVersion`, `dpctl.get_include`, and `dpctl.__version__`.

### DIFF
--- a/dpctl/tests/test_service.py
+++ b/dpctl/tests/test_service.py
@@ -1,0 +1,95 @@
+#                      Data Parallel Control (dpctl)
+#
+# Copyright 2020-2021 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


oleksandr-pavlyk:
@PokhodenkoSA Any idea why this test is failing 

```
    def test___version__():
        dpctl_ver = getattr(dpctl, "__version__", None)
        assert type(dpctl_ver) is str
>       assert "unknown" not in dpctl_ver
E       AssertionError: assert 'unknown' not in '0+unknown'
E         'unknown' is contained here:
E           0+unknown

/usr/share/miniconda/lib/python3.8/site-packages/dpctl/tests/test_service.py:88: AssertionError
=========================== short test summary info ============================
FAILED tests/test_service.py::test___version__ - AssertionError: assert 'unkn...
================= 1 failed, 785 passed, 1051 skipped in 2.56s ==================
Error: Process completed with exit code 1.
```

I think the expectation for it to not contain "unknown" is reasonable, and addition of this test has caught an actual issue.

oleksandr-pavlyk:
> @PokhodenkoSA Any idea why this test is failing
> 
> I think the expectation for it to not contain "unknown" is reasonable, and addition of this test has caught an actual issue.

This is likely caused by using limited depth cloning, specifically use of `- fetch_depth: 0` in [workflows](https://github.com/IntelPython/dpctl/blob/master/.github/workflows/conda-package.yml#L28). 

If we rely on git tags for deducing versions of the product, we should do a full checkout.

oleksandr-pavlyk:
Turns out this has been a regression I introduced nearly at the beginning of 0.9 development cycle. Now fixed, tests should start passing.

PR #578, opened 2021-09-08
Add methods __bool__, __float__ and __int__ to usm_ndarray

### LINKED ISSUE #545: 0d usm_ndarray must allow conversion to Python scalars
```
In [2]: import dpctl

In [3]: import dpctl.tensor as dpt

In [4]: dpt.usm_ndarray(tuple(), dtype='d')
Out[4]: <dpctl.tensor._usmarray.usm_ndarray at 0x7f562c2859f0>

In [5]: float(_)
<ipython-input-5-a06b00cb7231> in <module>
----> 1 float(_)

TypeError: float() argument must be a string or a number, not 'dpctl.tensor._usmarray.usm_ndarray'
```

In[5] should work. 

Same for Python integers, and Python bools.

densmirn:
Closes #545

### DIFF
--- a/dpctl/memory/_sycl_usm_array_interface_utils.pxi
+++ b/dpctl/memory/_sycl_usm_array_interface_utils.pxi
@@ -2,7 +2,7 @@
 
 cdef bint _valid_usm_ptr_and_context(DPCTLSyclUSMRef ptr, SyclContext ctx):
     usm_type = _Memory.get_pointer_type(ptr, ctx)
-    return usm_type in (b'shared', b'device', b'host')
+    return usm_type in (b"shared", b"device", b"host")
 
 
 cdef DPCTLSyclQueueRef _queue_ref_copy_from_SyclQueue(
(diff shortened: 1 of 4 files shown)


oleksandr-pavlyk:
@densmirn can you also implement `__index__`?

See https://github.com/data-apis/array-api/issues/231

oleksandr-pavlyk:
```python
import dpctl.tensor as dpt
X = dpt.usm_ndarray(10, 'd')
float(X[:1])   # raises TypeError
```

This works with NumPy:

```python
import numpy as np
X = np.ndarray(10, 'd')
float(X[:1])   # produces a float 
```

It breaks with the current code because `Xusm.usm_data` gives the container underlying the view which may be bigger than the view:

```python
X = dpt.usm_ndarray(10, 'd')
X[:1].usm_data.copy_to_host().view('d').shape  # outputs (10,)
```

Thus instead of `self.usm_data.copy_to_host().view('d').__float__()` we should do `self.usm_data.copy_to_host().view(self.dtype)[:0].__float__()`.

Ideally, in a future, we should avoid copying the entire underlying buffer.

densmirn:
Actually slice of array can be different, e.g.:
```python
import numpy as np
X = np.arange(10)
float(X[:1])  # 0.0
float(X[2:3])  # 2.0
```
So when we copy data to host we don't know the slice:
`self.usm_data.copy_to_host().view(self.dtype)[???].__float__()`

Looks like it should be somehow handled on `__getitem__` side.

densmirn:
@oleksandr-pavlyk could you please look at last comment?

oleksandr-pavlyk:
This is affected by #583

oleksandr-pavlyk (approved):

oleksandr-pavlyk:
This PR(tests) have uncovered another problem which I fixed. 

It was inability of memory objects to understand objects with `__sycl_usm_array_interface__` with  `'typestr': '|b1'`.

Also added tests to cover new functionality.

oleksandr-pavlyk:
I also added `__complex__` [magic (a.k.a. dunder) method](https://rszalski.github.io/magicmethods/).

PR #580, opened 2021-09-09
Use renamed DPC++ runtime package `dpcpp-cpp-rt`

PokhodenkoSA:
From @tomashek (SAT-4412):
> dpcpp_cpp_rt was renamed to dpcpp-cpp-rt. We need to make sure we are using the new ones. The old are still around for backward compatibility.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -28,7 +28,7 @@ requirements:
     run:
         - python
         - numpy >=1.17
-        - dpcpp_cpp_rt >=2021.2
+        - dpcpp-cpp-rt >=2021.2
 
 test:
     requires:


oleksandr-pavlyk (dismissed):
My only concern is that this renaming should not break public CI jobs

PokhodenkoSA:
My concern is that both packages `dpcpp-cpp-rt` and `dpcpp_cpp_rt` are installed. Let's wait for comments from @tomashek or @AndresGuzman-Ballen.

AndresGuzman-Ballen:
@oleksandr-pavlyk do you say there is no effect because during the build, you didn't see it pull 'dpcpp-cpp-rt' instead of dpcpp_cpp_rt'? This PR only changes the runtime package and what you see in the build logs is being brought in from the DPC++ compiler conda packages, so I wouldn't expect there to be any difference in CI during build time. Perhaps during the test phase, it would then be surprising that there was no effect.

AndresGuzman-Ballen:
@PokhodenkoSA having both installed at the same time isn't a problem at all since one is a metapackage for the other. There's no chance you'll experience clobbering, which I assume is your concern?

oleksandr-pavlyk:
> @oleksandr-pavlyk do you say there is no effect because during the build, you didn't see it pull 'dpcpp-cpp-rt' instead of dpcpp_cpp_rt'? 

I misspoke. I looked through logs of public jobs and it seems to work fine. I was simply calling attention to this aspect of the change. It would be wrong to make this change in the public repo if it was to break public CI jobs.

PR #586, opened 2021-09-15
usm_ndarray object special methods, and other changes 

oleksandr-pavlyk:
Use of `public` keyword when declaring `usm_ndarray` Cython class amounted to Cython exposing private struct members to Python (such `usm_ndarray.typenum_`, `usm_ndarray.nd_`, etc)

Keeping only `api` (no ``public``), these members are not longer exposed, while the python object and its type remain exposed.

Added ``__pyx_capi__`` functions to access these private members, paving the road to using these classes in C/C++ extensions, e.g. built with help of pybind11.

Added tests to test these ``__pyx_capi__`` functions.

This PR has also grown to add `__setitem__` special method to `usm_ndarray` class, as well as `dpctl.tensor.astype(array, new_dtype, order="K", copy=True)` and `dpctl.tensor.reshape(array, new_shape, order="C")`.

`usm_ndarray` can also be reshaped using `shape` setter if this can be done as a view.

### DIFF
--- a/dpctl/tensor/__init__.py
+++ b/dpctl/tensor/__init__.py
@@ -29,8 +29,16 @@
 
 """
 
+from dpctl.tensor._copy_utils import astype
+from dpctl.tensor._copy_utils import copy_from_numpy as from_numpy
+from dpctl.tensor._copy_utils import copy_to_numpy as to_numpy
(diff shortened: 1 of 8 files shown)


Alexander-Makaryev (approved):
👍

PR #589, opened 2021-09-18
Device memory sizes

### LINKED ISSUE #587: SyclDevice should expose global_mem_size, local_mem_size info descriptor
These would correspond to 

```c++
inline std::uint64_t get_device_global_mem_size(const sycl::device &d) {
    return d.get_info<sycl::info::device::global_mem_size>();
}

inline std::uint64_t get_device_local_mem_size(const sycl::device &d) {
    return d.get_info<sycl::info::device::local_mem_size>();
}
```

oleksandr-pavlyk:
Closes #587

Adds `DPCTLDevice_GetGlobalMemSize(DRef)`, `DPCTLDevice_GetLocalMemSize(DRef)`.

Also adds `dpctl.SyclDevice.global_mem_size` and `dpctl.SyclDevice.local_mem_size`.


oleksandr-pavlyk:
@diptorupd @reazulhoque Not sure when coveralls comes back to being functional, but can you look through the code and try it out?

PR #593, opened 2021-09-21
isort GH action changed and no longer accepts `args` input

oleksandr-pavlyk:
Changed to use configuration input to specify CLI arguments.

Prompted by warning in https://github.com/IntelPython/dpctl/actions/runs/1249104763

image


diptorupd (approved):
LGTM!

PR #596, opened 2021-09-26
Added new example of Python object exposing __sycl_usm_array_interface__

oleksandr-pavlyk:
Native Pybind11-generated extension implemented Python type bound to
`DMatrix` C++ class which allocates USM memory using ``sycl::usm_allocator``.

The Python object implements ``__sycl_usm_array_interface__``, which allows
dpctl.memory.as_usm_memory to create a view into that native USM allocation.
The example.py modifies that memory from Python, and uses object's own
`.tolist()` method to retrieve the memory using C++, demonstrating that
values changed.

Also added license missing headers to .cpp and .hpp files from other
examples

### DIFF
--- a/examples/pybind11/external_usm_allocation/_usm_alloc_example.cpp
+++ b/examples/pybind11/external_usm_allocation/_usm_alloc_example.cpp
@@ -1,0 +1,161 @@
+//==- _usm_alloc_example.cpp - Example of Pybind11 extension exposing   --===//
+//   native USM allocation to Python in such a way that dpctl.memory
+//   can form views into it.
+//
+//                      Data Parallel Control (dpctl)
+//
+// Copyright 2020-2021 Intel Corporation
+//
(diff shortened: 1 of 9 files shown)


examples/pybind11/external_usm_allocation/_usm_alloc_example.cpp
```diff
@@ -0,0 +1,161 @@
...
+    sycl::queue q_;
+    alloc_t alloc_;
+    vec_t vec_;
+};
+
+DMatrix create_matrix(py::object queue, size_t n, size_t m)
+{
+    PyObject *queue_ptr = queue.ptr();
+    if (PyObject_TypeCheck(queue_ptr, &PySyclQueueType)) {
+        DPCTLSyclQueueRef QRef =
```
diptorupd, line 94: Do not we need to free the `QRef`?

PR #598, opened 2021-09-27
Use safe load library functionality for Windows

PokhodenkoSA:
Based on code from oneDAL:
https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/src/services/service_daal_load_win_dynamic_lib.cpp

### DIFF
--- a/dpctl-capi/helper/source/dpctl_dynamic_lib_helper.cpp
+++ b/dpctl-capi/helper/source/dpctl_dynamic_lib_helper.cpp
@@ -1,19 +1,22 @@
-/* file: service_daal_load_win_dynamic_lib.cpp */
-/*******************************************************************************
-* Copyright 2021 Intel Corporation
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with the License.
-* You may obtain a copy of the License at
-*
-*     http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


tomashek:
Thank you @PokhodenkoSA

dpctl-capi/helper/source/dpctl_dynamic_lib_helper.cpp
```diff
@@ -0,0 +1,189 @@
...
+               (unsigned int)sverif);
+        break;
+    }
+
+    pWVTData.dwStateAction = WTD_STATEACTION_CLOSE;
+    WinVerifyTrust(NULL, &pgActionID, &pWVTData);
+    delete[] wPathBuf;
+
+    if (ERROR_SUCCESS != sverif) {
+        return NULL;
```
oleksandr-pavlyk, line 172: Should we have restored the state of the search path as well here?

oleksandr-pavlyk:
```suggestion
        // Restore current directory from the search path
        SetDllDirectory(NULL);
        return NULL;
```

oleksandr-pavlyk:
I have applied the change I suggested. It is the right thing to do according to [documentation page of SetDllDirectory](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setdlldirectorya).

I think this PR should graduate from the draft, and be merged. 

Is this code was borrowed from DAL, please communicate the proposed change to them as well.

PR #606, opened 2021-09-29
Silly typo behind slow copy

oleksandr-pavlyk:
The following example used to be very slow

```python
import dpctl.tensor as dpt
X = dpt.usm_ndarray((1_000_000, 4), "d")
Z1 = dpt.usm_ndarray((1_000, 4), "d")
X[:1_000] = Z1
```

due to missing return after early optimization call.


Alexander-Makaryev (approved):
👍

PR #612, opened 2021-09-30
Avoid passing nullptr to sycl::queue::prefetch to avoid buggy hang

oleksandr-pavlyk:
For bookkeeping filed an issue behind the hang as XDEPS-2765. 

Fix for the hang is to avoid passing nullptr argument to ``sycl::queue::prefetch``.

### DIFF
--- a/dpctl-capi/source/dpctl_sycl_queue_interface.cpp
+++ b/dpctl-capi/source/dpctl_sycl_queue_interface.cpp
@@ -514,15 +514,23 @@ DPCTLSyclEventRef DPCTLQueue_Prefetch(__dpctl_keep DPCTLSyclQueueRef QRef,
 {
     auto Q = unwrap(QRef);
     if (Q) {
-        sycl::event ev;
-        try {
-            ev = Q->prefetch(Ptr, Count);
-        } catch (sycl::runtime_error &re) {
+        if (Ptr) {
+            sycl::event ev;
+            try {
+                ev = Q->prefetch(Ptr, Count);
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
Very likely this hang was the reason why the Generate coverage data workflow never got to building the Python extensions.

oleksandr-pavlyk:
> Very likely this hang was the reason why the Generate coverage data workflow never got to building the Python extensions.

No the failure is due to a problem with `genhtml`:

```
2021-09-30T18:48:05.7792522Z genhtml: ERROR: cannot read /tmp/test_service-c3cfab.cpp
2021-09-30T18:48:05.7793677Z make[3]: *** [tests/CMakeFiles/lcov-genhtml.dir/build.make:72: tests/CMakeFiles/lcov-genhtml] Error 2
2021-09-30T18:48:05.7794817Z make[2]: *** [CMakeFiles/Makefile2:207: tests/CMakeFiles/lcov-genhtml.dir/all] Error 2
2021-09-30T18:48:05.7795883Z make[1]: *** [CMakeFiles/Makefile2:214: tests/CMakeFiles/lcov-genhtml.dir/rule] Error 2
2021-09-30T18:48:05.7796749Z make: *** [Makefile:221: lcov-genhtml] Error 2
```

On the next line of the workflow, we are executing import of dpctl, which fails

```
2021-09-30T18:48:05.7817325Z ModuleNotFoundError: No module named 'dpctl._sycl_context'
```

but this goes unnoticed, and the worflow script proceeds with running pytest.

PR #603, opened 2021-09-28
Remove use of DPC++ features deprecated in 2021.4 and open source intel/llvm compiler

oleksandr-pavlyk:
Also replaced ``sycl::ONEAPI::filter_selector`` with ``sycl::ext::oneapi::filter_selector``

The packages compiles using open source SYCL compiler after these changes, specifically I used

```
(py39) [13:20:15 ansatnuc04 dpctl]$ clang++ --version
clang version 14.0.0 (https://github.com/intel/llvm 7c17ad9a2bfbf6ff2757f8c7fbd2939841f35fa5)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /localdisk/work/opavlyk/sycl_workspace/llvm/build/bin
```

Only one meaningful failure in the so compiled package:

```
>       dpctl_prog.create_program_from_spirv(q, spirv)

dpctl/tests/test_sycl_program.py:93:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
dpctl/program/_program.pyx:178: in dpctl.program._program.create_program_from_spirv
    cpdef create_program_from_spirv(SyclQueue q, const unsigned char[:] IL,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   raise SyclProgramCompilationError()
E   dpctl.program._program.SyclProgramCompilationError

dpctl/program/_program.pyx:209: SyclProgramCompilationError
--------------------------------------------------------------------------- Captured stderr call ----------------------------------------------------------------------------
The level zero loader dynamic library could not be opened.
ZeModuleCreateFn is invalid.
========================================================================== short test summary info ==========================================================================
[...]

PR #617, opened 2021-10-03
Use readelf on pi_level_zero library to find exact name of ze_loader

oleksandr-pavlyk:
Store that in include/Config/dpctl_config.h and use the define in
dpctl_sycl_program_interface.cpp

CMake is running the equivalent of the following

```
readelf -d $DPCPP_ROOT/lib/libpi_level_zero.so | grep libze_loader | grep -Po "libze_loader[^\]]*"
libze_loader.so.1
```

This change, together with #603 and #616 make project buildable with OS compiler.

### DIFF
--- a/dpctl-capi/CMakeLists.txt
+++ b/dpctl-capi/CMakeLists.txt
@@ -50,6 +50,22 @@ if(DPCTL_ENABLE_LO_PROGRAM_CREATION)
     set(DPCTL_ENABLE_LO_PROGRAM_CREATION 1)
     include(GetLevelZeroHeaders)
     get_level_zero_headers()
+    if (UNIX)
+        find_library(PI_LEVEL_ZERO_LIB
+            NAMES pi_level_zero
+            HINTS ${IntelSycl_LIBRARY_DIR}
+        )
+        find_program(READELF_PROG readelf)
+        find_program(GREP_PROG grep)
+        execute_process(
+            COMMAND ${READELF_PROG} -d ${PI_LEVEL_ZERO_LIB}
+            COMMAND ${GREP_PROG} libze_loader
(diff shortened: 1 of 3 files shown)


dpctl-capi/CMakeLists.txt
```diff
@@ -50,6 +50,22 @@ if(DPCTL_ENABLE_LO_PROGRAM_CREATION)
     set(DPCTL_ENABLE_LO_PROGRAM_CREATION 1)
     include(GetLevelZeroHeaders)
     get_level_zero_headers()
+    if (UNIX)
+      find_library(PI_LEVEL_ZERO_LIB
+	NAMES pi_level_zero
```
diptorupd, line 55: Something is off with indentation here.

dpctl-capi/CMakeLists.txt
```diff
@@ -50,6 +50,22 @@ if(DPCTL_ENABLE_LO_PROGRAM_CREATION)
...
+	HINTS ${IntelSycl_LIBRARY_DIR}
+	)
+      find_program(READELF_PROG readelf)
+      find_program(GREP_PROG grep)
+      execute_process(
+	COMMAND ${READELF_PROG} -d ${PI_LEVEL_ZERO_LIB}
```
diptorupd, line 61: indentation is off.

diptorupd (commented):
Thanks for getting this done. Except for the minor indentation issues that need to be addressed, the PR is good to go.

dpctl-capi/CMakeLists.txt
oleksandr-pavlyk: I fixed the indentation.

diptorupd (approved):
LGTM!

PR #618, opened 2021-10-04
Rework the logic in memory's copy_from_device

oleksandr-pavlyk:
In deciding where a copy from device could be done using a direct
call to memcpy, we queried usm type of the source allocation with
respect to the context to which the destination allocation was
bound.

The test was assuming that any answer other than 'unknown' indicated
that the USM allocation was known and a memcpy could be called.

This did not work well when destination was a host device. For example,
this crashed:

```
SYCL_ENABLE_HOST_DEVICE=1 python -m pytest dpctl/tests/test_usm_ndarray_ctor.py::test_to_device
```

It passes now.


oleksandr-pavlyk:
Good to go in.

PR #542, opened 2021-08-20
DPCTLSyclInterface should avoid functions that print to std::cout

oleksandr-pavlyk:
The stream `std::cout` that the library writes to in a function like
`DPCTLDeviceMgr_PrintDeviceInfo` is not the same as Python's `sys.stdout`.

It makes it difficult to verify that printed output is what should be
expected.

Used `OutputGrabber` class to try to capture `std::cout` outputs of `DPCTLSyclInterface`.
The class is implemented in new `test/_redirector.py` file.

The changes in this commit allow the test to pass, while ensuring
that the captured output is non-empty, but pytest must be envoked with
`--capture=no` option.

It works as follows:

```
(idp) [14:36:03 ansatnuc04 dpctl]$ ipython
Python 3.7.10 (default, Jun  4 2021, 06:52:02)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.25.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import dpctl

In [2]: q = dpctl.SyclQueue()

In [3]: q.print_device_info()
    Name            Intel(R) UHD Graphics [0x9bca]
    Driver version  1.1.20389
    Vendor          Intel(R) Corporation
    Profile         FULL_PROFILE
    Filter string   level_zero:gpu:0

In [4]: import tests._redirector

In [5]: out = tests._redirector.OutputGrabber()

In [6]: with out: q.print_device_info()

In [7]: out.capturedtext
Out[7]: '    Name            Intel(R) UHD Graphics [0x9bca]\n    Driver version  1.1.20389\n    Vendor          Intel(R) Corporation\n    Profile         FULL_PROFILE\n    Filter string   level_zero:gpu:0\n'
```

It is all good, but `In[3]` for needed for `Out[7]` to be non-empty.

```
(idp) [14:38:36 ansatnuc04 dpctl]$ ipython
Python 3.7.10 (default, Jun  4 2021, 06:52:02)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.25.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import dpctl

In [2]: q = dpctl.SyclQueue()

In [3]: import tests._redirector

In [4]: out = tests._redirector.OutputGrabber()

In [5]: with out: q.print_device_info()

In [6]: out.capturedtext
Out[6]: ''
```

This is because `libDPCTLSyclInterface` does not initialize `std::cout`, and In[5] writes into a different `std::cout` that output grabber captured.

PR #620, opened 2021-10-05
Add a new utility function to return the device info as a C string object.

diptorupd:
The PR adds a new helper function in dpctl_sycl_device_manager called `DPCTLDeviceMgr_GetDeviceInfoStr`. The function returns a C string containing the device attributes currently printed out by `DPCTLDeviceMgr_PrintDeviceInfo`. The change will unblock #542.

Added a helper to convert C++ strings to C strings, Reduces code duplication.

### DIFF
--- a/dpctl-capi/include/dpctl_sycl_device_manager.h
+++ b/dpctl-capi/include/dpctl_sycl_device_manager.h
@@ -77,6 +77,22 @@ __dpctl_give DPCTLDeviceVectorRef
 DPCTLDeviceMgr_GetDevices(int device_identifier);
 
 /*!
+ * @brief Returns a set of device info attributes as a string.
+ *
+ * @param    DRef           Opaque pointer to a ``sycl::device``
+ * @return   A formatted C string capturing the following attributes:
+ *             - device name
+ *             - driver version
+ *             - vendor
+ *             - profiler support
+ *             - oneapi filter string
+ * @ingroup DeviceManager
+ */
+DPCTL_API
+__dpctl_give const char *
+DPCTLDeviceMgr_GetDeviceInfoStr(__dpctl_keep const DPCTLSyclDeviceRef DRef);
+
+/*!
  * @brief Returns an index on the given device in the vector returned by
  * #DPCTLDeviceMgr_GetDevices if found, -1 otherwise.
  *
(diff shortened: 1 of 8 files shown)


dpctl-capi/include/dpctl_sycl_device_manager.h
```diff
@@ -76,6 +76,22 @@ DPCTL_API
 __dpctl_give DPCTLDeviceVectorRef
 DPCTLDeviceMgr_GetDevices(int device_identifier);
 
+/*!
+ * @brief Returns as a string a set of device info attributes.
```
oleksandr-pavlyk, line 80:
```suggestion
 * @brief Returns a set of device info attributes as a string.
```

dpctl-capi/source/dpctl_sycl_device_manager.cpp
```diff
@@ -168,6 +168,32 @@ DPCTLDeviceMgr_GetDevices(int device_identifier)
...
+    if (D) {
+        try {
+            auto infostr = get_device_info_str(*D);
+            auto cstr_len = infostr.length() + 1;
+            cstr_info = new char[cstr_len];
+#ifdef _WIN32
+            strncpy_s(cstr_info, cstr_len, infostr.c_str(), cstr_len);
+#else
+            std::strncpy(cstr_info, infostr.c_str(), cstr_len);
+#endif
```
oleksandr-pavlyk, line 185: We have code snippets such as this sprinkled around the codebase. I think we should encapsulate them into a callable function and use that.

oleksandr-pavlyk (approved):
LGTM with small suggestions.

oleksandr-pavlyk:
By the way, the Python API should be changed too, in `_sycl_device.pyx`

diptorupd:
> 
> 
> By the way, the Python API should be changed too, in `_sycl_device.pyx`

Let us add that in #542

dpctl-capi/source/dpctl_sycl_device_manager.cpp
diptorupd: agreed. I will make the change.

dpctl-capi/include/dpctl_sycl_device_manager.h
diptorupd: Fixed.

dpctl-capi/source/dpctl_sycl_device_manager.cpp
diptorupd: Done in a58afe9c11e2a9bf8352af85fbde1ce9502f8c2d

oleksandr-pavlyk (approved):
Looks good.

PR #621, opened 2021-10-05
Add a workflow to install nightly intel/llvm + drivers

oleksandr-pavlyk:
Build dpctl with Open Source DPC++ toolchain. 

1. Download and unpack nightly build of SYCL bundle
2. Download and unpack OpenCL CPU driver
3. Download and unpack OpenCL FPGA emulator
4. Download and unpack oneAPI TBB
5. Set-up LD_LIBRARY_PATH and PATH
6. Build `dpctl` and run `pytest`.


.github/workflows/os-llvm-sycl-build.yml
```diff
@@ -0,0 +1,77 @@
...
+          mkdir -p sycl_bundle
+          cd sycl_bundle
+          git clone http://github.com/intel/llvm.git llvm_src
+          cd llvm_src
+          alias urlencode='python3 -c "import sys, urllib.parse as ul; print (ul.quote_plus(sys.argv[1]))"'
+          export NIGHTLY_TAG=$(urlencode $(git tag | sort -V | tail -1))
+          cd ..
+          wget https://github.com/intel/llvm/releases/download/${NIGHTLY_TAG}/dpcpp-compiler.tar.gz
+          wget https://github.com/intel/llvm/releases/download/2021-07/${OCLCPUEXP_FN}
+          wget https://github.com/intel/llvm/releases/download/2021-07/${FPGAEMU_FN}
```
diptorupd, line 36: copy-pasta?

oleksandr-pavlyk: URL prefixes are the same, I pushed a change to use env variable to store this prefix.

.github/workflows/os-llvm-sycl-build.yml
```diff
@@ -0,0 +1,84 @@
...
+  push:
+    branches: [master]
+
+jobs:
+  install-compiler:
+    name: Install Nightly build of DPC++ toolchain
+    runs-on: ubuntu-20.04
+
+    env:
+      OCLCPUEXP_FN: oclcpuexp-2021.12.6.0.19_rel.tar.gz
```
diptorupd, line 13: Is there a way to populate these values dynamically? Otherwise, we will soon be using a older driver and not be up-to-date with latest release.

oleksandr-pavlyk: intel/llvm uses https://github.com/intel/llvm/blob/sycl/buildbot/dependency.conf

.github/workflows/os-llvm-sycl-build.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: |
+          cd /home/runner/work
+          mkdir -p sycl_bundle
+          cd sycl_bundle
+          git clone http://github.com/intel/llvm.git llvm_src
+          cd llvm_src
+          export NIGHTLY_TAG=$(python3 -c "import sys, urllib.parse as ul; print (ul.quote_plus(sys.argv[1]))" $(git tag | sort -V | tail -1))
+          cd ..
+          export DOWNLOAD_URL_PREFIX=https://github.com/intel/llvm/releases/download
+          wget ${DOWNLOAD_URL_PREFIX}/${NIGHTLY_TAG}/dpcpp-compiler.tar.gz
```
diptorupd, line 34: We can cache these downloaded copies to speed up the workflow.

oleksandr-pavlyk: This would be great. I just need to know where to look for an example. GTest installation was being cached, so I will look there.

diptorupd: You do not need a conf file. As we are storing the cloned repo we might as well run `git tag | sort -V | tail -1` and compare the SHA with what the command I pasted returns.

PR #623, opened 2021-10-06
Correct the logic to return device count.

### LINKED ISSUE #601: `dpctl.has_gpu_devices()` returns False even when GPU devices are present
```
In [1]: import dpctl

In [2]: dpctl.has_gpu_devices()
Out[2]: False

In [3]: dpctl.lsplatform()
Intel(R) OpenCL OpenCL 3.0 LINUX
Intel(R) OpenCL HD Graphics OpenCL 3.0
Intel(R) Level-Zero 1.1
SYCL host platform 1.2

In [4]: dpctl.get_devices()
Out[4]:
[<dpctl.SyclDevice [backend_type.opencl, device_type.cpu,  Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz] at 0x7f1dbd1619f0>,
 <dpctl.SyclDevice [backend_type.opencl, device_type.gpu,  Intel(R) UHD Graphics [0x9bca]] at 0x7f1dbcc201b0>,
 <dpctl.SyclDevice [backend_type.level_zero, device_type.gpu,  Intel(R) UHD Graphics [0x9bca]] at 0x7f1dbcc20e30>]

In [5]: dpctl.has_gpu_devices()
Out[5]: False

In [6]: dpctl.has_cpu_devices()
Out[6]: False

In [7]: dpctl.has_accelerator_devices()
Out[7]: False
```

diptorupd:
We were not handling the cases where a device identifier
does not provide a backend or a device type. Changed the logic
to count all devices of a given type if no backend is specified.
Similarly, all devices in a backend are counted if no device
type is provided.

Fixes #601


diptorupd:
@oleksandr-pavlyk I found what the issue was. The existing logic returned zero for a device type if no backend was specified. I fixed that. However, we need some good way of testing it n both C API and Python.

`DPCTLDeviceMgr_GetPositionInDevices` also suffers from the same problem.

PR #630, opened 2021-10-14
Initial stubs for the error handling overhaul.

### LINKED ISSUE #35: Fix the error handling and reporting
The current dpctl error handling broke once we transitioned from the C++ API to the C API. The error reporting and handling in dpctl needs complete redesign.

### LINKED ISSUE #330: Project naming convention
We can see several names in different cases:
- `dpctl` (package name)
- `dpCtl` (documentation)
- `DPCTL` (C++ code)

I propose to make it simple and use `dpctl` in all places.

diptorupd:
Fixes #35 
Does not contain all the changes, more is coming. As of now I only added a new class to capture an "execution state" that for now is just an error handler. The next set of changes will include updating the API of all interfaces.

Fixes #330
Since we are changing the functions anyway, might as well implement the new naming convention proposed in #330.

### DIFF
--- a/dpctl-capi/helper/include/dpctl_error_handlers.h
+++ b/dpctl-capi/helper/include/dpctl_error_handlers.h
@@ -44,3 +44,28 @@ public:
 
     void operator()(const cl::sycl::exception_list &exceptions);
 };
+
+/*!
+ * @brief Provides a static error handling function that prints an error code
+ * and message to the standard error stream.
+ *
+ */
+class DPCTL_API DefaultErrorHandler
+{
+public:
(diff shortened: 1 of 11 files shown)


diptorupd:
@oleksandr-pavlyk I force-pushed again as I cleaned up the interface a bit. Here on, we can have a clean commit log and then chose to squash as desired before merging.

I have modified a single function in the `device_interface` to demonstrate how I want to handle the errors in C++. The Python side is still TBD.

oleksandr-pavlyk:
Subsumed by #683

PR #644, opened 2021-10-29
Fix public CI 

1e-to:
Fix channels priority


oleksandr-pavlyk (commented):
Please squash some commits, so that this PR has 2 commits, first for the ordering of channels, and the second for the upload conditions.

PokhodenkoSA:
> Please squash some commits, so that this PR has 2 commits, first for the ordering of channels, and the second for the upload conditions.

Agree. But better to create 2 separate PRs each with its intent.

oleksandr-pavlyk (approved):

PR #650, opened 2021-11-02
set the last byte in allocated char array to zero.

oleksandr-pavlyk:
Set the last byte of new array of chars to 0 to explicitly ensure null-termination.

Noting that the source of the copy, `str.c_str()` is always null-terminated  (see [c++-reference](https://www.cplusplus.com/reference/string/string/c_str/)), the terminating character of the destination depends on how many bytes the `strncpy_s` actually copies.

Feel free to dismiss.

### DIFF
--- a/dpctl-capi/helper/include/dpctl_string_utils.hpp
+++ b/dpctl-capi/helper/include/dpctl_string_utils.hpp
@@ -50,6 +50,11 @@ cstring_from_string(const std::string &str)
 #else
         std::strncpy(cstr, str.c_str(), cstr_len);
 #endif
+        // Added to resolve CheckMarx's false positive.
+        // NB: This is redundant because str.c_str() is guaranteed
+        // to be null-terminated and the copy function is asked to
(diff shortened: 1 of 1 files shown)


dpctl-capi/helper/include/dpctl_string_utils.hpp
```diff
@@ -50,6 +50,10 @@ cstring_from_string(const std::string &str)
 #else
         std::strncpy(cstr, str.c_str(), cstr_len);
 #endif
+        // This is redundant because str.c_str() is guaranteed to be
+        // null-terminated and copy function is asked to copy that
+        // null-character.
```
diptorupd, line 55: Will be good to add that we still need the line for the specific reason of satisfying Checkmarx. Otherwise it may confuse folks as to why we ever added the line knowingly.

oleksandr-pavlyk: Comment revised.

diptorupd (approved):

PR #655, opened 2021-11-05
Update README.md

samaid:
Intro section enhancements

### DIFF
--- a/dpctl/_sycl_queue_manager.pyx
+++ b/dpctl/_sycl_queue_manager.pyx
@@ -213,7 +213,7 @@ cpdef get_current_backend():
 @contextmanager
 def device_context(arg):
     """
-    Yields a SYCL queue corresponding to the input filter string.
+    Yields a SYCL queue corresponding to the input device object, the queue object, or the filter string.
 
     This context manager "activates", *i.e.*, sets as the currently usable
     queue, the SYCL queue defined by the argument `arg`.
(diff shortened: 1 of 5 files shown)


dpctl/_sycl_queue_manager.pyx
```diff
@@ -213,7 +213,7 @@ cpdef get_current_backend():
 @contextmanager
 def device_context(arg):
     """
-    Yields a SYCL queue corresponding to the input filter string.
+    Yields a SYCL queue corresponding to the input device object, the queue object, or the filter string.
```
oleksandr-pavlyk, line 216:
```suggestion
    Yields a SYCL queue corresponding to the input device object, queue object, or filter string.
```

dpctl/_sycl_queue_manager.pyx
```diff
@@ -223,25 +223,22 @@ def device_context(arg):
     usable queue on exiting the context manager.
 
     Args:
-
-        queue_str (str) : A string corresponding to the DPC++ filter selector.
+        arg : A :class:`.SyclDevice` object, or a :class:`.SyclQueue` object, or a filter selector string
```
oleksandr-pavlyk, line 226:
```suggestion
        arg : A :class:`dpctl.SyclQueue` object, or a :class:`dpctl.SyclDevice` object, or a filter selector string
```

dpctl/_sycl_queue_manager.pyx
```diff
@@ -223,25 +223,22 @@ def device_context(arg):
...
-        filter string.
+        :class:`.SyclQueue`: A SYCL queue corresponding to the specified input device, queue, or filter string.
 
     Raises:
         SyclQueueCreationError: If the SYCL queue creation failed.
 
     :Example:
-        To create a scope within which the Level Zero GPU number 0 is active,
-        a programmer needs to do the following.
+        The following example activates Level Zero GPU number 0 by respective filter selector string:
```
oleksandr-pavlyk, line 235:
```suggestion
        The following example sets current queue targeting specific device indicated with filter selector string in the scope of `with` block:
```

dpctl/_sycl_queue_manager.pyx
oleksandr-pavlyk: Reordered enumeration of argument types: explicit queue instance, explicit device instance, device filter selector string.

dpctl/_sycl_queue_manager.pyx
oleksandr-pavlyk: Note that the whole concept of `device_context` and queue manager is de-facto deprecated and will become obsolete once compute follows data is embraced.

samaid: @oleksandr-pavlyk, yes, the compute follows data will make device_context() almost obsolete. But what about scenario when I created NumPy array with NumPy on the host but want the kernel to be compiled for GPU with Numba, data inherently converted from NumPy array to underlying USM array?

import dpctl
import numpy as np
from numba import njit

@njit
def do_compute(a):
    return np.log(a+1.0)

a = np.arange(10)
with dpctl.device_context(dpctl.select_gpu_device())
    b = do_compute(a)
print(b)

I find it very convenient. I remember multiple discussions about the compute follow data but do not remember we agreed on deprecation of the device context mechanism.

oleksandr-pavlyk:
Split this RP into 5 separate PRs: #659, #660, #661, #662, #663 separate by the filename changed.

This PR can be now closed.

PR #662, opened 2021-11-08
Update intro.rst

oleksandr-pavlyk:
@samaid 

Made Intro section less SYCL centric. Main goal is to avoid references to SYCL as much as possible. Otherwise we will end up teaching Python users DPC++. I understand that references to SYCL in dpctl are unavoidable but still I truly believe we can educate general Python users how to use XPU without the need to read SYCL specs

### DIFF
--- a/docs/docfiles/intro.rst
+++ b/docs/docfiles/intro.rst
@@ -1,17 +1,14 @@
-Welcome to Data-parallel Control (dpctl)'s documentation!
-=========================================================
+Data-Parallel Control - The Library That Controls XPUs
+=======================================================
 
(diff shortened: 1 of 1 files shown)


diptorupd:
@oleksandr-pavlyk refer #670. I have rewritten the intro in that PR addressing some of the concerns raised here.

oleksandr-pavlyk:
This change has been superseded by changes in #712.

diptorupd:
@oleksandr-pavlyk I am fine with closing the PR. But, the main intro section on the landing page was not updated in #712 and remains a TODO.

PR #665, opened 2021-11-09
Replaced uses of clang compiler with icx executable

oleksandr-pavlyk:
Use `icx` executable from oneAPI instead of `clang` executable to avoid clashes with coexisting oneAPI and clang-LLVM installations.


diptorupd (approved):

PokhodenkoSA:
This fix should be addressed to release0.11.

PR #671, opened 2021-11-10
Improvements to documentation generation for dpctl Python API

oleksandr-pavlyk:
This module exposes `generate_class_rst(cls)` which would return
string containing content of rst file, that is currently manually
generated and placed into 'docs/docfiles/dpctl_pyapi/'

Eg.:

```python
import dpctl
import generate_rst as gr

print(gr.generate_class_rst(dpctl.SyclContext))
print(gr.generate_class_rst(dpctl.SyclQueue))
print(gr.generate_class_rst(dpctl.memory.MemoryUSMShared)
```


diptorupd:
@oleksandr-pavlyk One thing that I would like to add is that we should be able to run something like:

```python
import dpctl
import generate_rst as gr

# Automatically finds all classes and generate a rst file for each file and save it to provided path
gr.generate_class_rst(dpctl, path=".")
```

diptorupd:
@oleksandr-pavlyk there are lots of paper cuts, but it is almost there. Things that are pending:

1. The links to in the summary tables for functions, exceptions, enums on the module summary page need to be fixed. 
2. I have added a small check to raise an exception whenever an undocumented object is found that will make doc generation to fail. I am not sure if this is a good idea or not.
3. The `generate_rst` script has to be integrated into the CMake build script.
4. We still need to look at the rendered paged. I have not done that locally.
5. Fix docstrings in source files. This line item should be a separate PR.

PR #674, opened 2021-11-11
Update versions for tools used in pre-commit checks. Add a badge.

diptorupd:
Updated the tools used in our pre-commit hooks. Also added a badge for pre-commit to our README.


oleksandr-pavlyk (approved):

oleksandr-pavlyk:
Let's wait for the documentation to be generated before merging :)

PR #675, opened 2021-11-12
Attempt to fix permission on push issue

oleksandr-pavlyk:
The hope is that using `persist-credential:false` would fix the following
failure to push

remote: Permission to IntelPython/dpctl.git denied to github-actions[bot].


oleksandr-pavlyk:
The error was seen in logs of #674

PR #678, opened 2021-11-19
Extending `dpctl.device_context` with nested contexts

PokhodenkoSA:
Relates to https://github.com/IntelPython/numba-dppy/pull/630.

- [x] Reverse dependencies
  - [x] numba-dppy registers context manager
  - [x] dpctl provides common functionality to register nested contexts
- [x] tests
- [x] docs
- [x] changelog
- [ ] dpctl does not know numba-dppy. Now dpctl helps numba-dppy to register by importing it. Use technique like numba plug-ins.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
 ## [0.11.2] - 11/xx/2021
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.11.2] - 11/xx/2021
 
+### Added
+- Extending `dpctl.device_context` with nested contexts (#678)
+
 ## Fixed
 - Fixed issue #649 about incorrect behavior of `.T` method on sliced arrays (#653)
 
--- a/dpctl/__init__.py
+++ b/dpctl/__init__.py
@@ -56,6 +56,7 @@ from dpctl._sycl_queue import (
 )
 from dpctl._sycl_queue_manager import (
     device_context,
+    nested_context_factories,
     get_current_backend,
     get_current_device_type,
     get_current_queue,
--- a/dpctl/__init__.py
+++ b/dpctl/__init__.py
@@ -56,12 +56,12 @@ from dpctl._sycl_queue import (
 )
 from dpctl._sycl_queue_manager import (
     device_context,
-    nested_context_factories,
     get_current_backend,
     get_current_device_type,
     get_current_queue,
     get_num_activated_queues,
     is_in_device_context,
+    nested_context_factories,
     set_global_queue,
 )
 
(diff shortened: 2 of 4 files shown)


dpctl/_sycl_queue_manager.pyx
```diff
@@ -210,6 +210,14 @@ cpdef get_current_backend():
     return _mgr.get_current_backend()
 
 
+def _get_registered_context_manager(sycl_queue):
+    try:
+        from numba_dppy import get_context_manager
+        return get_context_manager(sycl_queue)
+    except:
```
oleksandr-pavlyk, line 217: I am surprised flake, or other linters did not complain. It should be `except Exception:` or some such.

diptorupd: My local PyLance linter in VSCode points it out.

oleksandr-pavlyk:
```suggestion
    except Exception:
```

oleksandr-pavlyk (dismissed):

dpctl/__init__.py
```diff
@@ -111,6 +112,7 @@
     "get_current_queue",
     "get_num_activated_queues",
     "is_in_device_context",
+    "nested_context_factories",
```
oleksandr-pavlyk, line 115: I do not see why this needs to be exposed, especially since `device_context` is on its way out. 
If it were to stay we should need to design an API to manage these factories. 

I.e. hold then in dictionary and provide functions to retrieve list of registered factories, to add a factory, to remove a factory by keys.

diptorupd: @PokhodenkoSA If we can avoid adding the function here, let us do it. It just adds more clean up for future.

@oleksandr-pavlyk Given that the changes need to make it to 0.12, I am willing to let it slide. Once, `device_context` is gone a whole lot of things will have to be deprecated and removed. Basically, all the functions (`get_current_queue` and family) above `nested_context_factories`.

PokhodenkoSA: > I.e. hold then in a dictionary and provide functions to retrieve the list of keys for the registered factories, to add a factory to this list with some key, to remove a factory by its key.

The factory is a key. So we can use list. Also I think we should not guaranty order of contexts.
Agree about functions.

> belong to `dpctl.utils` namespace, rather than `dpctl`

Agree.

> Once, `device_context` is gone a whole lot of things will have to be deprecated and removed.

Separate discussion needed: How to make numba redirect context active for compute follows data?

> Basically, all the functions (`get_current_queue` and family) above `nested_context_factories`.

Agree. They are located close, so clean up will be easy and in the same time.

CHANGELOG.md
```diff
@@ -4,10 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Next Release] - TBD
+## [Unreleased]
```
PokhodenkoSA, line 7: See [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)

dpctl/__init__.py
PokhodenkoSA: > If we can avoid adding the function here, let us do it.

I think list is simple enough. @oleksandr-pavlyk let me keep things as is. I know that API should be good isolation. But in this case I would like to be simple.

PokhodenkoSA:
I will merge to release0.11.2 branch.

PR #681, opened 2021-11-19
Cleanup/examples

oleksandr-pavlyk:
Updated deprecated SYCL code. 

Changes setup scripts to exclusively use `setuptools`, and not use `numpy.distutils` which have problems with `LLVM` toolchain.


diptorupd:
@oleksandr-pavlyk Please merge if you are satisfied. One thing worth looking at in another PR, should we build and run examples in CI to make sure they do not bit rot?

PR #684, opened 2021-11-23
[Refactoring] Add variables for skip decorators

PokhodenkoSA:
It could be reused in other tests.
Branch based on #678.


oleksandr-pavlyk (commented):
Why not move definitions of these decorators to `_helper.py`

This is where `has_gpu`, etc are defined anyway.

PokhodenkoSA:
Moving skips varibales to common place and reuse is the next step.

PokhodenkoSA:
Also pytest can skip the whole module if some clause not satisfied ([example](https://github.com/IntelPython/numba-dppy/blob/971298a68feb85f14bf0c6bdfbadde33c6f4af5f/numba_dppy/tests/test_debug_dppy_numba.py#L28)):
```python
pytestmark = pytest.mark.skipif(
    not shutil.which("gdb-oneapi"),
    reason="Intel® Distribution for GDB* is not available",
)
```

PokhodenkoSA:
> Why not move

If you agree that this style improves tests, then I will keep this style in my new code.

oleksandr-pavlyk:
Saving decorators as variables makes it more readable, but we are using programmatic way to skip tests based on `try/except` constructs in newer code, e.g.:

```python
try:
   q = dpctl.SyclQueue("cpu")
except dpctl.SyclQueueCreationError:
   pytest.skip("CPU queue could not be constructed")
```

oleksandr-pavlyk (approved):

PokhodenkoSA:
> we are using programmatic way to skip tests based on `try/except` constructs in newer code

This also could be checked in skip decorator.

oleksandr-pavlyk:
> This also could be checked in skip decorator. And this skip decorator could be transformed to variable.

It could, but use of decorator forces one to create a queue for the sake of checking in decorator, throw it away only to recreate it again in the body of the test.

PokhodenkoSA:
> throw it away only to recreate it again in the body of the test

It seems like a preliminary optimization. Why we need it?
We need to skip tests if we can not do some things. Moreover we will be able to do it on the level of module and only once. So it will be faster, just require reordering code in modules.

oleksandr-pavlyk (changes requested):
This PR has become mangled, I am rescinding my approval until this is resolved, or PR is split into smaller chunks.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `dpctl.tensor.asarray`, `dpctl.tensor.empty` implemented (#646).
 
+### Added
+- Extending `dpctl.device_context` with nested contexts (#678)
+
```
PokhodenkoSA, line 14:
```suggestion
```

PokhodenkoSA:
Fixed PR: merge w/ master and target to master.

oleksandr-pavlyk (approved):
Please squash all commits into 1, since all the changes affect single file.

PokhodenkoSA:
> Please squash all commits into 1

You mean force squash and force push to the same branch?

oleksandr-pavlyk:
Or squash when merging, net effect is the same

PokhodenkoSA:
> net effect is the same

I think it is not the same. I will squash to one commit and you apply `Create a merge commit`. It will create 2 commits on target branch.
`squash when merging` will create only 1 commit on target branch.

PR #683, opened 2021-11-22
Implement error_handler function in libsyclinterface/helper 

vlad-perevezentsev:
This PR adds two overloaded `error_handler` functions to pass an exception object and output an error.
Change all functions in `dpctl_sycl_plaftorm_interface.cpp` to use the `error_handler`.

PR #685, opened 2021-11-23
Dpctl c api constructors added, dpctl/include restructured 

oleksandr-pavlyk:
Refactored `dpctl/include` layout. Instead of containing all files from `libsyclinterface/include` it introduces substructure:

```
   dpctl/
      include/
         syclinterface/
              *.h   # all the files that used to be in dpctl/include
         syclinterface.h   
         dpctl_capi.h
         dpctl4pybind11.hpp
```

The newly added files are residing in new folder `dpctl/apis/include` from which `build_backend.py` copies then into `dpctl/include`

Modified `_backend.pxd` to include `"syclinterface/"` in the path. 

Modified `build_backend.py` to reassemble `dpctl/include` from `install/include` and `dpctl/apis/include`.

Added `make_Sycl*` C-API functions to create `SyclDevice`, `SyclQueue`, `SyclContext` and `SyclEvent` from corresponding opaque reference. Used `get_device_ref` and `make_SyclDevice` to construct pybind11 type caster directly mapping `dpctl.SyclDevice` Python object to `sycl::device` in pybind11.

Modified examples files to use new headers, and use casters.

Verified that all Cython extensions and Pybind11 extensions are building and are running as expected.


dpctl/_sycl_context.pyx
```diff
@@ -485,3 +485,12 @@ cdef api DPCTLSyclContextRef get_context_ref(SyclContext ctx):
     :class:`dpctl.SyclContext` instance.
     """
     return ctx.get_context_ref()
+
+
+cdef api SyclContext make_SyclContext(DPCTLSyclContextRef CRef):
```
diptorupd, line 490: Why are these changes here?

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <CL/sycl.hpp>
+#include <pybind11/pybind11.h>
+
+namespace py = pybind11;
+
+namespace pybind11
+{
+namespace detail
+{
+template <> struct type_caster<sycl::queue>
```
diptorupd, line 38: Please add docstrings. So that we can start generating these files in API docs under maybe programmer's guide.

dpctl/_sycl_context.pyx
oleksandr-pavlyk: Because without them pybind11 casters can not be written. One needs 2 way street to go from Python object to c++ class, and back.

dpctl/apis/include/dpctl_capi.h
```diff
@@ -0,0 +1,55 @@
...
+
+/*
+ * Function to import dpctl and make C-API functions available.
+ * C functions can use dpctl's C-API functions without linking to
+ * shared objects defining this symbols, if they call `import_dpctl()`
+ * prior to using those symbols.
+ */
+void import_dpctl(void)
+{
+    import_dpctl___sycl_device();
```
diptorupd, line 49: Oh My! that is some function name :smile:

dpctl/apis/include/syclinterface.h
```diff
@@ -0,0 +1,44 @@
...
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file includes all the headers of syclinterface/
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+// clang-format off
```
diptorupd, line 27: why format off?

dpctl/apis/include/syclinterface.h
```diff
@@ -0,0 +1,44 @@
...
+#include "syclinterface/dpctl_vector.h"
+#include "syclinterface/dpctl_utils.h"
+#include "syclinterface/dpctl_sycl_device_selector_interface.h"
+#include "syclinterface/dpctl_sycl_context_interface.h"
+#include "syclinterface/dpctl_sycl_device_interface.h"
+#include "syclinterface/dpctl_sycl_event_interface.h"
+#include "syclinterface/dpctl_sycl_platform_interface.h"
+#include "syclinterface/dpctl_sycl_queue_interface.h"
+#include "syclinterface/dpctl_sycl_usm_interface.h"
+
```
diptorupd, line 40: spurious new line?

examples/cython/sycl_buffer/use_sycl_buffer.cpp
```diff
@@ -28,7 +28,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "use_sycl_buffer.h"
-#include "dpctl_sycl_types.h"
+#include "syclinterface.h"
```
diptorupd, line 31: should the header be called `dpctl_sycl_interface.h`?

dpctl/apis/include/syclinterface.h
oleksandr-pavlyk: Because clang would reorder includes and include C-API headers before including `dpctl_sycl_types` and then the opaque reference identifiers would be unknown and compilation fails.

examples/cython/sycl_buffer/use_sycl_buffer.cpp
oleksandr-pavlyk: Ok, will make the change.

dpctl/apis/include/syclinterface.h
oleksandr-pavlyk: It was deliberate, to separate `_interface.h` from `_manager.h`, but I am going to remove it.

examples/cython/sycl_buffer/use_sycl_buffer.cpp
oleksandr-pavlyk: done

dpctl/apis/include/syclinterface.h
oleksandr-pavlyk: Pushed the change to remove the line and rename the file.

dpctl/apis/include/dpctl4pybind11.hpp
oleksandr-pavlyk: Done. Added comments to document what's going on.

oleksandr-pavlyk:
@densmirn @Alexander-Makaryev Most likely `dpnp` is going to be affected by this change. Once this PR is merged, the downstream (relative to `dpctl`) libraries will need to ``#include "dpctl_sycl_interface.h"`` instead of including individual headers, such as `dpctl_sycl_types.h` and `dpctl_sycl_queue_manager.h`.

@napetrov Please forward this to relevant people working on `scikit-learn-intelex`.

oleksandr-pavlyk:
@diptorupd This change is going to cause some disruption in downstream code:

1. https://github.com/IntelPython/numba-dppy/blob/main/numba_dppy/dpctl_iface/usm_shared_allocator_ext.c#L30
2. https://github.com/IntelPython/dpnp/blob/master/dpnp/backend/src/queue_sycl.hpp#L44
3. https://github.com/intel/scikit-learn-intelex/blob/master/src/dpctl_interop/daal_context_service.cpp#L20

I am not sure what `DPPLQueueMgr_GetCurrentQueue` is in scikit-learn-intelex, and I did not see it defined anywhere in the project. Perhaps @napetrov could clarify.

But perhaps to minimize the disruption, we could temporarily add symbolic links in `dpctl/include/` for some commonly used header files?

diptorupd:
> But perhaps to minimize the disruption, we could temporarily add symbolic links in `dpctl/include/` for some commonly used header files?

No please. It is a minimal change that we have enough time to fix in downstream. I will suggest that we keep the change out of 0.12 and then fix it.

oleksandr-pavlyk:
> I will suggest that we keep the change out of 0.12 and then fix it.


I am afraid I need parts of this PR (support for dpctl4pybind11) in 0.12.

How about we add a header file (`dpctl/include/dpctl_sycl_queue_manager.h` that contains `#error` with the text explaining what to do.

This header is to be removed before 0.12 tag is created.

Alexander-Makaryev (dismissed):

oleksandr-pavlyk:
> I would therefore merge this to master once this change is added to downstream projects.

Actually, testing locally, there are no conflicts between this branch and the work in #683

PR #691, opened 2021-11-26
[Public CI] Return condition for tag upload package

1e-to:
Added the condition for starting the СI for setting the tag


PokhodenkoSA:
It works but I did not find documentation that supports it.
Moreover when edit in GitHub editor it shows warning:
image

1e-to:
The cause of the problem was different. Another PR will be made

PR #704, opened 2021-12-03
API doc generation automation for dpctl Python API

diptorupd:
The PR incorporates various improvements to dpctl's documentation build system to automate the generation of rst files from the Python modules.

- A new `generate_rst.py` module is added that generates summary and detailed pages for dpctl Python modules and members of those modules.
- Clean ups to the CMake infrastructure to build the rst files using `generate_rst.py`.

PR #707, opened 2021-12-03
Make sure to fail worfklow step in doc generation of make Sphinx errs

oleksandr-pavlyk:
Use `make Sphinx || exit 1`.

See https://stackoverflow.com/a/57907146/594376

Were this change in place, it would have prevented #704 docs generation job from succeeding and erasing existing directory.


.github/workflows/generate-docs.yml
```diff
@@ -90,7 +90,7 @@ jobs:
           echo `pwd`
           cd master
           git rm -rf *
-          mv ~/docs/* .
+          mv ~/docs/* . || exot 1
```
diptorupd, line 93: exit?

oleksandr-pavlyk:
```suggestion
          mv ~/docs/* . || exit 1
```

PR #715, opened 2021-12-07
Adds docstring to event_status_type enum

oleksandr-pavlyk:
Adding docstring.

### DIFF
--- a/dpctl/enum_types.py
+++ b/dpctl/enum_types.py
@@ -33,14 +33,19 @@ class device_type(Enum):
     """
     An enumeration of supported SYCL device types.
 
-    ==================   ============
-    Device type          Enum value
-    ==================   ============
-    gpu                  1
-    cpu                  2
-    accelerator          3
-    host                 4
-    ==================   ============
+    :Example:
+        .. code-block:: python
+
+            import dpctl
+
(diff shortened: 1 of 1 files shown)


dpctl/enum_types.py
```diff
@@ -75,6 +75,19 @@ class backend_type(Enum):
 
 
 class event_status_type(Enum):
+    """
+    An enumeration of SYCL event states.
+
+    ==================   ============
```
diptorupd, line 81: Is there a way to automate generating this table? Do users care about the value of an enum?

diptorupd: My worry is next time we update the enum, the table is not going to be updated and will be incorrect. It has happened to backend and device type.

oleksandr-pavlyk: I think it is fine. The `enum` values has to be added to Python class `event_status_type` anyway, and docstring should be updated while extending the enum.

diptorupd: Ok. Can you fix the docs for the other two enums as well in that case 😄

oleksandr-pavlyk: Done. Removed the table, used that space to give an example.

PR #730, opened 2021-12-23
Update version of clang-format used

oleksandr-pavlyk:
Updated clang-format requested by the workflow from 11 to 12 (13 is the default version used by the action).


oleksandr-pavlyk:
This did not help.

Issue #170, opened 2020-11-06
Add use of nogil to dpctl


oleksandr-pavlyk:
It would be prudent to add `nogil` to appropriate functions in `dpctl`, and use of `with nogil:` Cython context when calling them.

Releasing GIL allows other blocked Python threads to proceed.

oleksandr-pavlyk:
Closing, as resolved with merge of #736

PR #736, opened 2022-01-05
Synchronization functions and USM allocations functions release GIL

oleksandr-pavlyk:
`DPCTLQueue_Wait` and friends are declared nogil, and their use
is moved into `with nogil` context.

This allows us to use `sycl::host_task` to manipulate reference count on a Python object without introducing a deadlock.

Since USM allocation functions are also synchronous and can take some time for 
large allocations, we release GIL for them too. We may perhaps revisit this in the future, only release GIL for large enough allocations.


diptorupd:
Does it address #170?

diptorupd (approved):

oleksandr-pavlyk:
The latest batch of changes closes #170

PR #741, opened 2022-01-07
Add a stub for queues to the user manual.

diptorupd:
Adds a page for queues to the user manual.


docs/docfiles/user_guides/manual/dpctl/queues.rst
```diff
@@ -0,0 +1,19 @@
+.. _queues:
+
+#####
+Queue
+#####
+
+A queue is used to specify a device and a specific set of features of that
```
oleksandr-pavlyk, line 7: I think we should take an entry from basic concepts section verbatim to example what a queue is:

> A queue is needed to schedule execution of any computation, or data copying on the device. Queue construction requires specifying a device and a context targeting that device as well as additional properties, such as whether profiling information should be collected or whether submitted tasks are executed in the order in which they were submitted.

oleksandr-pavlyk:
@diptorupd This is ready for review now. Should we add a section about using `onetrace`?

diptorupd:
> @diptorupd This is ready for review now. Should we add a section about using `onetrace`?

@oleksandr-pavlyk Thanks for filling the text up.  I made some last few edits and this one is good to go. Let us do the `onetrace` section as a follow up. The PR has been pending for a while now, so let us merge it for now.

oleksandr-pavlyk:
To ensure that examples render correctly this PR depends on the later one #827

PR #744, opened 2022-01-12
libsyclinterface/CMakeLists.txt should use CMAKE_CURRENT_SOURCE_DIR

oleksandr-pavlyk:
It should use `CMAKE_CURRENT_SOURCE_DIR` rather than `CMAKE_SOURCE_DIR` to work correctly if included as subdirectory from upper level.


diptorupd (approved):
Good catch

PR #750, opened 2022-01-27
No need to link DPCTLSyclInterface with OpenCL library

oleksandr-pavlyk:
That linkage is redundant as this PR shows.


oleksandr-pavlyk:
Ok, we need it on Windows. Will modify the cmake script to only link on Windows then.

oleksandr-pavlyk:
Going to close this due to conflicts. The right way to fix this is to use OpenCL loader, like we do it for level zero

PR #746, opened 2022-01-14
Feature/build with scikit build and cmake

oleksandr-pavlyk:
Closes #747 
Closes #743 

This is PR transition `dpctl` to use scikit-build project to control building both libsyclinterface and python extensions via CMake. Minimum required Cmake version [is 3.21](https://www.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compiler-setup/using-the-command-line/using-cmake-with-data-parallel-c-dpc.html) to support `find_package(IntelDPCPP REQUIRED)`. 

DPC++ is now being used to build both the library and the Python API compoenents.

To build in develop, use

```sh
python setup.py develop -- -G "Unix Makefiles" -- -DCMAKE_CXX_COMPILER:PATH=icpx  -DDPCTL_ENABLE_LO_PROGRAM_CREATION=ON
```

Alternatively one can use driver scripts:

```sh
python scripts/build_sycl_nightly.py --oneapi
```

Previous syntax of using `--sycl-compiler-prefix` will no longer work.

PR #752, opened 2022-01-28
CYTHON_TRACE must be set when generating coverage

oleksandr-pavlyk:
Setting of this preprocessor variable went missing in the move to
cmake + scikit-build (#746), resulting in coverage not going through Cython
files anymore. This change should restore that, and thus improve the
coverage percentage.


oleksandr-pavlyk:
Black breakage exists in master too and is fixed by #753

PR #754, opened 2022-01-30
Update build_locally.py for Windows support and compilation mode selection

vlad-perevezentsev:
This PR adds Windows OS support for build_locally.py and adds a command line argument `--dubug` to change the compilation mode

### DIFF
--- a/scripts/build_locally.py
+++ b/scripts/build_locally.py
@@ -5,6 +5,7 @@ import sys
 
 def run(
     use_oneapi=True,
+    build_type="Relese",
     c_compiler=None,
     cxx_compiler=None,
     level_zero=True,
@@ -34,7 +35,7 @@ def run(
         "--",
         "-G",
         build_system,
-        "-DCMAKE_BUILD_TYPE=Debug",
+        "-DCMAKE_BUILD_TYPE=" + build_type,
         "-DCMAKE_C_COMPILER:PATH=" + c_compiler,
         "-DCMAKE_CXX_COMPILER:PATH=" + cxx_compiler,
         "-DDPCTL_ENABLE_LO_PROGRAM_CREATION=" + ("ON" if level_zero else "OFF"),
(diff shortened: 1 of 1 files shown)


scripts/build_locally.py
```diff
@@ -70,6 +67,14 @@ def run(
         dest="oneapi",
         action="store_true",
     )
+    driver.add_argument(
```
oleksandr-pavlyk, line 70: Can we use `choices` keyword for `add_argument` here with default to "Release"?

vlad-perevezentsev: Yes, I will add a list of allowable values for `--debug` argument.

scripts/build_locally.py
```diff
@@ -73,6 +73,7 @@ def run(
         nargs="?",
         const="Debug",
         default="Release",
+        choices=["Release", "Debug"],
```
oleksandr-pavlyk, line 76: If the intention was to support `--debug` option with no arguments, the line should be

```python
 driver.add_argument("--debug",
     type=str,
     default="Release",
     const="Debug",
     action="store_const"
  )
```

For example:

```python
In [10]: parser.parse_args([])
Out[10]: Namespace(debug='Release')

In [11]: parser.parse_args(["--debug"])
Out[11]: Namespace(debug='Debug')
```

If the intent is to specify the built type explicitly:

```python
driver.add_argument("--build_type",
    type=str,
    default="Release",
    choices=["Release", "Debug"]
)
```

The usage:

```python
In [15]: parser.parse_args([])
Out[15]: Namespace(debug='Release', build_type='Release')

In [16]: parser.parse_args(["--build_type=Debug"])
Out[16]: Namespace(debug='Release', build_type='Debug')
```

vlad-perevezentsev: Thanks for these examples. I wanted to implement support  `--debug` option without arguments.

oleksandr-pavlyk: Ok, great. Then please make the change. Notice that `narg` parameter is no needed to accomplish that.

oleksandr-pavlyk (dismissed):

scripts/build_locally.py
```diff
@@ -5,25 +5,22 @@
 
 def run(
     use_oneapi=True,
+    build_type="Relese",
```
diptorupd, line 8: Typo

oleksandr-pavlyk, line 8:
```suggestion
    build_type="Release",
```

PR #758, opened 2022-02-02
Cleanup: sycl deprecation warnings, includes

oleksandr-pavlyk:
Clean up for `#includes`, replacing use of `#include "../helper/include/header.h"` with `#include "header.h"`, since include directories already point to `../header/include`.

Replaced deprecated sycl declarations with their intended replacements:

```c++
  sycl::backend::level_zero -> sycl::backend::ext_oneapi_level_zero
  sycl::queue::barrier -> sycl::queue::ext_oneapi_barrier
  sycl::aspect::usm_system_allocator -> sycl::aspect::usm_system_allocations
````

Also removed uses of reprecated `sycl::runtime_error` in favor of use of `sycl::exception`.

Because of deprecation of `sycl::aspect::usm_system_allocator`, changed the name of dpctl enum entry name accordingly, which is a user visible change.

### DIFF
--- a/libsyclinterface/source/dpctl_sycl_device_interface.cpp
+++ b/libsyclinterface/source/dpctl_sycl_device_interface.cpp
@@ -25,11 +25,11 @@
 //===----------------------------------------------------------------------===//
 
 #include "dpctl_sycl_device_interface.h"
-#include "../helper/include/dpctl_error_handlers.h"
-#include "../helper/include/dpctl_string_utils.hpp"
-#include "../helper/include/dpctl_utils_helper.h"
 #include "Support/CBindingWrapping.h"
+#include "dpctl_error_handlers.h"
+#include "dpctl_string_utils.hpp"
 #include "dpctl_sycl_device_manager.h"
+#include "dpctl_utils_helper.h"
 #include <CL/sycl.hpp> /* SYCL headers   */
 #include <algorithm>
 #include <cstring>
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

**Comment by oleksandr-pavlyk** [comment]
Aspects support was also expanded in 2022.0 DPC++ to allow us to address some of #309, to be addressed in a separate PR.

File: libsyclinterface/source/dpctl_sycl_device_interface.cpp [hunk 1]
```diff
@@ -25,11 +25,11 @@
...
 #include "dpctl_sycl_device_interface.h"
-#include "../helper/include/dpctl_error_handlers.h"
-#include "../helper/include/dpctl_string_utils.hpp"
-#include "../helper/include/dpctl_utils_helper.h"
 #include "Support/CBindingWrapping.h"
+#include "dpctl_error_handlers.h"
+#include "dpctl_string_utils.hpp"
 #include "dpctl_sycl_device_manager.h"
+#include "dpctl_utils_helper.h"
 #include <CL/sycl.hpp> /* SYCL headers   */
```
diptorupd, line 33: How about switching to `sycl.hpp` too?

diptorupd (approved):
Thanks!

libsyclinterface/source/dpctl_sycl_device_interface.cpp
oleksandr-pavlyk: Tried both `<sycl/sycl.hpp>` and `<sycl.hpp>`. Both errored out. Going to merge as is.

Issue #651, opened 2021-11-02
dpctl should not vendor versioneer, but use one from PyPI or conda-forge


oleksandr-pavlyk:
`dpctl` uses `versioneer.py` to produce `_version.py` included in the installed package. 

`dpctl` need not vendor the `versioneer.py` in its repo. It can get it either from [PyPI](https://pypi.org/project/versioneer/), or from [conda-forge](https://anaconda.org/conda-forge/versioneer).

Unfortunately, `versioneer` is not available on pkgs/main.

oleksandr-pavlyk:
The versioneer's use instruction states:

1. Install versioneer from pip
2. Configure `setup.cfg`
3. Run `versioneer install` which copies some files in the project's layout to be vendored. 

In view of this, closing this ticket.

PR #764, opened 2022-02-03
Update vendored versioneer from 0.18 to 0.21

oleksandr-pavlyk:
Following instructions from [versioneer's install page](https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md) updated the vendored version of versioneer.py from 0.18 to 0.21

The instruction actually explicitly calls for vendoring, so I suppose #651 should be just closed.


dpctl/_version.py
```diff
@@ -41,8 +43,8 @@ def get_config():
     cfg.VCS = "git"
     cfg.style = ""
     cfg.tag_prefix = ""
-    cfg.parentdir_prefix = "DP-Glue-"
-    cfg.versionfile_source = "dpglue/_version.py"
+    cfg.parentdir_prefix = "None"
+    cfg.versionfile_source = "dpctl/_version.py"
```
diptorupd, line 47: Oh my! we still had `dpglue` in our code.

diptorupd:
Let's merge if you are happy with the patch.

PR #770, opened 2022-02-07
Fixed type in utility `_memory_overlap`.

oleksandr-pavlyk:
To intervals `[B1..E1)` and `[B2..E2)` do not overlap when
`E1 <= B2 || E2 <= B1` <=> `B1 >= E2 || B2 >= E1`.

The typo in the code was the reason why the coverage was jumping
around.

The overlap may occur only if `(B1 < E2) and (B2 < E1)`.


oleksandr-pavlyk:
I think one can do better, `[B1..E1)` and `[B2..E2)` do not overlap when `E1 <= B2` or `E2 <= B1`.

oleksandr-pavlyk:
Failure in the internal is CI is unrelated (CMake configure failure while testing linker). Merging.

PR #776, opened 2022-02-09
Only call PyThread_Ensure from host_task if the main-thread interpreter is initialized

oleksandr-pavlyk:
This PR applies the solution worked out in a toy POC

Host tasks are executed in a separate threads from
the main thread which submitted the kernel and incremented
reference counts.

When kernel submission occurs near the end of the script,
the CPython interpreter may have begun finalization process
by the time the kernel completes the execution.

`PyThread_Ensure` would crash in that case. So execute the host
task only if `Py_IsInitialized()`.

The only testing of this scenario is to submit the kernel at
the end and not call queue synchronization.


oleksandr-pavlyk:
Internal CI failure is not related to changes in this PR:  failed to start the job due to a setup issue.

oleksandr-pavlyk:
@PokhodenkoSA @diptorupd Ping

PR #778, opened 2022-02-12
Fix for issue 777, restoring versioned library symbolic links in installed package on Linux

### LINKED ISSUE #777: Creating links for `libDPCTLSyclInterface.so` in building via `python setup.py install`
When build via `python setup.py develop` it creates links for `libDPCTLSyclInterface.so`:
```
lrwxrwxrwx  1 spokhode intelall     26 Feb 11 14:59 libDPCTLSyclInterface.so -> libDPCTLSyclInterface.so.0
lrwxrwxrwx  1 spokhode intelall     29 Feb 11 14:50 libDPCTLSyclInterface.so.0 -> libDPCTLSyclInterface.so.0.12
-rw-r--r--  1 spokhode intelall 231432 Feb 11 14:50 libDPCTLSyclInterface.so.0.12
```
When build via `python setup.py install` it creates copies of `libDPCTLSyclInterface.so`:
```
-rwxr-xr-x  1 spokhode intelall 231432 Feb 11 16:26 libDPCTLSyclInterface.so
-rw-r--r--  1 spokhode intelall 231432 Feb 11 16:26 libDPCTLSyclInterface.so.0
-rw-r--r--  1 spokhode intelall 231432 Feb 11 16:26 libDPCTLSyclInterface.so.0.12
```

oleksandr-pavlyk:
Closes #777

When scikit-build assembles dpctl, it runs cmake with `_skbuild/*/cmake-build` being
as its build directory, `_skbuild/*/cmake-install` being its install directory. It then
runs setuptools.setup to execute `build_py` steps using `_skbuild/*/setuptools` as its
`build_base`. Since sycl interface library is registered in `package_data`, it is copied
from `cmake-install` folder (where symbolic links are intact) to `setuptools` (where 
unpatched `setuptools.command.build_py` command follows them creating 
hard copies).

Then, when running `install_lib` step, files from `_skbuild/*/setuptools/lib/dpctl` are
copied into `site-packages/dpctl` using `copy_tree` function, which also follows 
symbolic links this creating hard copies).

This PR transfers logic from pre-scikit-build setup.py to fix hard links as post
`setuptools.command.install.run()` step).

A test is added to `tests/test_service.py` to verify that on Linux some of library files
are symbolic links.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,25 +14,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import glob
 import os.path
 import pathlib
 import shutil
+import sys
 
 import skbuild
 import skbuild.setuptools_wrap
 import skbuild.utils
 from setuptools import find_packages
+from skbuild.command.build_py import build_py as _skbuild_build_py
+from skbuild.command.install import install as _skbuild_install
 
 import versioneer
 
 # Get long description
 with open("README.md", "r", encoding="utf-8") as file:
     long_description = file.read()
-
-
-def _get_cmdclass():
-    cmdclass = versioneer.get_cmdclass()
-    return cmdclass
 
 
 def cleanup_destination(cmake_manifest):
@@ -52,7 +51,9 @@ def cleanup_destination(cmake_manifest):
     return cmake_manifest
 
 
-def _patched_copy_file(src_file, dest_file, hide_listing=True):
+def _patched_copy_file(
+    src_file, dest_file, hide_listing=True, preserve_mode=True
+):
     """Copy ``src_file`` to ``dest_file`` ensuring parent directory exists.
 
     By default, message like `creating directory /path/to/package` and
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -79,6 +80,40 @@ def _patched_copy_file(src_file, dest_file, hide_listing=True):
 skbuild.setuptools_wrap._copy_file = _patched_copy_file
 
 
+class BuildPyCmd(_skbuild_build_py):
+    def copy_file(self, src, dst, preserve_mode=True):
+        _patched_copy_file(src, dst, preserve_mode=preserve_mode)
+        return (dst, 1)
+
+
+class InstallCmd(_skbuild_install):
```
PokhodenkoSA, line 89: Will it work for `python setup.py develop` too? or only for `python setup.py intell` ?

oleksandr-pavlyk: It works with both, and since the test was added to the test suite, CI is checking it for install in conda-packages workflow, and for decelop in generate-coverage and build-with-os-llvm workflows.

diptorupd (approved):

PR #781, opened 2022-02-17
[FEATURE] Pybind11-based extension for usm_ndarray copying-and-casting

oleksandr-pavlyk:
This PR adds a pybind11 extension `dpctl.tensor._tensor_impl` to the project, built from `dpctl/tensor/libtensor/source/tensor_py.cpp`.

It contains an implementation of `_tensor_impl._copy_usm_ndarray_into_usm_ndarray(src, dst, queue=None, depends=[])` self-descriptive utility function. 

`dpctl/tensor/libtensor/tests/test_copy.py` contains the test file that checked correctness of the implementation.

This PR is for **discussion** only. 

- [x] Git history would need to be revised
- [x] Some files need to be deleted
- [x] The implemented copy & cast utility deployed in `dpctl.tensor` functions.
- [ ] The copy & cast function need to add logic to deal with the case when memory underlying X and Y overlap, by copying via a temporary.
- [x] `host_task` calls need to be added to ensure that `src` and `dst` are not garbage collected while underlying kernels are executing.


oleksandr-pavlyk:
The crash during test run in internal CI on Windows Py 3.8 can not be reproduced manually running exactly the same `conda build --test` command on that very same machine. It is also not reproducible in the public CI, not is it reproducible with Py 3.7 or Py 3.9 on Windows. 

I therefore discount this crash error as caused by some host misconfiguration. @AndresGuzman-Ballen

AndresGuzman-Ballen:
@diptorupd do you think this is ready to merge? Sasha is out so I don't know whether he would want this merged now or when he gets back

oleksandr-pavlyk:
@AndresGuzman-Ballen Thank you for the fix. How did you come to realize this is the fault of pytest's faulthandler plugin?

oleksandr-pavlyk:
Answering your question, the PR is still awaiting review. I intend to merge it Sunday evening, and address any feedback in a separate PRs.

PR #788, opened 2022-03-09
Implementation of expand_dims function

vlad-perevezentsev:
This PR adds `expand_dims` functions according to [Python array API standard](https://data-apis.org/array-api/latest/API_specification/generated/signatures.manipulation_functions.expand_dims.html) for usm_ndarray.

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -81,6 +81,6 @@ def expand_dims(X, axes):
     axes = normalize_axis_tuple(axes, out_ndim)
 
     shape_it = iter(X.shape)
-    shape = [1 if ax in axes else next(shape_it) for ax in range(out_ndim)]
+    shape = tuple(1 if ax in axes else next(shape_it) for ax in range(out_ndim))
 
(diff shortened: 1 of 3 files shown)


dpctl/tensor/_manipulation_functions.py
```diff
@@ -61,3 +62,25 @@ def permute_dims(X, axes):
...
+    if not isinstance(X, dpt.usm_ndarray):
+        raise TypeError(f"Expected usm_ndarray type, got {type(X)}.")
+    if not isinstance(axes, (tuple, list)):
+        axes = (axes,)
+
+    out_ndim = len(axes) + X.ndim
+    axes = normalize_axis_tuple(axes, out_ndim)
+
+    shape_it = iter(X.shape)
+    shape = [1 if ax in axes else next(shape_it) for ax in range(out_ndim)]
```
oleksandr-pavlyk, line 84:
```suggestion
    shape = tuple(1 if ax in axes else next(shape_it) for ax in range(out_ndim))
```

oleksandr-pavlyk (approved):

PR #791, opened 2022-03-16
Implementation of broadcast_to function

vlad-perevezentsev:
This PR adds `broadcast_to` functions according to [Python array API standard](https://data-apis.org/array-api/latest/API_specification/generated/signatures.data_type_functions.broadcast_to.html) for usm_ndarray.

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -15,9 +15,27 @@
 #  limitations under the License.
 
 
+import numpy as np
 from numpy.core.numeric import normalize_axis_tuple
 
 import dpctl.tensor as dpt
+
+
+def _broadcast_strides(X_shape, X_strides, res_ndim):
+    """
+    Broadcasts strides to match the given dimensions;
+    returns tuple type strides.
+    """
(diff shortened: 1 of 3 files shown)


dpctl/tensor/_manipulation_functions.py
```diff
@@ -102,3 +120,26 @@ def squeeze(X, axes=None):
...
+    """
+    broadcast_to(X: usm_ndarray, shape: tuple or list) -> usm_ndarray
+
+    Broadcast an array to a new shape; returns the broadcasted
+    array as a view.
+    """
+    if not isinstance(X, dpt.usm_ndarray):
+        raise TypeError(f"Expected usm_ndarray type, got {type(X)}.")
+
+    new_array = np.broadcast_to(
```
oleksandr-pavlyk, line 135: It may seem that calls to `np.broadcast_to` are not needed. `new_array.ndim` is just `len(shape)` and `new_array.shape` is just normalized `shape`. However, the call also validates input parameter `shape`. Perhaps this warrants a comment.

oleksandr-pavlyk (approved):

PR #797, opened 2022-03-22
Get coerced usm type

oleksandr-pavlyk:
Adds `dpctl.utils.get_coerced_usm_type` to deduce the usm type for the output array based on usm types of input arrays.

Also modified behavior of `dpctl.utils.get_execution_queue`. The notion of equivalency has changed. Now two queues are equivalent if they compare equal, i.e. `q1 == q2`. This is true when the underlying C++ class instances are copies of the same underlying instance.

Note that `dpctl.SyclQueue() == dpctl.SyclQueue()` returns `False`, and so would `dpctl.SyclQueue("cpu") == dpctl.SyclQueue("cpu")`.

To bring two arrays to equivalent queues, do `Yc = Y.to_device(X.device)`, and then `foo(X, Yc)` if `foo(X, Y)` was raising due to incompatible queues.

### DIFF
--- a/dpctl/tensor/_device.py
+++ b/dpctl/tensor/_device.py
@@ -29,6 +29,8 @@ class Device:
     or ``sycl_device``.
     """
 
+    __device_queue_map__ = dict()
+
     def __new__(cls, *args, **kwargs):
         raise TypeError("No public constructor")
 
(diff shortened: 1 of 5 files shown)


oleksandr-pavlyk:
A forthcoming PR would implement caching for the mapping from filter strings and unpartitioned devices to queues.

oleksandr-pavlyk:
@PokhodenkoSA @Alexander-Makaryev @diptorupd The change to `dpctl.utils.get_execution_queue` is backward incompatible and may require changes in `dpnp` and `numba-dpex` tests. Implementation of aforementioned caching should ease the burden.

I would therefore not merge this until such caching is implemented.

oleksandr-pavlyk:
> 
> I would therefore not merge this until such caching is implemented.

The caching has been implemented now:

```

In [3]: X = dpt.empty((10,))

In [4]: Y = dpt.empty((5,2), 'd', usm_type="shared")

In [5]: X.sycl_queue is Y.sycl_queue
Out[5]: True

In [6]: X = dpt.empty((10,), device="cpu")

In [7]: Y = dpt.empty((5,2), 'd', usm_type="shared", device="opencl:cpu:0")

In [8]: X.sycl_queue is Y.sycl_queue
Out[8]: True
```

PR #801, opened 2022-03-29
Implementation of flip function 

vlad-perevezentsev:
This PR adds `flip` functions according to [Python array API standard](https://data-apis.org/array-api/latest/API_specification/generated/signatures.manipulation_functions.flip.html) for usm_ndarray.

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -207,3 +207,24 @@ def broadcast_arrays(*args):
         return args
 
     return [broadcast_to(X, shape) for X in args]
+
+
+def flip(X, axes=None):
+    """
+    flip(X: usm_ndarray, axes: int or tuple or list) -> usm_ndarray
+
+    Reverses the order of elements in an array along the given axis.
+    The shape of the array is preserved, but the elements are reordered;
+    returns a view of X with the entries of axis reversed.
+    """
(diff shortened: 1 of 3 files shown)


dpctl/tensor/_manipulation_functions.py
```diff
@@ -207,3 +207,25 @@ def broadcast_arrays(*args):
...
+    returns a view of X with the entries of axis reversed.
+    """
+    if not isinstance(X, dpt.usm_ndarray):
+        raise TypeError(f"Expected usm_ndarray type, got {type(X)}.")
+    X_ndim = X.ndim
+    if axes is None:
+        indexer = (np.s_[::-1],) * X_ndim
+    else:
+        axes = normalize_axis_tuple(axes, X_ndim)
+        indexer = [np.s_[:]] * X.ndim
```
oleksandr-pavlyk, line 227: More direct `indexer = tuple( np.s_[::-1] if i in axes else np.s_[:] for i in range(X.ndim))`.

This way a list temporary is not needed.

vlad-perevezentsev: Great remark. Thanks

vlad-perevezentsev:
Tests were added for an empty array with 0-size dimensions
`test_flip_0d` was updated

oleksandr-pavlyk (approved):
LGTM! Thanks @vlad-perevezentsev

PR #805, opened 2022-03-30
Add .mT, change .T properties of dpctl.tensor.usm_ndarray 

oleksandr-pavlyk:
This PR modifies `usm_ndarray.T` property behavior per array API spec, and introduces `usm_ndarray.mT` property.

Tests added.


vlad-perevezentsev (approved):
LGTM

PR #818, opened 2022-04-20
Improve coverage

oleksandr-pavlyk:
This PR changes generation of coverage of libDPCTLSyclInterface, building ctest executable and linking it against the library, as opposed to recompiling all the library together with test files into one executable. 

Thanks to @diptorupd insight on how to generate coverage for code from shared object, it becomes possible to measure coverage of the DPCTLSyclInterface library generated by calls from Python API, specifically exercised by pytest.


diptorupd:
@oleksandr-pavlyk do you see any cases for `libsyclinterface` where the profiling of the cython tests cases leads to a previously uncovered code line to be now covered?

diptorupd (approved):
The plan worked! All the other CMake clean ups are good. Most exciting is that we could boost libsyclinterface coverage using the Python package tests.

Issue #749, opened 2022-01-19
Add an example showing the creation of a toy PyBind11 extension for oneMKL using dpctl


diptorupd:
[oneMKL](https://github.com/oneapi-src/oneMKL) is an open-source implementation of the oneMKL Data Parallel C++ (DPC++) interface according to the oneMKL specification. The library provides backends for CUDA as well as Intel MKL. In addition, it is possible to dynamically select a backend based on a SYCL queue.

Our example should demonstrate how dpctl's infrastructure can be used to easily create a Python module for oneMKL.

A high-level design can we as follows:

- [ ] Expose the `onemkl::gemm` routine as `matmul`.
- [ ] Introduce convenience types `mat` and `vec` that are based on `dpctl.tensor.usm_ndarray`.
- [ ] The function should accept a `SyclQueue`, `mat`, and optionally a list of `SyclEvent` objects and return a `SyclEvent` object.
- [ ] Demonstrate the usage of `SyclTimer` to time execution.
- [ ] Demonstrate ways to print a matrix/vector by first copying them into `Numpy.ndarray`.
- [ ] Demonstrate ways to populate a matrix/vector.

diptorupd:
Has been added a while back.

PR #821, opened 2022-04-24
onemkl_gemv example expanded: iterative linear solver example working

oleksandr-pavlyk:
This is the example with Chebyshev solver and conjugate gradient descent solver based on oneMKL gemv call.


oleksandr-pavlyk:
I think this PR completed #749

PR #827, opened 2022-05-02
Add default context

oleksandr-pavlyk:
Adds `DPCTLPlatform_GetDefaultContext` that exposes ``sycl::platform::ext_oneapi_get_default_context()`` ([ext_oneapi_default_context](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_default_context.asciidoc) extension).

### DIFF
--- a/dpctl/_sycl_device.pyx
+++ b/dpctl/_sycl_device.pyx
@@ -644,7 +644,7 @@ cdef class SyclDevice(_SyclDevice):
         )
 
     @property
-    def platform(self):
+    def sycl_platform(self):
         """ Returns the platform associated with this device.
 
         Returns:
--- a/dpctl/tests/test_sycl_device.py
+++ b/dpctl/tests/test_sycl_device.py
@@ -497,7 +497,7 @@ def check_profiling_timer_resolution(device):
 
 
 def check_platform(device):
-    p = device.platform
+    p = device.sycl_platform
     assert isinstance(p, dpctl.SyclPlatform)
 
 


oleksandr-pavlyk:
@diptorupd Do you think `SyclDevice.platform` should be `SyclDevice.sycl_platform` instead?

dpctl/_sycl_device.pyx
```diff
@@ -639,6 +643,22 @@ cdef class SyclDevice(_SyclDevice):
...
+    def platform(self):
+        """ Returns the platform associated with this device.
+
+        Returns:
+            :class:`dpctl.SyclPlatform`: The platform associated with this
+                device.
+        """
+        cdef DPCTLSyclPlatformRef PRef = (
+            DPCTLDevice_GetPlatform(self._device_ref)
+            )
```
diptorupd, line 656: indent is off. Should line up with the `cdef`

dpctl/_sycl_device.pyx
```diff
@@ -639,6 +643,22 @@ cdef class SyclDevice(_SyclDevice):
...
+        if (PRef == NULL):
+            raise RuntimeError("Could not get platform for device.")
+        else:
+            return SyclPlatform._create(PRef)
```
diptorupd, line 660: I am assuming `_create` will consume the `PRef`?

oleksandr-pavlyk: Yes, it does.

dpctl/_sycl_device.pyx
oleksandr-pavlyk: Fixed

dpctl/tests/test_sycl_platform.py
```diff
@@ -87,6 +87,11 @@ def check_repr(platform):
     assert r != ""
 
 
+def check_default_context(platform):
+    r = platform.default_context
+    assert type(r) is dpctl.SyclContext
```
diptorupd, line 92: Do we ever expect `r` to be None?

diptorupd: Guess not, but can run into an Exception. The test case can handle that?

oleksandr-pavlyk: Yes, an exception would get raised and pytest will report it

libsyclinterface/source/dpctl_sycl_platform_interface.cpp
```diff
@@ -202,3 +203,18 @@ __dpctl_give DPCTLPlatformVectorRef DPCTLPlatform_GetPlatforms()
...
+__dpctl_give DPCTLSyclContextRef
+DPCTLPlatform_GetDefaultContext(__dpctl_keep const DPCTLSyclPlatformRef PRef)
+{
+    auto P = unwrap(PRef);
+    if (P) {
+        auto default_ctx = P->ext_oneapi_get_default_context();
+        return wrap(new context(default_ctx));
+    }
+    else {
+        error_handler("Driver version cannot be looked up for a NULL platform.",
```
diptorupd, line 216: Hmm what is this error message? Copy pasta?

oleksandr-pavlyk: Right. Sorry.

libsyclinterface/tests/test_sycl_queue_interface.cpp
```diff
@@ -446,7 +446,6 @@ TEST_P(TestDPCTLQueueMemberFunctions, CheckMemset)
 
     ASSERT_NO_FATAL_FAILURE(DPCTLfree_with_queue(p, QRef));
 
-    bool equal = true;
```
diptorupd, line 449: unrelated?

dpctl/tests/test_sycl_device.py
```diff
@@ -552,6 +557,8 @@ def check_profiling_timer_resolution(device):
     check_repr,
     check_get_global_mem_size,
     check_get_local_mem_size,
+    check_profiling_timer_resolution,
```
diptorupd, line 560: unrelated?

libsyclinterface/tests/test_sycl_queue_interface.cpp
oleksandr-pavlyk: Yes, and in a separate commit.

dpctl/tests/test_sycl_device.py
oleksandr-pavlyk: Yes, I realized this check added earlier was not exercised, so added it along with the `check_platform`

diptorupd (approved):

PR #832, opened 2022-05-11
Improves the logic to implement IntelSycl version requirement.

diptorupd:
The logic to compare package versions inside the CMake module `FindIntelSycl.cmake` is incorrect. We are using lexicographic string comparison that can lead to an version check of `1.7.9 > 1.7.15` evaluating to `True`.  The PR fixes the issues by introducing a correct way to compare versions.

### DIFF
--- a/libsyclinterface/cmake/modules/FindIntelSycl.cmake
+++ b/libsyclinterface/cmake/modules/FindIntelSycl.cmake
@@ -39,6 +39,48 @@ find_package(IntelDPCPP REQUIRED)
 set(clangxx_cmd "${CMAKE_CXX_COMPILER}")
 set(clangxx_arg "--version")
 
+# Compares the two version string that are supposed to be in x.y.z format
+# and reports if the argument VERSION_STR1 is greater than or equal than
+# version_str2. The strings are compared lexicographically after conversion to
+# lists of equal lengths, with the shorter string getting zero-padded.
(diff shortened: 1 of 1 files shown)


libsyclinterface/cmake/modules/FindIntelSycl.cmake
```diff
@@ -39,6 +39,48 @@ find_package(IntelDPCPP REQUIRED)
 set(clangxx_cmd "${CMAKE_CXX_COMPILER}")
 set(clangxx_arg "--version")
 
+# Compares the two version string that are supposed to be in x.y.z format
+# and reports if the argument VERSION_STR1 is greater than or equal than
+# version_str2. The strings are compared lexicographically after conversion to
+# lists of equal lengths, with the shorter string getting zero-padded.
+function(compare_versions VERSION_STR1 VERSION_STR2 OUTPUT)
```
oleksandr-pavlyk, line 46: Perhaps a more suitable name is `versions_greater_equal`.   I would be expecting `compare_versions` to return -1, 0, or 1, like `strcmp` does.

oleksandr-pavlyk (dismissed):
LGTM, but would suggest to rename the function.

libsyclinterface/cmake/modules/FindIntelSycl.cmake
diptorupd: Done.

oleksandr-pavlyk (approved):

PR #837, opened 2022-05-16
Improve type casters

oleksandr-pavlyk:
Improved efficiency of pybind11 classes and type casters. 

Type casters in pybind11 generated with `PYBIND11_TYPE_CASTER` macro rely on default constructor of C++ type. The default constructed value is then overwritten by `load`  method. C++ types such as `sycl::device` and `sycl::queue` do non-trivial amount of work in default constructor. This work was ranking high in certain workloads (such as `example/pybind11/onemkl_gemv/sycl_timing_solver.py`).

Same applies to auto-generated type-casters for classes `dpctl::memory::usm_memory` and `dpctl::tensor::usm_ndarray`.

This PR introduces DPCTL_TYPE_CASTER macro that defines `unique_ptr` to the C++ type, rather than the type itself. This avoids superfluous call to default constructor altogether.

Instead of redefining `pyobject_caster` automatically generated for `dpctl::tensor::usm_ndarray` and `dpctl::memory::usm_memory` a singleton class `dpctl::detail::dpctl_api` is added that creates objects used by default constructors of these classes.

### DIFF
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -39,6 +39,45 @@ namespace pybind11
 namespace detail
 {
 
+#define DPCTL_TYPE_CASTER(type, py_name)                                       \
+protected:                                                                     \
+    std::unique_ptr<type> value;                                               \
+                                                                               \
+public:                                                                        \
+    static constexpr auto name = py_name;                                      \
+    template <                                                                 \
+        typename T_,                                                           \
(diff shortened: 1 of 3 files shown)


oleksandr-pavlyk:
On my development Gen9 box this change speeds up the test suite run-time from 85 seconds to 54 seconds.

Now:

```
======================== 3107 passed, 173 skipped, 1 xfailed in 53.81s ===============================
```

Before:

```
====================== 3107 passed, 173 skipped, 1 xfailed in 84.72s (0:01:24) =======================
```

PR #840, opened 2022-05-18
Excised use of DPCTL_COVERAGE preprocessor variable

oleksandr-pavlyk:
Enabled code previously excluded due to use of `DPCTL_COVERAGE` variable to work around issues in compiler which have now been addressed.


oleksandr-pavlyk:
This PR activates a number of tests previously excluded to work around issues in the compiler toolchain.

PR #853, opened 2022-06-23
Install cmake scripts

oleksandr-pavlyk:
This PR fixes an issue where `.cmake` scripts needed for CMake to find `dpctl` were not installed into `site-packages/dpctl`. It can now be found in `site-packages/dpctl/resources/cmake`.  

- `CMakeLists.txt` was modified to copy `*.cmake` files from `cmake/` to `dpctl/resources/cmake`.
- `MANIFEST.in` file was modified to ensure `*.cmake` files are copied over.
- `__main__.py` was added to allow scripts to get
   - `python -m dpctl --includes` - prints include flags for compiling source using dpctl's include files
   - `python -m dpctl --library` - prints linker flags that needs to link to SyclInterface library
   - `python -m dpctl --cmakdir` - prints the folder where cmake scripts are which you set `DPCTL_ROOT` to for cmake to be able to execute `find_packages(dpctl)`.
   - Added tests to check that `__main__` works correctly
 

--- 

Also modified vendored `IntelDPCPPConfig.cmake` to align it with that included in oneAPI DPC++ compiler layout.


oleksandr-pavlyk:
With this change building a Python extension using scikit-build with open-source DPC++ from nightly sycl bundle becomes possible by doing 

```bash
IntelDPCPP_DIR=$(python -m dpctl --cmakedir) CXX=$(which clang++) python setup.py develop
```

PR #867, opened 2022-07-26
Added dpctl.tensor.concat feature and tests

npolina4:
concat() function concatenates several arrays along one of axis and follows [array API spec](https://data-apis.org/array-api/latest/API_specification/generated/signatures.manipulation_functions.concat.html).

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -298,7 +298,7 @@ def concat(arrays, axis=0):
     if n == 0:
         raise TypeError("Missing 1 required positional argument: 'arrays'")
 
-    if not isinstance(arrays, list) and not isinstance(arrays, tuple):
+    if not isinstance(arrays, (list, tuple)):
         raise TypeError(f"Expected tuple or list type, got {type(arrays)}.")
 
     for X in arrays:
@@ -314,8 +314,12 @@ def concat(arrays, axis=0):
         raise ValueError("All the input arrays must have usm_type")
 
     X0 = arrays[0]
-    if any(X0.dtype != arrays[i].dtype for i in range(1, n)):
-        raise ValueError("All the input arrays must have same dtype")
+    if not all(Xi.dtype.char in "?bBhHiIlLqQefdFD" for Xi in arrays):
+        raise ValueError("Unsupported dtype encountered.")
+
+    res_dtype = X0.dtype
+    for i in range(1, n):
+        res_dtype = np.promote_types(res_dtype, arrays[i])
 
     for i in range(1, n):
         if X0.ndim != arrays[i].ndim:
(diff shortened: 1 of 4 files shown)


dpctl/tensor/_manipulation_functions.py
```diff
@@ -285,3 +286,86 @@ def roll(X, shift, axes=None):
...
+    """
+    concat(arrays: tuple or list of usm_ndarrays, axis: int) -> usm_ndarray
+
+    Joins a sequence of arrays along an existing axis.
+    """
+    n = len(arrays)
+    if n == 0:
+        raise TypeError("Missing 1 required positional argument: 'arrays'")
+
+    if not isinstance(arrays, list) and not isinstance(arrays, tuple):
```
oleksandr-pavlyk, line 301:
```suggestion
    if not isinstance(arrays, (list, tuple)):
```

dpctl/tensor/_manipulation_functions.py
```diff
@@ -285,3 +286,86 @@ def roll(X, shift, axes=None):
...
+    exec_q = dputils.get_execution_queue([X.sycl_queue for X in arrays])
+    if exec_q is None:
+        raise ValueError("All the input arrays must have same sycl queue")
+
+    res_usm_type = dputils.get_coerced_usm_type([X.usm_type for X in arrays])
+    if res_usm_type is None:
+        raise ValueError("All the input arrays must have usm_type")
+
+    X0 = arrays[0]
+    if any(X0.dtype != arrays[i].dtype for i in range(1, n)):
```
oleksandr-pavlyk, line 317: It was my mistake, I misread the spec, but the description of the output [states](https://data-apis.org/array-api/latest/API_specification/generated/signatures.manipulation_functions.concat.html) 

> If the input arrays have different data types, normal [Type Promotion Rules](https://data-apis.org/array-api/latest/API_specification/type_promotion.html#type-promotion) must apply. If the input arrays have the same data type, the output array must have the same data type as the input arrays.

Thus we should compute data-type of the output array using `numpy.promote_types` which will later be replaced with `dpt.promote_types`. 

The code is also lacking verification that dtypes of input arrays are valid. Such a check for `X0` would be 

```python
X0.dtype.kind in "?bBhHiIlLefdFD"
```

the white-list string was obtained from

```ipython
In [12]: ''.join([np.dtype(t).char for t in [np.bool_, np.int8, np.uint8, np.int16, np.uint16, 
    ...: np.int32, np.uint32, np.int64, np.uint64, np.float16, np.float32, np.float64,
    ...: np.complex64, np.complex128]])
Out[12]: '?bBhHiIlLefdFD'
```

oleksandr-pavlyk (approved):
LGTM! Thank you @npolina4

PR #868, opened 2022-07-28
Update the channel to which to upload data-parallel Python extensions

oleksandr-pavlyk:
This PR modifies the worfklow to upload built conda-tarballs to `py-dp-exts` conda channel instead of `dppy`.

The reasons being
  - the token associated with dppy channel has expired and we can not regenerate it
  - use of dppy acronym was deprecated in favor of 'data-parallel-python-extensions' and py-dp-exts reflects that


oleksandr-pavlyk:
Since `dppy` token has been regenerated, closing this for now.

PR #876, opened 2022-08-06
Fix build break with open-source intel/llvm DPC++

oleksandr-pavlyk:
Due to change in intel/llvm#6467 the build of SYCLInterface library is failing when using nightly sycl_bundle of intel/llvm DPC++.

This PR uses `__SYCL_COMPILER_VERSION` to keep both current oneAPI DPC++ and intel/llvm DPC++ working when compiling the library.

Note that `sycl::device::info::max_work_item_sizes` parameter is [no longer in the SYCL 2020 spec rev 5](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors), instead one should use `sycl::device::info::max_work_item_sizes<1>`, `sycl::device::info::max_work_item_sizes<2>`, `sycl::device::info::max_work_item_sizes<3>` and hence `dpctl` interfaces must be updated at some point.


oleksandr-pavlyk:
I wonder if it makes sense to define `dpctl.SyclDevice.max_work_item_sizes1d`, `dpctl.SyclDevice.max_work_item_sizes2d` and `dpctl.SyclDevice.max_work_item_sizes3d` alongside `dpctl.SyclDevice.max_work_item_sizes` and deprecate the latter. 

For older compilers the 1d, 2d version would simply take the first and the first two entries from the triplet returned by `max_work_item_sizes`.

oleksandr-pavlyk:
@diptorupd I have implemented `dpctl.SyclDevice.max_work_item_sizes1d` and high dimensional analogs. The `dpctl.SyclDevice.max_work_item_sizes` issues Python warning.  

The `DPCTLDevice_GetMaxWorkItemSizes` also carries deprecation warning. 

Frankly, I dislike introducing these warnings. I'd say we should just remove them. It is not used in `dpnp`, `numba_dpex` so I think it should be safe to do.

oleksandr-pavlyk:
Ok @dipto, I think I am going to merge this to fix build with intel/llvm DPC++ and address your feedback in a separate PR.

PR #879, opened 2022-08-10
Fix c interface

oleksandr-pavlyk:
One of DPCTLSyclInterface function was deprecated in #876, and `__attributes__((deprecated(what, repl))` was used.

This broke building of C extensions using the library. This PR uses `[[deprecated(msg)]]` instead, which is understood by both the gcc and icx.

An example of native Python extension written in C and that is using DPCTLSyclInterface is added. It is being compiled in the workflow `conda-package.yml` workflow when testing that examples can be compiled on Linux.

Incidentally, `dpctl.lsplatform` implementation replaced informational `print` with seemingly more appropriate use of `warnings.warn`.

@mingjie-intel


mingjie-intel:
Tested with numba-dpex, with the fix, it can be built correctly.

PR #881, opened 2022-08-12
Python array API conformity workflow added

ndgrigorian:
Tested locally here: https://github.com/ndgrigorian/dpctl/pull/4


oleksandr-pavlyk:
Please ping me when this is ready. The PR should contain a single commit, unless it could be partitioned into sequence of meaningful changes to that same file.

PR #897, opened 2022-09-03
Use [[deprecated("msg")]] annotation only for C++ builds

oleksandr-pavlyk:
Changed `libsyclinterface/include/dpctl_sycl_device_interface.h` to only use `[[deprecated("msg")]]` annotation for C++ builds.

This fixes build of numba-dpex's native extension. This remained broken despite claim to the contrary in #879.

@mingjie-intel


diptorupd:
LGTM! Please merge.

diptorupd (approved):

oleksandr-pavlyk:
Actually, perhaps the check can be relaxed somewhat. The C extension added in #879 builds fine with gcc 11.2 packaged in conda c compiler package, while It was failing for me locally with gcc 9.4 installed on Ubuntu 20.04.

oleksandr-pavlyk:
`[[deprecated]]` is part of C23 standard. It is supported in GCC, see https://en.cppreference.com/w/c/23 

This explains why it breaks for me on GCC 9.4, but works with Anaconda's compiler which is GCC 11.2

PR #900, opened 2022-09-07
Improve queue compatibility testing

oleksandr-pavlyk:
This introduces `dpctl::utils::queues_are_compatible(exec_q, {alloc_q1, alloc_q2, ...})` and deploys it in `tensor_py.cpp`.

The check replaces notion of compatibility from queues have equal contexts to queues are equal, aligning check with the logic in `_compute_follows_data.pyx`.

Furthermore, this PR implements optimization for transfer of shape/strides from host-allocated metadata to kernels.
Previously, transfer was done with 3 or 4 calls to `exec_q.copy`. It is replaced with allocation of host temporary where necessary `std::copy` calls are made to populate it, followed by a single call to `exec_q.copy` to copy data into USM allocation for use in kernel. 

No other logic has changed.

### DIFF
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -547,6 +547,19 @@ sycl::event keep_args_alive(sycl::queue q,
     return host_task_ev;
 }
 
+template <std::size_t num>
+bool queues_are_compatible(sycl::queue exec_q,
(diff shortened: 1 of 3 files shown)


diptorupd (changes requested):
I am fine in general, but the `copy_usm_ndarray_into_usm_ndarray` function has become too big. I feel a refactor is required for readability.

oleksandr-pavlyk:
Merging.

PR #908, opened 2022-09-12
Use pytest.warns instead of fixure filterwarning

oleksandr-pavlyk:
Improved `check_max_work_item_sizes` to use `pytest.warns` context manager instead of `pytest.marks.filterwarnings` decorator.

This is better since it assers that warning is issues, and filters it out, so test run of `test_sycl_device.py` is green, rather than darker-yellow.


ndgrigorian (commented):
LGTM

PR #911, opened 2022-09-17
Fix os-llvm-sycl-build workflow

oleksandr-pavlyk:
Since some sycl-nightly tags do no thave associated binaries attached the workflow was failing.

This commit enhances the download logic to go through recent tags until one is where where binary file is available.


oleksandr-pavlyk:
Workflow `os-llvm-sycl-build` is working now. 

Merging

PR #912, opened 2022-09-18
Improve performance of array metadata transfer to device

oleksandr-pavlyk:
Change to use USM-host allocation for packed shape/strides on the host. 
USM-host allocations take less time to copy them into USM-device allocations.

The difference is going to be especially pronounced on discrete GPU devices.


diptorupd:
@oleksandr-pavlyk re: "USM-host allocations take less time to copy them into USM-device allocations."

Less time compared to what? previously were we using regular malloc?

oleksandr-pavlyk:
`q.copy<T>(standard_alloc_vector.data(), usm_device_ptr, n)` is slower than `q.copy<T>(usm_host_alloc_vector.data(), usm_device_ptr, n)`. The PR changes to use the latter.

Issue #890, opened 2022-08-17
 Keyword 'order' not recognized. Expecting 'C' or 'F', got f


npolina4:
I can use uppercase and lowercase letters in some function. 
For example:
``` python
dpt.ones((2,2), order='f')
dpt.ones((2,2), order='F')
```
But I cannot use lowercase letter in dpt.reshape() function. 
``` python
import dpctl.tensor as dpt
a = dpt.ones((3,2))
dpt.reshape(a, (2,3), order='f')
ValueError: Keyword 'order' not recognized. Expecting 'C' or 'F', got f
```

oleksandr-pavlyk:
Fixed in #915

PR #915, opened 2022-09-21
Fixed issue #890 in reshape() function.

npolina4:
Added support for uppercase and lowercase letters for order argument to reshape() function.

### DIFF
--- a/dpctl/tensor/_reshape.py
+++ b/dpctl/tensor/_reshape.py
@@ -86,7 +86,11 @@ def reshape(X, newshape, order="C", copy=None):
         raise TypeError
     if not isinstance(newshape, (list, tuple)):
         newshape = (newshape,)
-    if order not in "cCfF":
+    if order in "c":
+        order = "C"
+    elif order in "f":
+        order = "F"
+    elif order not in ["C", "F"]:
         raise ValueError(
             f"Keyword 'order' not recognized. Expecting 'C' or 'F', got {order}"
         )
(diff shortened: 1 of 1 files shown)


dpctl/tensor/_reshape.py
```diff
@@ -86,7 +86,7 @@ def reshape(X, newshape, order="C", copy=None):
         raise TypeError
     if not isinstance(newshape, (list, tuple)):
         newshape = (newshape,)
-    if order not in ["C", "F"]:
+    if order not in "cCfF":
```
oleksandr-pavlyk, line 89: This is not sufficient. The keyword argument must also be normalized to the upper case.

The code below makes the assumption that `order` is either "C" or "F".

oleksandr-pavlyk (changes requested):

dpctl/tensor/_reshape.py
```diff
@@ -86,7 +86,11 @@ def reshape(X, newshape, order="C", copy=None):
         raise TypeError
     if not isinstance(newshape, (list, tuple)):
         newshape = (newshape,)
-    if order not in "cCfF":
+    if order in "c":
+        order = "C"
+    elif order in "f":
+        order = "F"
+    elif order not in ["C", "F"]:
```
oleksandr-pavlyk, line 93:
```suggestion
    if order in "cfCF":
        order = order.upper()
    else:
```

oleksandr-pavlyk (approved):
LGTM! Thank you @npolina4

PR #919, opened 2022-10-01
set CMAKE_FIND_ROOT_PATH_MODE_INCLUDE to BOTH

oleksandr-pavlyk:
In scikit-build 0.14.0 `-DCMAKE_FIND_PATH_MODE_INCLUDE=ONLY` started being set. This breaks working of IntelDPCPP integration with cmake, since `find_file` then ignores specified paths and hints.

Setting the value in `CMakeLists.txt` overrides that value


AndresGuzman-Ballen (dismissed):
looks good to me!

PR #921, opened 2022-10-03
Implements dpctl.tensor._flags

ndgrigorian:

Reworks usm_ndarray.flags to return a dpctl.tensor._flags object, which streamlines checking flags, such as array memory order and if it's writable.

e.g., for an array X, X.flags now prints the truth values of the currently implemented flags: C-contiguous, F-contiguous, and Writable, while X.flags.forc returns True if one of C-contiguous and F-contiguous is true, and False otherwise.

Note: this will fail the flake8 style guide check as flake8 does not recognize 'cimport' in _flags.pyx as valid syntax.

### DIFF
--- a/dpctl/tensor/_flags.pyx
+++ b/dpctl/tensor/_flags.pyx
@@ -18,16 +18,26 @@
 # cython: language_level=3
 # cython: linetrace=True
 
+from libcpp cimport bool as cpp_bool
+
 from dpctl.tensor._usmarray cimport (
     USM_ARRAY_C_CONTIGUOUS,
     USM_ARRAY_F_CONTIGUOUS,
     USM_ARRAY_WRITEABLE,
+    usm_ndarray,
 )
 
 
-class Flags:
+cdef cpp_bool _check_bit(int flag, int mask):
+    return (flag & mask) == mask
 
-    def __init__(self, arr, flags):
+
+cdef class Flags:
+    """Helper class to represent flags of :class:`dpctl.tensor.usm_ndarray`."""
+    cdef int flags_
+    cdef usm_ndarray arr_
+
+    def __cinit__(self, usm_ndarray arr, int flags):
         self.arr_ = arr
         self.flags_ = flags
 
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk:
Can you please either merge main branch into this PR, or rebase it on top of the main branch to fix the CI?

oleksandr-pavlyk:
@ndgrigorian The linter failed. Is pre-commit passing in your local workflow?

ndgrigorian:
> @ndgrigorian The linter failed. Is pre-commit passing in your local workflow?

Yes, but it skips flake8. When I run flake8 on _flags.pyx, it thinks cimport is a syntax error.

_dlpack.pxd fails flake8 runs for the same reason, but pre-commit passes it as well, because it skips flake8.

dpctl/tests/test_usm_ndarray_ctor.py
```diff
@@ -1402,7 +1402,7 @@ def test_triu_order_k(order, k):
     Xnp = np.arange(np.prod(shape), dtype="int").reshape(shape, order=order)
     Ynp = np.triu(Xnp, k)
     assert Y.dtype == Ynp.dtype
-    assert X.flags == Y.flags
+    assert X.flags.flags == Y.flags.flags
```
oleksandr-pavlyk, line 1405: Perhaps this change suggest that `_flags` object can use `__equal__` method implementation.

dpctl/tests/test_usm_ndarray_ctor.py
```diff
@@ -59,13 +59,13 @@ def test_allocate_usm_ndarray(shape, usm_type):
...
 
 def test_usm_ndarray_flags():
-    assert dpt.usm_ndarray((5,)).flags == 3
-    assert dpt.usm_ndarray((5, 2)).flags == 1
-    assert dpt.usm_ndarray((5, 2), order="F").flags == 2
-    assert dpt.usm_ndarray((5, 1, 2), order="F").flags == 2
-    assert dpt.usm_ndarray((5, 1, 2), strides=(2, 0, 1)).flags == 1
-    assert dpt.usm_ndarray((5, 1, 2), strides=(1, 0, 5)).flags == 2
-    assert dpt.usm_ndarray((5, 1, 1), strides=(1, 0, 1)).flags == 3
+    assert dpt.usm_ndarray((5,)).flags.flags == 3
```
oleksandr-pavlyk, line 62: Same this here, `_flags.__equal__` can implement the logic for comparing to integers, as well as to another `_flags` object.

But most like the test must change, e.g.

```python
assert dpt.usm_ndarray((5, 2)).flags["C"]
assert dpt.usm_ndarray((5, 2), order="F").flags["F"]
assert dpt.usm_ndarray((5, 1, 1), strides=(1, 0, 1)).flags["FC"]
```

dpctl/tests/test_usm_ndarray_ctor.py
```diff
@@ -1423,7 +1423,7 @@ def test_tril_order_k(order, k):
     Xnp = np.arange(np.prod(shape), dtype="int").reshape(shape, order=order)
     Ynp = np.tril(Xnp, k)
     assert Y.dtype == Ynp.dtype
-    assert X.flags == Y.flags
+    assert X.flags.flags == Y.flags.flags
```
oleksandr-pavlyk, line 1426: If `__equal__` was implemented, we could just say `X.flags == Y.flags`.

dpctl/tensor/_flags.pyx
```diff
@@ -0,0 +1,85 @@
...
+#  limitations under the License.
+
+# distutils: language = c++
+# cython: language_level=3
+# cython: linetrace=True
+
+from dpctl.tensor._usmarray cimport (
+    USM_ARRAY_C_CONTIGUOUS,
+    USM_ARRAY_F_CONTIGUOUS,
+    USM_ARRAY_WRITEABLE,
```
oleksandr-pavlyk, line 24: Once this PR and #923 are merged, `USM_ARRAY_WRITEABLE` should be renamed to syntactically correct `USM_ARRAY_WRITABLE`.

diptorupd:
> > @ndgrigorian The linter failed. Is pre-commit passing in your local workflow?
> 
> Yes, but it skips flake8. When I run flake8 on _flags.pyx, it thinks cimport is a syntax error.
> 
> _dlpack.pxd fails flake8 runs for the same reason, but pre-commit passes it as well, because it skips flake8.

@ndgrigorian Please add the following entry for the `_flags.pyx` in the `per-file-ignore` section of the `.flake8` config. 
> dpctl/tensor/_flags.pyx: E999, E225, E226, E227

@oleksandr-pavlyk We need to document how to handle flake8 related config for pyx files it in the CONTRIBUTING document.

oleksandr-pavlyk:
I have made changes warranting the use of Cython:
  - class `Flags` is now a Cython class, i.e. `cdef class Flags:` instead of `class Flags:`
  - class has two typed members, `usm_ndarray arr_` and `int flags_`
  - wrote typed bit checking routine, and used it to implement properties
  - implemented `__eq__` method

Also modified the test file to leverate `__eq__`, and to add tests to covert each property/method of `Flags` class.

oleksandr-pavlyk (dismissed):

oleksandr-pavlyk:
@ndgrigorian Please look over the changes and merge.

ndgrigorian:
> @ndgrigorian Please look over the changes and merge.

I've committed a small change: Numpy's FNC flag is F-contiguous-and-not C-contiguous, so I moved the previous implementation of fnc to a new property, fc, and changed fnc to match Numpy's. I also adjusted the tests to reflect that.

oleksandr-pavlyk (approved):

oleksandr-pavlyk:
Given my confusion of the meaning of `Flags.fnc`, perhaps property docstrings are in order. They wont be user accessible, since Cython does support this yet, but it would be inspectable through source code. This can be filed as an issue and added in a separate pr.

PR #923, opened 2022-10-07
Cleanup tensor_py

oleksandr-pavlyk:
This housekeeping PR breaks down the monolithic, single-translation-unit `tensor_py.cpp` into bunch of individual `.cpp` files implementing individual Python binding functions.

This separation should improve incremental rebuilding, make it easier to add more functions in separate `.hpp` and `.cpp` files.

It also factors out kernel implementations into stand-alone header files which can potentially be reused.

While refactoring, several problems were fixed:
 1. fixed sneaking "[static initialization order fiasco](https://en.cppreference.com/w/cpp/language/siof)" present in the sources but masked by use of the single translation unit.
 2. fixed missing `api` keyword in `_memory.pxd` for `PyMemoryUSMDevice` class
 3. refactored `_tensor_imp.default_device_fp_type`, et. al. functions.

### DIFF
--- a/dpctl/.gitignore
+++ b/dpctl/.gitignore
@@ -1,5 +1,5 @@
 *.so
-*.cpp
+_*.cpp
 *.cxx
 *.c
 *.h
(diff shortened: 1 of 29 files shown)


oleksandr-pavlyk:
I am going to close this PR, and open smaller PRs for review.

PR #925, opened 2022-10-08
Moved implementation of kernels out to dedicated header files.

oleksandr-pavlyk:
This PR moves out implementations of implementation kernels into separate header files.

This PR is the first of the series of smaller PRs separated from #923.

### DIFF
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -371,6 +371,55 @@ protected:
 
 namespace tensor
 {
+
+std::vector<py::ssize_t> c_contiguous_strides(int nd,
+                                              const py::ssize_t *shape,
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk:
This PR is the first step to be merged to `cleanup/tensor` branch which will incorporate all the changes from #923, and will then be merged to master.

diptorupd:
@oleksandr-pavlyk A general comment is regarding the documentation of the code. As these kernels can potentially be used outside of dpctl, we definitely need docstrings and API docs generated.

oleksandr-pavlyk:
Yes, I will add it, but it would be best to do it in a separate PR to the target branch. I want to check in other pending changes which I have restructured into separate commits.

oleksandr-pavlyk:
@diptorupd I have pushed changes adding doxygen docstrings for exported templated `*_impl` functions in added header files.

PR #932, opened 2022-10-13
[cleanup/tensor, step 3] Implement dpctl_capi loader in dpctl4pybind11 header file

oleksandr-pavlyk:
This PR creates `dpctl::detail::dpctl_capi` singleton class that load Cython-generated CAPI functions implemented in Cython modules of dpctl, and re-exports them as class members. 

This PR removes another singleton `dpctl::tensor::detail::usm_ndarray_types` which is not a simple struct with trivial constructor.

This moves all initialization of static variables exported and implemented by Python into a single place, thus avoiding static order initialization fiasco.

### DIFF
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -179,9 +179,15 @@ private:
         : default_sycl_queue{}, default_usm_memory{}, default_usm_ndarray{},
           as_usm_memory{}
     {
-        // Import Cython CAPI for dpctl
+        // Import Cython-generated C-API for dpctl
+        // This imports python modules and initializes
+        // static variables such as function pointers for C-API,
+        // e.g. SyclDevice_GetDeviceRef, etc.
+        // pointers to Python types, i.e. PySyclDeviceType, etc.
+        // and exported constants, i.e. USM_ARRAY_C_CONTIGUOUS, etc.
         import_dpctl();
 
+        // Python type objects for classes implemented by dpctl
         this->Py_SyclDeviceType_ = &Py_SyclDeviceType;
         this->PySyclDeviceType_ = &PySyclDeviceType;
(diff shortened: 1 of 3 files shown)


dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -34,6 +34,283 @@
...
+
+    static auto &get()
+    {
+        static dpctl_capi api = lookup();
+        return api;
+    }
+
+    py::object default_sycl_queue_pyobj()
+    {
+        return *default_sycl_queue;
```
diptorupd, line 145: Where is the `default_sycl_queue` pointer defined? If it is a global, then a comment should indicate that and the name of the variable should reflect that it is a global (e.g. `g_default_sycl_queue`).

diptorupd: Never mind found it 30 lines below :)

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -34,6 +34,283 @@
...
+    std::shared_ptr<py::object> as_usm_memory;
+
+    dpctl_capi()
+        : default_sycl_queue{}, default_usm_memory{}, default_usm_ndarray{},
+          as_usm_memory{}
+    {
+        // Import Cython CAPI for dpctl
+        import_dpctl();
+
+        this->Py_SyclDeviceType_ = &Py_SyclDeviceType;
```
diptorupd, line 185: All these are global pointers that are initialized by `import_dpctl`? We need comments at least here. Moreover, can we create aliases for when we are initializing the static variables `SyclDevice_GetDeviceRef` _etc._ that have a name such as `gSyclDevice_GetDeviceRef`? The idea is to make the code easier to follow.

diptorupd (approved):
Overall looks good.

diptorupd (approved):

PR #933, opened 2022-10-15
[cleanup/tensor, part 4] Made contract_iter and contract_iter2 functions templated to avoid needing to include pybind11 headers

oleksandr-pavlyk:
Code for `contract_iter` and `contract_iter2` functions in "strided_iters.hpp" used identifiers from `::pybind11` namespace.

Made these functions templated passing necessary types via template parameters.

This is avoid using py::value_error and py::ssize_t and py::type_error in there explicitly.

Use in "tensor_py.cpp" has been adjusted.


oleksandr-pavlyk:
I think this is pretty straightforward commit, so I am going to merge it.

Issue #948, opened 2022-10-19
dpctl.tensor.asarray incorrect dtype


npolina4:
``` python
import dpctl.tensor as dpt
import numpy as np

a = np.ndarray(1, dtype=np.uint64)
b = dpt.asarray(a)
print('a.dtype =', a.dtype, '\nb.dtype =', b.dtype)
```
output:
```
a.dtype = uint64
b.dtype = int64
```

oleksandr-pavlyk:
Fixed by #949

PR #949, opened 2022-10-21
Fixed issue in asarray_from_numpy for uint64 dtype. 

npolina4:
Fixed issue #948  in asarray_from_numpy for uint64 dtype. And changed call copy_to function

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -193,6 +193,28 @@ def _asarray_from_usm_ndarray(
     return res
 
 
+def _map_to_device_dtype(dt, q):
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk:
FYI, change from `res[...] = ary` to call to `ti._copy_numpy_ndarray_into_usm_ndarray(src=ary, dst=res, sycl_queue=copy_q)` break test runs on Iris Xe, because logic in `__setitem__` adapts to this scenario by downcasting on the host.

oleksandr-pavlyk:
Please run `pre-commit install` in your environment.

oleksandr-pavlyk:
I pushed a commit to address the weakness in deducing `asarray(x)` output type from the input type and the device information.

PR #959, opened 2022-11-02
Rebuild with 2023 performance libraries

AndresGuzman-Ballen:
Increment build number so it builds with 2023 performance libraries


ekomarova:
This will not work now, as there are no necessary changes to gold/2021

PR #960, opened 2022-11-03
Header for dpctl sycl interface type casters

oleksandr-pavlyk:
This PR adds "libsyclinterface/include/dpctl_sycl_type_casters.hpp" which defines 
`dpctl::syclinterface::dpctl_device_selector` class corresponding to `DPCTLDeviceSelectorRef` opaque pointer.

The class is derived from `sycl::device_selector` to support oneAPI DPC++ 2022.2 and older, but is a stand-alone
class if `__SYCL_COMPILER_VERSION >= 20221020L`. 

Use of `dpctl_device_selector` in `sycl::device` and `sycl::platform` constructors is adjusted based on the compiler version too.

This PR also exposes `dpctl::syclinterface::wrap<T>` and `dpctl::syclinterface::unwrap<T>` templated functions where the template type must be the underlying sycl type to the pointer of which the opaque pointer should be `reinterpret_cast`-ed.

For example:

```c++
   DPCTLSyclDeviceRef DRef;
   auto Dev = ::dpctl::syclinterface::unwrap<sycl::device>(DRef);
```

libsyclinterface/source and libsyclinterface/tests are adjusted to use exported wrap/unwrap.


oleksandr-pavlyk:
Should this PR also remove setting of `SYCL_ENABLE_HOST_DEVICE=1` in workflows?

libsyclinterface/helper/source/dpctl_utils_helper.cpp
```diff
@@ -242,9 +251,13 @@ std::string DPCTL_AspectToStr(aspect aspectTy)
 aspect DPCTL_StrToAspectType(const std::string &aspectTyStr)
 {
     aspect aspectTy;
-    if (aspectTyStr == "host") {
+    if (false) {
```
diptorupd, line 254: Why not just move the "cpu" check on top?

libsyclinterface/include/dpctl_device_selection.hpp
```diff
@@ -0,0 +1,167 @@
...
+#include <CL/sycl.hpp>
+
+namespace dpctl
+{
+namespace syclinterface
+{
+
+#if __SYCL_COMPILER_VERSION >= __SYCL_COMPILER_2023_SWITCHOVER
+
+class DPCTL_API dpctl_device_selector
```
diptorupd, line 40: As discussed this class should be in a separate header.

diptorupd (dismissed):
I did not go very deep as the changes are fairly mechanical and we have already discussed them offline.

Please merge.

libsyclinterface/helper/source/dpctl_utils_helper.cpp
oleksandr-pavlyk: Good point

PR #963, opened 2022-11-04
Add missing docstrings to SyclDevice properties.

### LINKED ISSUE #962: Suggested improvement to API docs
I need to use the `max_work_item_sizes` property of `SyclDevice` in Numba-dpex. The API doc does not specify the return type of the property, nor does it specify the deprecation warning. Both these things should be specified in the docstring so that one does not need to
refer the source code.

diptorupd:
Several properties of the SyclDevice class are undocumented or the documentation is incomplete (missing return attribute and deprecation notices). The PR adds and updates the missing docstring entries.

Closes #962

### DIFF
--- a/dpctl/_sycl_device.pyx
+++ b/dpctl/_sycl_device.pyx
@@ -393,61 +393,115 @@ cdef class SyclDevice(_SyclDevice):
 
     @property
     def has_aspect_host(self):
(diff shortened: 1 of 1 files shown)


diptorupd:
@oleksandr-pavlyk what is the `atomic64` aspect that we support? I could not find it in the spec.

PR #982, opened 2022-11-08
Added support for arrays for fill_value for full() function

npolina4:
Added support for numpy.ndarray and dpctl.tensor.usm_ndarray for fill_value for dpctl.tensor.full() function.


dpctl/tensor/_ctors.py
```diff
@@ -763,6 +763,8 @@ def full(
         order = order[0].upper()
     dpctl.utils.validate_usm_type(usm_type, allow_none=False)
     sycl_queue = normalize_queue_device(sycl_queue=sycl_queue, device=device)
+    if dtype is None and isinstance(fill_value, (dpt.usm_ndarray, np.ndarray)):
+        dtype = fill_value.dtype
```
oleksandr-pavlyk, line 767: This is not enough, I am afraid. If `fill_value` is `np.ndarray`, we need to ensure it has just 1 element, to extract the scalar from it, and get `dtype` like you are doing. 

If `fill_value` is `usm_ndarray` we need to check that compute-follows-data rules apply, then broadcast `fill_value` to `sh`, and apply copy-with-casting operator.

All other cases must proceed as before.

dpctl/tensor/_ctors.py
```diff
@@ -761,8 +761,20 @@ def full(
...
         order = order[0].upper()
-    dpctl.utils.validate_usm_type(usm_type, allow_none=False)
+    dpctl.utils.validate_usm_type(usm_type, allow_none=True)
     sycl_queue = normalize_queue_device(sycl_queue=sycl_queue, device=device)
+
+    if isinstance(fill_value, (dpt.usm_ndarray, np.ndarray, tuple, list)):
+        X = dpt.asarray(
+            fill_value,
+            dtype=dtype,
+            device=device,
```
oleksandr-pavlyk, line 771:
```suggestion
```

Since `normalize_queue_device` has been called, we only need to use `sycl_queue` argument.

oleksandr-pavlyk:
More tests are needed. Check support for working with 0d arrays, e.g.

```
y_sum = dpt.arange(10, dtype='i4')
x_usm = dpt.full(10, y_usm[3])
```

oleksandr-pavlyk (approved):
Thank you @npolina4 !

This looks great. Let's wait for the CI and merge.

PR #990, opened 2022-11-15
Expanded doc-string of dpctl.tensor.usm_ndarray.to_device

oleksandr-pavlyk:
Expanded doc-string for `to_device` method of `usm_ndarray` class.

### DIFF
--- a/dpctl/tensor/_usmarray.pyx
+++ b/dpctl/tensor/_usmarray.pyx
@@ -638,13 +638,43 @@ cdef class usm_ndarray:
         res.array_namespace_ = self.array_namespace_
         return res
 
-    def to_device(self, target_device):
+    def to_device(self, target):
         """
-        Transfer array to target device
+        Transfers this array to specified target device.
+
+        :Example:
+            .. code-block:: python
+
+                import dpctl
+                import dpctl.tensor as dpt
+
+                x = dpt.full(10**6, 2, dtype="int64")
+                q_prof = dpctl.SyclQueue(
+                    x.sycl_device, property="enable_profiling")
+                # return a view with profile-enabled queue
+                y = x.to_device(q_prof)
+                timer = dpctl.SyclTimer()
+                with timer(q_prof):
+                    z = y * y
+                print(timer.dt)
+
+        Args:
+            target: array API concept of target device.
+                It can be a oneAPI filter selector string,
+                an instance of :class:`dpctl.SyclDevice` corresponding to a
+                non-partitioned SYCL device, an instance of
+                :class:`dpctl.SyclQueue`, or a :class:`dpctl.tensor.Device`
+                object returned by :attr:`dpctl.tensor.usm_array.device`.
+
+        Returns:
+            A view if data copy is not required, and a copy otherwise.
+            If copying is required, it is done by copying from the original
+            allocation device to the host, followed by copying from host
+            to the target device.
         """
         cdef c_dpctl.DPCTLSyclQueueRef QRef = NULL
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
The failure of pre-commit workflow is being addressed in #991

npolina4 (approved):

dpctl/tensor/_usmarray.pyx
```diff
@@ -638,13 +638,43 @@ cdef class usm_ndarray:
...
+                import dpctl.tensor as dpt
+
+                x = dpt.full(10**6, 2, dtype="int64")
+                q_prof = dpctl.SyclQueue(
+                    x.sycl_device, property="enable_profiling")
+                # return a view with profile-enabled queue
+                y = x.to_device(q_prof)
+                timer = dpctl.SyclTimer()
+                with timer(q_prof):
+                    z = y * y
```
antonwolfy, line 658: I've tried the example, but got an exception:
```
In [1]: import dpctl
   ...: import dpctl.tensor as dpt
   ...:
   ...: x = dpt.full(10**6, 2, dtype="int64")
   ...: q_prof = dpctl.SyclQueue(
   ...:     x.sycl_device, property="enable_profiling")
   ...: # return a view with profile-enabled queue
   ...: y = x.to_device(q_prof)
   ...: timer = dpctl.SyclTimer()
   ...: with timer(q_prof):
   ...:     z = y * y
   ...: print(timer.dt)
Input In [1], in <cell line: 10>()
      9 timer = dpctl.SyclTimer()
     10 with timer(q_prof):
---> 11     z = y * y
     12 print(timer.dt)

TypeError: unsupported operand type(s) for *: 'dpctl.tensor._usmarray.usm_ndarray' and 'dpctl.tensor._usmarray.usm_ndarray'

```
but if I add `import dpnp` and run exactly the same code again, it will pass:
```
In [2]: import dpnp

In [3]: import dpctl
   ...: import dpctl.tensor as dpt
   ...:
   ...: x = dpt.full(10**6, 2, dtype="int64")
   ...: q_prof = dpctl.SyclQueue(
   ...:     x.sycl_device, property="enable_profiling")
   ...: # return a view with profile-enabled queue
   ...: y = x.to_device(q_prof)
   ...: timer = dpctl.SyclTimer()
   ...: with timer(q_prof):
   ...:     z = y * y
   ...: print(timer.dt)
(0.10838821064680815, 0.10794714400000001)
```
since, as I've got, dpctl will dispatch to `dpnp.multiply` then.
I believe `import dpnp` is missing in the example.

antonwolfy (commented):
I've checked the generated page and it looks that:

1. The copyright year is not updated:
```
 Copyright 2020-21, Intel Corp..
```
2. The link below from `usm_ndarray.usm_type` description isn't working:
```
USM type of underlying memory. Can be ‘device’, ‘shared’, or ‘host’.

See: https://docs.oneapi.com/versions/latest/dpcpp/iface/usm.html
```

dpctl/tensor/_usmarray.pyx
oleksandr-pavlyk: Yes, this example presently requires `dpnp`, because `dpctl.tensor.__mul__` falls back to `dpnp` if present or returns `NotImplemented`. This is transient and support for Python arithmetic operations is coming soon.

For the purpose of illustrating `usm_ndarray.to_device` I probably should have used `z = foo(y)` to emphasize that any offloading routine we want to time could be used.

oleksandr-pavlyk:
Merging. Will update copyright years with a tool in a separate commit.

PR #995, opened 2022-11-18
Fixed missing copy for dpctl.tensor.full() function if fill_value is array

npolina4:
Fixed missing copy for full() function if fill_value is array.
Added tests that check order and strides for dpctl.tensor.full() function.


oleksandr-pavlyk (approved):
LGTM! Thank you @npolina4

PR #999, opened 2022-11-25
Fixed gh-998: exception raised while forming an expected TypeError exception message  

### LINKED ISSUE #998: Failed to build TypeError exception
The example code should cause `TypeError` exception in dpctl:
```
import numpy
import dpctl.tensor as dpt
obj = dpt.empty((6,), dtype=numpy.object_)
```
but it failed to build the exception due to bellow error:
```
File dpctl/tensor/_usmarray.pyx:193, in dpctl.tensor._usmarray.usm_ndarray.__cinit__()

TypeError: can only concatenate str (not "numpy.dtype[object_]") to str
```

oleksandr-pavlyk:
This change fixes formation of `TypeError` message string.

Closes #998. 

It also reworks `dtype` argument validation, fixing unreferenced issue with `dpt.usm_ndarray(1, dtype=np.dtype(">f4"))`.

### DIFF
--- a/dpctl/tensor/_types.pxi
+++ b/dpctl/tensor/_types.pxi
@@ -102,7 +102,7 @@ cdef str _make_typestr(int typenum):
     return type_to_str[typenum] + str(type_bytesize(typenum))
 
 
-cdef int typenum_from_format(str s) except *:
+cdef int typenum_from_format(str s):
     """
     Internal utility to convert string describing type format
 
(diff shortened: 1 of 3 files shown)


dpctl/tensor/_types.pxi
```diff
@@ -143,9 +144,11 @@ cdef int dtype_to_typenum(dtype) except *:
     else:
         try:
             dt = np.dtype(dtype)
-            if hasattr(dt, 'descr'):
-                return descr_to_typenum(dt)
-            else:
-                return -1
+        except TypeError:
```
antonwolfy, line 147: I believe `except *` in the definition of `dtype_to_typenum` isn't needed anymore also, since no exception is raised by the function.

antonwolfy (dismissed):

dpctl/tensor/_types.pxi
oleksandr-pavlyk: Correct. This is an optimization change though. It does not hurt to keep it.

I will push the suggested change and merge.

PR #1000, opened 2022-11-25
Fixed marginal memory leak in singleton class dpctl_capi 

oleksandr-pavlyk:
dpctl_capi ctor initialize all members, dtor resets smart pointers

This addressed accessing uninitialized variables SDL tool warnings.


oleksandr-pavlyk:
Additional scan of this PR by the tool confirmed that access of uninitialized variable warnings and memory leak warnings are gone.

PR #1008, opened 2022-12-06
MAINT Include tuple in 'utils/strided_iters.hpp'

jjerphan:
Fixes https://github.com/IntelPython/dpctl/issues/1007.

This include directive seems to have been forgotten and it is impossible for me to compile dpctl without it.

---

- [x] Have you provided a meaningful PR description?
- [x] Have you added a test, reproducer or referred to an issue with a reproducer?
- [ ] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [x] If this PR is a work in progress, are you filing the PR as a draft?

PR #1009, opened 2022-12-06
Added step to display version of clang-format used in pre-commit workflow

oleksandr-pavlyk:
Given unexpected pre-commit failure in #1008 due to `clang-format`, add a step to `pre-commit.yaml` workflow to output the version of `clang-format` used. 

This would allow to quickly tell if the pre-commit failure was caused by version update of `clang-format` in GitHub runner image.


oleksandr-pavlyk:
`clang-format` was updated to `14.0.0-1ubuntu1`. 

`clang-format` started to change some source files which were up-to-snuff with earlier version. This likely means that our file `.clang-format` should be more stringent.

oleksandr-pavlyk:
The update to `clang-format` came with update of `ubuntu-latest` image to Ubuntu 22.04. Hence fixed it to Ubuntu 20.04, as this is the version we currently develop on.

PR #1024, opened 2022-12-19
Step NumPy pin to 1.22 in conda build config

antonwolfy:
Intel 2023.0 release includes NumPy 1.22.
Thus `dpctl` should switch pinning NumPy version to the same value in config file.

Otherwise it is impossible to correctly build `dpctl`, when there is no new 2023 compiler in the host environment. This is because the current NumPy 1.21 can't be installed with the new 2023 compiler.
As a result it leads to improper linkage of `libDPCTLSyclInterface.so` with `libsycl.so.6`, which came with 2023 compiler only and present in build env, but missing in the host env.


antonwolfy:
A verification job `array-api-conformity` seems failed, because the PR is from forked repository and entire step couldn't post the result as a PR comment.
While the result of conformance tests looks OK (the same as in previous runs):
```
    MESSAGE: Array API standard conformance tests for dpctl=0.14.1dev0=py310h76be34b_24 ran successfully.
    Passed: 33
    Failed: 801
    Skipped: 280
```

PR #1026, opened 2022-12-20
Use USMHost as underlying buffer for result

oleksandr-pavlyk:
This should improver transfer timing per GPU optimization guidelines

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -46,7 +46,10 @@ def _has_memory_overlap(x1, x2):
 def _copy_to_numpy(ary):
     if not isinstance(ary, dpt.usm_ndarray):
         raise TypeError
-    h = ary.usm_data.copy_to_host().view(ary.dtype)
+    nb = ary.usm_data.nbytes
+    hh = dpm.MemoryUSMHost(nb, queue=ary.sycl_queue)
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
This change results in dramatic speed-up of `dpt.asnumpy`. For example, before the change:

```

In [2]: x = dpt.ones(10**7 + 2, dtype='i2', device='gpu')

In [3]: %time y = dpt.asnumpy(x)
CPU times: user 285 µs, sys: 140 ms, total: 141 ms
Wall time: 100 ms

In [4]: %timeit y = dpt.asnumpy(x)
70.7 ms ± 4.13 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

after:

```

In [2]: x = dpt.ones(10**7 + 2, dtype='i2', device='gpu')

In [3]: %time y = dpt.asnumpy(x)
CPU times: user 19.3 ms, sys: 69.4 ms, total: 88.7 ms
Wall time: 69.4 ms

In [4]: %timeit y = dpt.asnumpy(x)
15.9 ms ± 956 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Consequentially, it improves performance of `__repr__`.

ndgrigorian (approved):
LGTM!

PR #1034, opened 2023-01-10
Use no associative math

oleksandr-pavlyk:
This PR changes to `_tensor_impl` with `-fno-associate-math` compiler flag. See https://clang.llvm.org/docs/UsersManual.html#controlling-floating-point-behavior for how to control FP-behavior in clang.

This fixes output of `dpt.linspace(dpt.finfo('f4').max, dpt.finfo('f4').max, num=16, dtype='f4')` which unexpectedly contained `nan` values as discovered by @npolina4 


oleksandr-pavlyk:
@npolina4 I fixed the issue by using `sycl::fma` to prevent compiler from reordering terms in multiplication and creating sub-expressions prone to overflowing.

npolina4 (approved):

PR #1050, opened 2023-01-30
Add usm_ndarray creation c-api

### LINKED ISSUE #1049: Add C-API function UsmNDArray_MakeFromMemory  and UsmNDArray_SetWritableFlag
This issue is to implement `dpctl.tensor` C-API function 

```
cdef api object UsmNDArray_MakeFromMemory(int nd, Py_ssize_t *shape, int typenum, _Memory mobj, Py_ssize_t offset)
```

to create a `usm_ndarray` from given memory object and offset.

Also to implement

```
cdef api void UsmNDArray_SetWritableFlag(usm_ndarray arr, int flag)
```

See https://github.com/IntelPython/numba-dpex/issues/890#issuecomment-1407659316 for context.

oleksandr-pavlyk:
Closes #1049 

This PR adds 3 new C-API for usm_ndarray: 

- `void UsmNDArray_SetWritableFlag(usm_array arr, int flag)` : Set WRITABLE flag of `arr` if `bool(flag)` is `True`, and clears it otherwise
- `object UsmNDArray_MakeFromMemory(int nd, const Py_ssize_t *shape, int typenum, c_dpmem._Memory memobj, Py_ssize_t offset)`: create `usm_ndarray` of given shape and data-type from USM-memory object and offset
- `object UsmNDArray_MakeFromPtr(size_t nelems, int typenum, DPCTLSyclUSMRef ptr, DPCTLSyclQueueRef QRef, object owner)`:  create 1D `usm_ndarray`  of requested data-type from pointer and queue reference; `owner=None` is interpreter as transfer of memory ownership, and memory will be deallocated once the usm_ndarray instance is deleted. If `owner` is not `None`, deleter of `usm_ndarray` instance will call the deleter of `owner` which is supposed to clean up the USM allocation. The `owner` object must thus capture the pointer and the queue reference, or the context to make the correct call to `sycl::free`.

`ctypes`-based tests are added to the test-suite.

New C-API functions are added to `dpctl::detail::dpctl_capi` singleton class, but are yet unused. 

@samir-nasibli

### DIFF
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -114,6 +114,17 @@ public:
     int (*UsmNDArray_GetFlags_)(PyUSMArrayObject *);
     DPCTLSyclQueueRef (*UsmNDArray_GetQueueRef_)(PyUSMArrayObject *);
     py::ssize_t (*UsmNDArray_GetOffset_)(PyUSMArrayObject *);
+    void (*UsmNDArray_SetWritableFlag_)(PyUSMArrayObject *, int);
+    PyObject *(*UsmNDArray_MakeFromMemory_)(int,
+                                            const py::ssize_t *,
+                                            int,
+                                            Py_MemoryObject *,
+                                            py::ssize_t);
+    PyObject *(*UsmNDArray_MakeFromPtr_)(size_t,
+                                         int,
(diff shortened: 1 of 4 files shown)


oleksandr-pavlyk:
Right now `UsmNDArray_MakeFromMemory` creates a C-contiguous array. Should the api allow for `order` keyword?

PR #1051, opened 2023-01-30
Changed generating integer output for linspace() function.

npolina4:
Changed generating integer output for dpctl.tensor.linspace() function like in numpy.linspace().

``` python
xp.linspace(0.1, 9.1, 11, endpoint=True, dtype=int)
# old out:
array([0, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
# new out:
array([0, 1, 1, 2, 3, 4, 5, 6, 7, 8, 9])
# numpy out:
array([0, 1, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -1119,7 +1119,7 @@ def linspace(
     num = operator.index(num)
     if num < 0:
         raise ValueError("Number of points must be non-negative")
-    ((start, stop,), dt) = _coerce_and_infer_dt(
+    _, dt = _coerce_and_infer_dt(
         start,
         stop,
         dt=dtype,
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk:
@npolina4 Could you please provide an example of input where the suggested change makes a difference?

oleksandr-pavlyk:
:+1: for fixing `usm_type` behavior. 

The change modifies behavior outside of scope of array-API spec  (presumable less common/useful combination of parameters). It is an improvement towards aligning with NumPy's behavior, but it only goes half-way (does not fix negative integers):

```
In [7]: np.linspace(-1.9,9.1, 13, endpoint=True, dtype=int)
Out[7]: array([-2, -1, -1,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9])

In [8]: dpt.linspace(-1.9,9.1, 13, endpoint=True, dtype=int)
Out[8]: usm_ndarray([-1,  0,  0,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9])
```

Nonetheless, this change does not hurt either, so I am approving it. 

I think we should change C++ kernel to handle this without additional copy, but this task is not an immediate priority, in my opinion. This note is to create a new issue to make sure this does become forgotten.

oleksandr-pavlyk (approved):

PR #1053, opened 2023-01-31
Added dtype compatibility check for device for result_type() and can_cast() functions.

npolina4:
Added dtype compatibility check for device for dpctl.tensor.result_type() and dpctl.tensor.can_cast() functions.


oleksandr-pavlyk:
Superseded by gh-1488

PR #1061, opened 2023-02-04
Changes the API for DPCTLUSM_GetPointerType.

diptorupd:
    - DPCTLUSM_GetPointerType used to return a global string literal
      and had an API that is different from all other libsyclinterface
      functions that return a C string.

      Returning a pointer to a global literal as opposed to a heap
      allocated memory meant the ownership of the pointer was not
      clear when crossing language boundary.

      Calling DPCTLCString_Delete on the previously returned const char*
      will cause a segfault

    - The PR introduces an enum for usm types and returns an enum value
      from DPCTLUSM_GetPointerType.

    - All Python API is unaffected.

Existing unit tests have been updated.

No

Yes

### DIFF
--- a/dpctl/_backend.pxd
+++ b/dpctl/_backend.pxd
@@ -34,6 +34,12 @@ cdef extern from "syclinterface/dpctl_utils.h":
 
 
 cdef extern from "syclinterface/dpctl_sycl_enum_types.h":
+    ctypedef enum _usm_type 'DPCTLSyclUSMType':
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk (dismissed):
LGTM! Thanks @diptorupd

oleksandr-pavlyk (dismissed):

oleksandr-pavlyk (approved):

PR #1064, opened 2023-02-07
Merge new content from master to gold/2021

ekomarova:
This PR is necessary to rebuild the chain of gpu components with new perflibraries 2023.1, and for this we need to add new content to gold/2021


ekomarova:
Since numpy version is no longer set in recipe on master, and after the PR is merged, it will also not be set on gold, it will be set as conda-build parameter `--numpy`

oleksandr-pavlyk:
I am going to close this PR and do it another way.

PR #1072, opened 2023-02-15
Intel LLVM is to use conda's gcc toolchain, sysroot and target libraries

oleksandr-pavlyk:
Use compiler CFG files for Intel LLVM to use conda's sysroot and toolchain


AndresGuzman-Ballen (dismissed):
Thought I saw unbalanced "" but it turns out you're just adding those flags to the file ;)

looks good to me!

Issue #1071, opened 2023-02-14
from_dlpack raises unexpected error on strided dpctl.tensor.usm_ndarray input


oleksandr-pavlyk:
```
In [13]: x_dpt = dpt.reshape(dpt.arange(4*30),(10,12))[::2,::-2]

In [14]: x_dpt.shape, x_dpt.strides, x_dpt.__sycl_usm_array_interface__['offset']
Out[14]: ((5, 6), (24, -2), 11)

In [15]: dpt.from_dlpack(x_dpt)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [15], in <cell line: 1>()
----> 1 dpt.from_dlpack(x_dpt)

File ~/repos/dpctl/dpctl/tensor/_dlpack.pyx:449, in dpctl.tensor._dlpack.from_dlpack()

File ~/repos/dpctl/dpctl/tensor/_dlpack.pyx:499, in dpctl.tensor._dlpack.from_dlpack()

File ~/repos/dpctl/dpctl/tensor/_dlpack.pyx:435, in dpctl.tensor._dlpack.from_dlpack_capsule()

File ~/repos/dpctl/dpctl/tensor/_usmarray.pyx:257, in dpctl.tensor._usmarray.usm_ndarray.__cinit__()

ValueError: buffer='<SYCL(TM) USM-device allocation of 856 bytes at 0xffffb80200020400>' can not accomodate the requested array.
```

The code apparently does not take into account non-zero offset correctly, since strided input with zero offset works fine:

```

In [16]: x_dpt = dpt.reshape(dpt.arange(4*30),(10,12))[::2,::2]

In [17]: x_dpt.shape, x_dpt.strides, x_dpt.__sycl_usm_array_interface__['offset']
Out[17]: ((5, 6), (24, 2), 0)

In [18]: dpt.from_dlpack(x_dpt)
Out[18]:
usm_ndarray([[  0,   2,   4,   6,   8,  10],
             [ 24,  26,  28,  30,  32,  34],
             [ 48,  50,  52,  54,  56,  58],
             [ 72,  74,  76,  78,  80,  82],
             [ 96,  98, 100, 102, 104, 106]])
```

PR #1073, opened 2023-02-16
Fix gh 1071 from dlpack

oleksandr-pavlyk:
Fixes gh-1071. Updated DLPack to 0.8 with support for `kDLBool` elemental tensor data type.

Exception reported #1071 would arise every time array would not start at the beginning of the underlying allocation, i.e. `x0 = dpt.arange(6); x1 = x0[::-2]; dpt.from_dlpack(x1)` fails, but `x0 = dpt.arange(7); x1 = x0[::-2]; dpt.from_dlpack(x1)` works. 

Test was added to cover such inputs too.


antonwolfy (approved):
Try to run different examples, no issue observed. LGTM!

PR #1091, opened 2023-03-01
Try out a hunch that crashes are caused by dangling host_task-s.

oleksandr-pavlyk:
Added synchronization for every `host_task` that is supposed to manage life-time of temporary vector with usm_host allocator.

- [x] If this PR is a work in progress, are you filing the PR as a draft?

### DIFF
--- a/dpctl/tensor/libtensor/source/integer_advanced_indexing.cpp
+++ b/dpctl/tensor/libtensor/source/integer_advanced_indexing.cpp
@@ -599,22 +599,21 @@ usm_ndarray_take(dpctl::tensor::usm_ndarray src,
     std::shared_ptr<shT> host_ind_offsets_shp =
         std::make_shared<shT>(k, ind_allocator);
 
-    std::copy(ind_sh_sts.begin(), ind_sh_sts.end(),
-              host_ind_shapes_strides_shp->begin());
-    std::copy(ind_ptrs.begin(), ind_ptrs.end(), host_ind_ptrs_shp->begin());
-    std::copy(ind_offsets.begin(), ind_offsets.end(),
-              host_ind_offsets_shp->begin());
-
     std::vector<sycl::event> host_task_events;
     host_task_events.reserve(5);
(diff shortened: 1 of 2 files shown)


dpctl/tensor/libtensor/source/integer_advanced_indexing.cpp
```diff
@@ -1078,9 +1094,10 @@ usm_ndarray_put(dpctl::tensor::usm_ndarray dst,
             sycl::free(packed_ind_offsets, ctx);
         });
     });
+    host_task_events.push_back(put_generic_ev);
 
     return std::make_pair(
-        keep_args_alive(exec_q, {dst, py_ind, val}, {put_generic_ev}),
+        keep_args_alive(exec_q, {dst, py_ind, val}, {host_task_events}),
```
oleksandr-pavlyk, line 1100: I think `{host_task_events}` should become `host_task_events`. The function takes vector of events in the last argument.

dpctl/tensor/libtensor/source/integer_advanced_indexing.cpp
```diff
@@ -690,9 +700,10 @@ usm_ndarray_take(dpctl::tensor::usm_ndarray src,
             sycl::free(packed_ind_offsets, ctx);
         });
     });
+    host_task_events.push_back(take_generic_ev);
 
     sycl::event host_task_ev =
-        keep_args_alive(exec_q, {src, py_ind, dst}, {take_generic_ev});
+        keep_args_alive(exec_q, {src, py_ind, dst}, {host_task_events});
```
oleksandr-pavlyk, line 706: Please replace `{host_task_events}` with `host_task_events`.

oleksandr-pavlyk:
The only working commit was 7ee92ee5145dc18b879a2ba4c3eddf1a64528cb8. 

This is the one where wait was inserted after each call to `host_task`. I am going to try this next.

PR #1093, opened 2023-03-02
Removed trailing slash in directory path

oleksandr-pavlyk:
Removed trailing slash in the directory name to avoid double slash in the absolute path of the ICD file.

Related: SAT-5617


antonwolfy (approved):
Tested locally, works fine. LGTM!

PR #1103, opened 2023-03-07
Hotfix/incorrect mask positions

oleksandr-pavlyk:
This PR fixes an issue found by @antonwolfy where the following test would not clear the assertion:

```python
import dpctl.tensor as dpt
m = dpt.full((60, 80,), True)
assert m[m].size == m.size
```

This was caused by `inclusive_scan_rec` implementation unconditionally applying non-zero indicator transformation on the input data, even for the recursive call where it was unintended. 

The fix is to add the `transformer` argument to the function, and set to it to no-op for the recursive call.

A test is added.


antonwolfy (commented):
@oleksandr-pavlyk, thank you! LGMT!

PR #1106, opened 2023-03-09
Update empty vals check in dpt.place

vlad-perevezentsev:
This PR complements #1105 .
The original behavior of `numpy.place` throws an exception when it gets  an empty array of `vals` argument
BUT if the `mask` argument contains all `False`, `numpy.place` works without exceptions ([example](https://github.com/numpy/numpy/blob/main/numpy/lib/tests/test_function_base.py#L1335)).

### DIFF
--- a/dpctl/tensor/_indexing_functions.py
+++ b/dpctl/tensor/_indexing_functions.py
@@ -293,7 +293,7 @@ def place(arr, mask, vals):
         raise dpctl.utils.ExecutionPlacementError
     if arr.shape != mask.shape or vals.ndim != 1:
         raise ValueError("Array sizes are not as required")
-    if vals.size == 0:
+    if vals.size == 0 and dpt.nonzero(mask)[0].size:
         raise ValueError("Cannot insert from an empty array!")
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_indexing_functions.py
```diff
@@ -293,7 +293,7 @@ def place(arr, mask, vals):
         raise dpctl.utils.ExecutionPlacementError
     if arr.shape != mask.shape or vals.ndim != 1:
         raise ValueError("Array sizes are not as required")
-    if vals.size == 0:
+    if vals.size == 0 and dpt.nonzero(mask)[0].size:
```
oleksandr-pavlyk, line 296: The number of set elements in the mask is computed on line 299 and assigned to variable `nz_count`. 

Perhaps it is cleaner to move the check for `vals.size == 0` till after `nz_count` value is known?

oleksandr-pavlyk (approved):
LGTM! Thank you for catching the issue and providing the fix @vlad-perevezentsev

PR #1110, opened 2023-03-12
Improve array API conformity

ndgrigorian:
Minor changes throughout dpctl to improve array API conformity.

Function signature changes to match array API
Notably introduces finfo_object subclass of np.finfo. finfo_objects return eps, max, min, and smallest_normal as Python floats.

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -472,7 +472,12 @@ def asarray(
 
 
 def empty(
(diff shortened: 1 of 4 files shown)


oleksandr-pavlyk:
Just to be explicit, `dpt.iinfo` need to no changes since it already returns struct with fields of type `int`.

oleksandr-pavlyk (approved):

PR #1118, opened 2023-03-13
Fixed build break on Debian 11 with recent OS SYCL bundle

oleksandr-pavlyk:
Renamed inserter to __appender, moved functions into concat_impl namespace

For open-source SYCL bundle based on clang 17.0 and using GCC 10.2 toolchain compilation error about ambiguous call to inserter was encountered. This was because of name clash between the definition in this file and ``std::inserter``.

Renaming helped resolve that.

Thank you @ZzEeKkAa for reporting this issue.

### DIFF
--- a/dpctl/tensor/libtensor/source/boolean_advanced_indexing.cpp
+++ b/dpctl/tensor/libtensor/source/boolean_advanced_indexing.cpp
@@ -44,36 +44,44 @@ namespace tensor
 namespace py_internal
 {
 
+namespace concat_impl
+{
+
 struct sink_t
 {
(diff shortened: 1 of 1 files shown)


ZzEeKkAa (approved):
LGTM. Thank you for fixing it. I'm not sure if the issue is directly related to GCC headers. Issue was presented on Debian bullseye (GCC 10.2) and Debian bust (GCC 8.*) but was not presented on Ubuntu 20.04 (GCC 9.*) and Ubuntu 22.04 (GCC 11.*).

Issue #1116, opened 2023-03-13
All tensor functions docstrings should provide a short description


diptorupd:
Some `dpctl.tensor` functions do not provide a short description. For example, 

arange() | -> usm_ndarray Args: start: device (optional): array API concept of device where the output array is created. device can be None, a oneAPI filter selector string, an instance of dpctl.SyclDevice corresponding to a non-partitioned SYCL device, an instance of dpctl.SyclQueue, or a Device object returned by dpctl.tensor.usm_array.device. Default: None.
-- | --

The description of arange is hard to parse and understand.

A better example is

astype() | Returns a copy of the array, cast to a specified type.
-- | --

Issue #1114, opened 2023-03-13
Improving the tensor API documentation


diptorupd:
The sphinx html generated from the docstring for tensor functions is a bit weird. The "->" in the description looks awkward and seems like a typo. Is it the intended rendering? E.g., 

broadcast_to() | -> usm_ndarray Broadcast an array to a new shape; returns the broadcasted array as a view.
-- | --

I suggest that the "->" be taken out.

diptorupd:
The issue stems from how dpctl's documentation generation script extracts the summary line from a docstring in case a signature is specified.

The following logic is the culprit:  `generate_rst.py`: https://github.com/IntelPython/dpctl/blob/8f828f24ada9829ed4d9d5dc56e6d7f39dd9ac3c/docs/generate_rst.py#L378

The function assumes that the closing parenthesis signifies the end of the signature and does not check for an `->` followed by the return type.

PR #1123, opened 2023-03-14
Closes gh-1114 and gh-1116

oleksandr-pavlyk:
Closes gh-1114 and closes gh-1116.

A pass through doc-strings of dpctl.tensor functions and classes to improve rendering of the documentation and to expand docstring.

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -128,11 +128,22 @@ def from_numpy(np_ary, device=None, usm_type="device", sycl_queue=None):
     `numpy.ndarray`.
 
     Args:
-        arg: An instance of `numpy.ndarray`
+        arg: An instance of input `numpy.ndarray`
         device: array API specification of device where the output array
-            is created.
-        sycl_queue: a :class:`dpctl.SyclQueue` used to create the output
-            array is created
+            is created
+        usm_type: The requested USM allocation type for the output array
+        sycl_queue: a :class:`dpctl.SyclQueue` instance that determines
+            output array allocation device as well as placement of data
+            movement operation. The `device` and `sycl_queue` arguments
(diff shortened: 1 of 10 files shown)


diptorupd:
@oleksandr-pavlyk Found few left over issues: `reshape`, `from_dlpack` still have the `->` in the summary.

diptorupd:
> @oleksandr-pavlyk Found few left over issues: `reshape`, `from_dlpack` still have the `->` in the summary.

It is a Sphinx issue IMO. The header of the detail section looks so much better with the `->` and it even auto links to the `usm_ndarray` doc. The summary is butchered up.

diptorupd:
@oleksandr-pavlyk Can you please check the following function in our `generate_rst.py`: https://github.com/IntelPython/dpctl/blob/8f828f24ada9829ed4d9d5dc56e6d7f39dd9ac3c/docs/generate_rst.py#L378?

Maybe we need to fix the logic here and the `->` do not need to be removed from the signature in the docstring. Having the arrows generates much better detailed sections, and the butchered summary seems to be our doing.

oleksandr-pavlyk:
> @oleksandr-pavlyk Can you please check the following function in our `generate_rst.py`:
> 
> https://github.com/IntelPython/dpctl/blob/8f828f24ada9829ed4d9d5dc56e6d7f39dd9ac3c/docs/generate_rst.py#L378
> 
> Maybe we need to fix the logic here and the `->` do not need to be removed from the signature in the docstring. Having the arrows generates much better detailed sections, and the butchered summary is in fact our doing.

I think we should be using Python's typing support in `dpctl.tensor` functions for IDE to pick up arrows and more details. 
It does not belong to the docstrings.

diptorupd (approved):

PR #1128, opened 2023-03-16
dpctl.tensor.take negative index wrapping fix

### LINKED ISSUE #1126: dpctl.tensor.take returns an invalid result
When we try to get the first element of array using a `negative index`, `dpctl.tensor.take` always `returns an usm_ndarray with 0`.
Only when `mode='wrap'`
```
#NUMPY
np_arr = np.arange(1,6)
np_ind = np.array([-5])

#DPT
dpt_arr = dpt.asarray(np_arr)
dpt_ind = dpt.asarray(np_ind)


np.take(np_arr,np_ind,mode='wrap')
>> array([1])

dpt.take(dpt_arr,dpt_ind,mode='wrap')
>> usm_ndarray([0])

# Works in other cases
np_ind = np.array([-6])
dpt_ind = dpt.asarray(np_ind)
dpt.take(dpt_arr,dpt_ind,mode='wrap')
>> usm_ndarray([5])
```

ndgrigorian:
Closes #1126 
This PR fixes incorrect logic in the WrapIndex class.
The previous logic was implemented to handle cases where modulus returns negative, but broke the implementation for indices lower than `-max_item`.
This re-implementation solves this.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/integer_advanced_indexing.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/integer_advanced_indexing.hpp
@@ -68,7 +68,8 @@ public:
     void operator()(py::ssize_t max_item, py::ssize_t &ind) const
     {
         max_item = std::max<py::ssize_t>(max_item, 1);
-        ind = (ind < 0) ? ind % max_item + max_item : ind % max_item;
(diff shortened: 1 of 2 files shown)


vlad-perevezentsev (approved):
I have tested this PR in dpnp. Tests passed.
@ndgrigorian thank you!

PR #1132, opened 2023-03-19
Changes to integer indexing modes

### LINKED ISSUE #1129: dpct.tensor.take does not disable negative indexing
When we pass negative indexes to `dpctl.take` with `clip` mod (default), it returns different array elements depending on the negative index.
This is the wrong behavior.
`clip` mode means that all indices that are too large are replaced by the index that addresses the last element along that axis.
**Note that this disables indexing with negative numbers.**

```
#NUMPY
np_arr = np.arange(1,6)
np_ind = np.array([-1,-2,-3,-4,-5])

#DPT
dpt_arr = dpt.asarray(np_arr)
dpt_ind = dpt.asarray(np_ind)


np.take(np_arr,np_ind,mode='clip')
>> array([1, 1, 1, 1, 1])

dpt.take(dpt_arr,dpt_ind)
>> usm_ndarray([5, 4, 3, 2, 1])

# work with positive over indexes
dpt_ind = dpt.asarray([5,6,7,8,9])
dpt.take(dpt_arr,dpt_ind)
>> usm_ndarray([5, 5, 5, 5, 5])
```

ndgrigorian:
Resolves #1129 
- Re-implements ClipIndex to clip indices from 0 <= i < n
- Adds FancyIndex class, which clips indices to -n <= i < n then wraps negative indices (old ClipIndex)
- Integer indexing functions have "fancy" mode (old "clip") and "clip" mode (uses new behavior)
- dpctl.tensor.take and dpctl.tensor.put now default to "fancy" mode rather than "clip" mode
- Indexing mode tests now test for new clip behavior

### DIFF
--- a/dpctl/tensor/_indexing_functions.py
+++ b/dpctl/tensor/_indexing_functions.py
@@ -26,8 +26,19 @@ import dpctl.tensor._tensor_impl as ti
 from ._copy_utils import _extract_impl, _nonzero_impl
 
 
-def take(x, indices, /, *, axis=None, mode="clip"):
-    """take(x, indices, axis=None, mode="clip")
+def _get_indexing_mode(name):
+    modes = {"fancy": 0, "clip": 1, "wrap": 2}
+    try:
+        return modes[name]
+    except KeyError:
+        raise ValueError(
+            "`mode` must be `fancy`, `clip`, or `wrap`."
+            "Got `{}`.".format(name)
+        )
+
+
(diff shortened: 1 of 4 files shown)


oleksandr-pavlyk (approved):
Thank you, @ndgrigorian ! LGTM

vlad-perevezentsev (approved):
Many thanks @ndgrigorian !

Issue #1135, opened 2023-03-22
Using of Boolean scalar for slicing usm_ndarray needs work


oleksandr-pavlyk:
Slicing in `dpctl.tensor` behaves differently than in NumPy:

```
In [15]: x = dpt.arange(4)

In [16]: x[True]
Out[16]: usm_ndarray(1)

In [17]: import numpy as n

In [18]: x_np = np.arange(5)

In [19]: x_np[True]
Out[19]: array([[0, 1, 2, 3, 4]])
```

Moreover, `x[dpt.asarray(True)]` raises `ValueError` instead of behaving like Python object.

Slicing high dimensional arrays also runs into similar problems. `x_2d[True]` behaves like `x_2d[1]` unlike in NumPy, and `x_2d[dpt.asarray(True)]` raises `ValueError`.

PR #1136, opened 2023-03-23
Slicing bug gh 1135

oleksandr-pavlyk:
Closes gh-1135.

Handle `x[True]` and `x[False]` as NumPy does, even though the behavior
may be undocumented. NumPy treats `True` as `None` (insert axis with size 1),
and treats `False` as `None` followed by empty slicing (insert axis with size 0).

Changed the logic of `_basic_slice_meta` utility function to correctly handle
boolean scalars (surprisingly, `insinstance(True, int)` evaluates to `True`).
0d arrays are handled on the same footing as the Python scalars.

Introduced `_is_integral` and `_is_boolean` utilty functions and used them in
`_basic_slice_meta` utility.

### DIFF
--- a/dpctl/tensor/_slicing.pxi
+++ b/dpctl/tensor/_slicing.pxi
@@ -1,6 +1,6 @@
 #                       Data Parallel Control (dpctl)
 #
-#  Copyright 2020-2022 Intel Corporation
+#  Copyright 2020-2023 Intel Corporation
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


antonwolfy:
The below example works differently with numpy:
```
import numpy
import dpctl.tensor as dpt

numpy.ones((3, 2))[True, False].shape
Out: (0, 3, 2)

dpt.ones((3, 2))[True, False].shape
Out: (1, 0, 3, 2)
```

oleksandr-pavlyk:
Actually this behavior in NumPy is undocumented and we should not rely on it: https://stackoverflow.com/questions/75828008/docs-about-boolean-scalars-in-indexing-of-numpy-array 

 I will change the implementation to raise `IndexError` for boolean scalar indexing.

antonwolfy:
Looks it was answered [here](https://stackoverflow.com/questions/45493270/what-does-xfalse-do-in-numpy/45493750#45493750), referred from numpy issue [#9515](https://github.com/numpy/numpy/issues/9515), and stated as expected behavior intended for 0-dimensional boolean index in numpy.

PR #1155, opened 2023-04-04
Fixed typo resulting in backwards FNC flag

ndgrigorian:
- FNC (F-not-C-contiguous) flag was returning CNF instead
- Added tests for correct behavior

### DIFF
--- a/dpctl/tensor/_flags.pyx
+++ b/dpctl/tensor/_flags.pyx
@@ -110,8 +110,8 @@ cdef class Flags:
         instance is F-contiguous and not C-contiguous.
         """
         return (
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk (approved):

oleksandr-pavlyk:
It appears that FNC, FC and FORC flags are not tested by the test suite.

ndgrigorian:
> It appears that FNC, FC and FORC flags are not tested by the test suite.

Yes, FORC isn't tested anywhere. FC and FNC are via flags.fc and flags.fnc, but the getitem method is largely ignored, which is hurting coverage.

I'll add tests where they're missing.

PR #1165, opened 2023-04-08
Simplify copy and cast kernels

oleksandr-pavlyk:
This PR works copy-and-cast functionality by simplifying underlying functors, removing specialized implementation for 2d arrays based on `CIndexer_array` as it is only very marginally faster than general strided implementation based on `CIndexer_vector`. 

This PR also adds special implementation for copy-with-casting of contiguous arrays, making type casting of contiguous arrays several times faster.


oleksandr-pavlyk:
@ndgrigorian Good catch!

ndgrigorian (approved):
LGTM

PR #1171, opened 2023-04-13
Fixes incorrect output in dpctl.tensor.where strided implementation

### LINKED ISSUE #1170: dpct.tensor.where wrong results with different shapes
If shape of `condition` parameter is not equal to one of shapes of  `x` or `y` , dpctl.tensor.where returns an incorrect result.
For example
cond_shape = (4,) , x_shape = (4,), y_shape=(4,) - correct
cond_shape = (2, 3) , x_shape = (2, 3), y_shape=(3,) - correct 
cond_shape = (4,) , x_shape = (2, 3), y_shape=(2, 3) -  incorrect

```
#NUMPY
np_cond = [False, True, True, False]
np_x = np.ones((3,4))
np_y = np.zeros((3,4))

#DPT 
dpt_cond = dpt.asarray(np_cond)
dpt_x = dpt.asarray(np_x)
dpt_y = dpt.asarray(np_y)

np.where(np_cond, np_x, np_y)
>> array([[0., 1., 1., 0.],
          [0., 1., 1., 0.],
          [0., 1., 1., 0.]])

dpt.where(dpt_cond, dpt_x, dpt_y)
>> usm_ndarray([[0., 0., 0., 1.],
                [1., 1., 1., 1.],
                [1., 0., 0., 0.]], dtype=float32)

```

ndgrigorian:
This PR implements simplify_iteration_space_4 and indexers for offsets in 4 arrays to resolve #1170.
Additionally, implements contract_iter4 and simplifies calls to std::min in strided_iters.

The issue was caused by an oversight in the mapping of elements between arrays in the strided where kernel. As the destination array was not being simplified, there was not a guarantee that the nth element of the destination corresponded to the correct element in x1 or x2.

### DIFF
--- a/dpctl/tensor/libtensor/source/simplify_iteration_space.hpp
+++ b/dpctl/tensor/libtensor/source/simplify_iteration_space.hpp
@@ -87,6 +87,39 @@ void simplify_iteration_space_3(int &,
                                 py::ssize_t &,
                                 py::ssize_t &);
 
+void simplify_iteration_space_4(int &,
+                                const py::ssize_t *&,
+                                // src1
+                                const py::ssize_t *&,
+                                py::ssize_t,
+                                bool,
+                                bool,
+                                // src2
+                                const py::ssize_t *&,
+                                py::ssize_t,
+                                bool,
+                                bool,
+                                // src3
+                                const py::ssize_t *&,
+                                py::ssize_t,
+                                bool,
+                                bool,
+                                // dst
+                                const py::ssize_t *&,
+                                py::ssize_t,
+                                bool,
+                                bool,
+                                // output
+                                std::vector<py::ssize_t> &,
+                                std::vector<py::ssize_t> &,
+                                std::vector<py::ssize_t> &,
+                                std::vector<py::ssize_t> &,
+                                std::vector<py::ssize_t> &,
+                                py::ssize_t &,
+                                py::ssize_t &,
+                                py::ssize_t &,
+                                py::ssize_t &);
+
 } // namespace py_internal
 } // namespace tensor
 } // namespace dpctl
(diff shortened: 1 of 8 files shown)


oleksandr-pavlyk:
I have added a test:

```python
def test_where_invariants():
    test_sh = (6, 8,)
    mask = dpt.asarray(np.random.choice([True, False], size=test_sh))
    p = dpt.ones(test_sh, dtype=dpt.int16)
    m = dpt.full(test_sh, -1, dtype=dpt.int16)
    inds_list = [(np.s_[:3], np.s_[::2],), (np.s_[::2], np.s_[::2],), (np.s_[::-1], np.s_[:],),]
    for ind in inds_list:
        r1 = dpt.where(mask, p, m)[ind]
        r2 = dpt.where(mask[ind], p[ind], m[ind])
        assert (dpt.asnumpy(r1) == dpt.asnumpy(r2)).all()
```

dpctl/tensor/libtensor/source/simplify_iteration_space.hpp
```diff
@@ -87,6 +87,39 @@ void simplify_iteration_space_3(int &,
                                 py::ssize_t &,
                                 py::ssize_t &);
 
+void simplify_iteration_space_4(int &,
```
oleksandr-pavlyk, line 90: One day, it would be nice to figure out how to make this nicer, perhaps by packing data about `src1`, etc. into structs.

oleksandr-pavlyk (approved):

dpctl/tensor/libtensor/source/simplify_iteration_space.hpp
oleksandr-pavlyk: And of course, this same comment applies to `simplify_iteration_space_3` and others.

PR #1172, opened 2023-04-15
Fix tests for py3.11

oleksandr-pavlyk:
This PR addresses deprecation warning reported when `dpt.arange(512*1024, 0, step=-1, dtype="i2")` is evaluated issued by NumPy 1.24.

This PR also fixes `test_sycl_queue_manager.py:: test_device_context_activates_nested_context` which fails with Python 3.11 due
to the change the exception type raised when attempting to nest contexts using object which does not implement context protocol.

### DIFF
--- a/.github/workflows/generate-coverage.yaml
+++ b/.github/workflows/generate-coverage.yaml
@@ -11,7 +11,7 @@ jobs:
 
     env:
       ONEAPI_ROOT: /opt/intel/oneapi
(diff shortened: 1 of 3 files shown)


oleksandr-pavlyk:
The second commit in this PR updates `"generate-coverage.yaml"` workflow to run on Python 3.11 and to use GTest 1.13.0

PR #1176, opened 2023-04-18
Return events for comp. tasks, rather than temp-clean up host tasks

oleksandr-pavlyk:
Return events for computational tasks, rather than temp-clean up host tasks

This resolves a hang that Anton has uncovered. The gist of the bug is

```python
for i in range(op_count):
    v[i] = dpt.empty_like(a[i], order='F', dtype=v_type)

    ht_copy_ev[i], copy_ev = ti._copy_usm_ndarray_into_usm_ndarray(src=a[i], dst=v[i], sycl_queue=a_sycl_queue)

    di._wait(copy_ev)
```

Here is `di` is a Pybind11 extensions, whose function `_wait` was implemented as

```c++
   m.def("_wait", [](sycl::event e) -> void { e.wait(); }, "");
```

The script would hang on the second iteration of the loop if `di._wait` call. The hang disappears after the change in this commit. Out attempts to build a C++ reproducer have not been successful thus far.

### DIFF
--- a/dpctl/tensor/libtensor/source/boolean_advanced_indexing.cpp
+++ b/dpctl/tensor/libtensor/source/boolean_advanced_indexing.cpp
@@ -988,8 +988,7 @@ std::pair<sycl::event, sycl::event> py_nonzero(
     sycl::event py_obj_management_host_task_ev = dpctl::utils::keep_args_alive(
         exec_q, {cumsum, indexes}, host_task_events);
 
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk:
A reproducer for a hang fixed by this PR was filed as JIRA CMPLRLLVM-46895.

ndgrigorian (approved):

PR #1188, opened 2023-04-24
Simplified API for simplify_iteration_space* functions

oleksandr-pavlyk:
This PR simplifies calling signatures of `simplify_iteration_space*` functions. Both shape pointer and input strides are const references. Passing strides as vectors obtained using `dpctl::tensor::usm_ndarray::get_strides_vector()` method allows to do away with passing contiguity flags and `itemsize` parameter.

Also changed implementation to use STL functions rather than using custom loops for operations such as copying, or scaling.
This should improve performance of stride simplification calls.


ndgrigorian (approved):
LGTM, this is a great change, thanks @oleksandr-pavlyk.

Issue #1222, opened 2023-06-01
_empty_like_pair_orderK in BinaryElementwiseFunc allocates the wrong result shape


vlad-perevezentsev:
When we pass differently shaped arrays to binary functions with the same `size` without `out` parameter, `_empty_like_pair_orderK` allocates memory for the resulting array using `return dpt.empty_like(X1, dtype=dt, usm_type=usm_type...)` for F,C contiguous and for strided arrays. This size is wrong if we use broadcasting for the resulting size so we got `ValueError: Array shapes are not the same`
For example:
```
>>> import dpctl.tensor as dpt
>>> a = dpt.ones((6,1))
>>> b = dpt.ones((1,6))
>>> dpt.multiply(a,b)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/vperevez/work/dpctl/dpctl/tensor/_elementwise_common.py", line 467, in __call__
    ht_, _ = self.binary_fn_(
ValueError: Array shapes are not the same
```

```
#print in BinaryElementwiseFunc 
print('src1', src1.shape)
print('src2', src2.shape)
print('out', out.shape)
print('res_shape', res_shape)

src1 (6, 1)
src2 (1, 6)
out (6, 1)
res_shape (6, 6)

```

PR #1224, opened 2023-06-02
Update _empty_like_pair_orderK for support arrays of different shapes

vlad-perevezentsev:
This PR solves #1222.

Now `_empty_like_pair_orderK` takes a new `res_shape` parameter to check if the shape of the resulting array has changed after broadcasting. If the resulting array shape is not equal the shape of the input arrays, `dpt.empty` will be called instead of `dpt.empty_like`. 

Also this PR updates the logic for stride input arrays in `_empty_like_pair_orderK` for cases of different stride list lengths. The previous implementation expected `st1` and `st2` to have the same stride list length and always used the `X1` array shape for the resulting array.

Added tests only for `test_add`. This is enough to check for new changes and to reduce the number of test lines. 

### DIFF
--- a/dpctl/tensor/_elementwise_common.py
+++ b/dpctl/tensor/_elementwise_common.py
@@ -433,7 +433,7 @@ class BinaryElementwiseFunc:
             if out is None:
                 if order == "K":
                     out = _empty_like_pair_orderK(
-                        src1, src2, res_dt, res_usm_type, exec_q
+                        src1, src2, res_dt, res_shape, res_usm_type, exec_q
                     )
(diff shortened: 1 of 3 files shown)


ndgrigorian (dismissed):
LGTM, thank you @vlad-perevezentsev!

oleksandr-pavlyk:
Please check the coverall report to find out why the coverage went down by that much and add tests.

ndgrigorian (approved):
The nightly build failure is unrelated to this PR, so I'm going to go ahead and approve it.

PR #1238, opened 2023-06-12
Fix dpctl.tensor.sum for support zero-dimensional array

vlad-perevezentsev:
This PR changes a bit the behavior for  `dpctl.tensor.sum` function for scalars. 
The previous implementation always returned `dpt.zeros` for zero-dimensions of the input array and worked incorrectly for scalars. According to the Python Array API: `If N is 0, the sum is 0 (i.e., the empty sum)` [[sum](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.sum.html?highlight=sum)]. 
Now if  `dpctl.tensor.sum` takes a scalar it returns a scalar has the same behavior as numpy

```
#old impl
a = dpt.asarray(1)
dpt.sum(a)
usm_ndarray(0., dtype=float32)

a = dpt.asarray(())
dpt.sum(a)
usm_ndarray(0., dtype=float32)

#new impl
a = dpt.asarray(1)
dpt.sum(a)
usm_ndarray(1., dtype=float32)

a = dpt.asarray(())
dpt.sum(a)
usm_ndarray(0., dtype=float32)

```

### DIFF
--- a/dpctl/tensor/_reduction.py
+++ b/dpctl/tensor/_reduction.py
@@ -122,10 +122,12 @@ def sum(arr, axis=None, dtype=None, keepdims=False):
         res_dt = _to_device_supported_dtype(res_dt, q.sycl_device)
 
     res_usm_type = arr.usm_type
-    if red_nd == 0:
+    if arr.size == 0:
         return dpt.zeros(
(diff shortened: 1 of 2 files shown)


ndgrigorian (approved):
LGTM!

PR #1244, opened 2023-06-14
Optimized in-place operators for rows and matrices

ndgrigorian:
This PR significant improves the performance of in-place operations between C-contiguous rows and C-contiguous matrices by implementing a variant kernel similar to that which is used for standard binary operations.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/elementwise_functions/add.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/elementwise_functions/add.hpp
@@ -480,6 +480,61 @@ struct AddInplaceStridedFactory
     }
 };
 
+template <typename argT, typename resT>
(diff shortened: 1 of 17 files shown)


vtavana (commented):
I also noticed that, for instance, in `add.hpp` we define `class add_strided_strided_kernel` on line 221 which can be changed to `class add_strided_kernel` (obviously its usage on line 239 should also be modified). It is the same for all binary functions, so you may want to modify all of them in this PR.

npolina4 (dismissed):
Thanks!

vtavana (dismissed):
Thank you @ndgrigorian

vlad-perevezentsev (approved):
LGTM!
Thank you @ndgrigorian

vtavana (approved):

PR #1259, opened 2023-06-21
skip numpy_usm_shared tests

ndgrigorian:
With this PR, all tests for the numpy_usm_shared module are now skipped pending eventual removal.

This also fixes the coverage CI, which currently fails due to changes in Numpy 1.25.0 that break tests for the numpy_usm_shared module.

### DIFF
--- a/dpctl/tests/test_dparray.py
+++ b/dpctl/tests/test_dparray.py
@@ -18,8 +18,13 @@
 """
 
 import numpy
+import pytest
 
 from dpctl.tensor import numpy_usm_shared as dparray
+
+pytestmark = pytest.mark.skip(
+    "skipping numpy_usm_shared unit tests pending submodule removal"
+)
 
 
 def get_arg():


diptorupd:
@ndgrigorian The PR in itself is fine.

However, the `numpy_usm_shared` module is no longer used by `numba_dpex` and is no longer really needed. I have discussed it with @oleksandr-pavlyk and the plan is to remove the whole module and update associated documentation, tests and issues.

ndgrigorian:
> @ndgrigorian The PR in itself is fine.
> 
> However, the `numpy_usm_shared` module is no longer used by `numba_dpex` and is no longer really needed. I have discussed it with @oleksandr-pavlyk and the plan is to remove the whole module and update associated documentation, tests and issues.

In that case, I'll update this PR to mute the dparray tests instead.
The drive behind the PR was test failures leading to the the coverage CI not passing.

vlad-perevezentsev (approved):
Yesterday I investigated the issue with `numpy_usm_shared` module in due to the new numpy 1.25. But as @diptorupd said this module is not in use now and will be removed so it LGTM!

PR #1262, opened 2023-06-23
Return default floating point type of the device in dpctl.tensor.astype when newdtype=None

vlad-perevezentsev:
This PR makes minor changes to `dpctl.tensor.astype` function by replacing the getting `target_dtype`. 
The old implementation did not consider whether the device where input array is allocated supported double-precision floating point types and returned `float64` for `None` type which raised `ValueError`. 

```
import dpctl.tensor as dpt

a = dpt.asarray([1, 1, 1], dtype='i4')
dpt.astype(a, None)

ValueError: Device Intel(R) Graphics [0x46a6] does not provide native support 
for double-precision floating point type.
```

This PR uses `_get_dtype`  function from dpctl.tensor._ctors which fixes wrong behavior. 
For `None` returns default floating point type supported by device.
For all other types returns `dpt.dtype(dtype)`

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -23,6 +23,7 @@ import dpctl.memory as dpm
 import dpctl.tensor as dpt
 import dpctl.tensor._tensor_impl as ti
 import dpctl.utils
+from dpctl.tensor._ctors import _get_dtype
 from dpctl.tensor._device import normalize_queue_device
 
 __doc__ = (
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_copy_utils.py
```diff
@@ -364,7 +365,8 @@ def astype(usm_ary, newdtype, order="K", casting="unsafe", copy=True):
         array (usm_ndarray):
             An input array.
         new_dtype (dtype):
-            The data type of the resulting array.
+            The data type of the resulting array. If None, gives default
```
oleksandr-pavlyk, line 368:
```suggestion
            The data type of the resulting array. If `None`, gives default
```

ndgrigorian (approved):

PR #1265, opened 2023-06-27
Set default dtype of usm_ndarray depending on capabilities of device 

vlad-perevezentsev:
This PR solves the issue of creating `usm_ndarray` without `dtype` argument. The default value of `dtype` in `__cinit__` of `usm_ndarray` is defined as `dtype='f8'` which raised an exception on `Iris Xe` which does not support 64-bit precision floating point type.
```
import dpctl.tensor as dpt

a = dpt.usm_ndarray(())
ValueError: Device Intel(R) Graphics [0x46a6] does not support f8 natively.

```
Idea of this PR is to change default value `dtype="f8"` to `dtype=None` and return default floating point type supported by the device when `dtype=None`.

```
import dpctl.tensor as dpt

dpt.usm_ndarray(())
usm_ndarray(0., dtype=float32)

dpt.usm_ndarray((), dtype="f8")
ValueError: Device Intel(R) Graphics [0x46a6] does not support f8 natively.
```

Note:
I'm not sure if it's correct to get device using `dpctl.SyclDevice()` when queue is unknown. 


ndgrigorian:
I think it may be more correct to use ``dpctl.select_default_device`` here rather than ``dpctl.SyclDevice``, it makes the intent more clear.

ndgrigorian (approved):
LGTM
Thanks @vlad-perevezentsev!

oleksandr-pavlyk (commented):
Thank you @vlad-perevezentsev, great addition!

PR #1266, opened 2023-06-28
Fix out of bounds integer behavior in dpctl.tensor.full

### LINKED ISSUE #1230: dpctl.tensot.full() RuntimeError: Unable to cast Python instance to C++ type
Wrong result type return on Windows for case:
```python
import numpy as np
import dpctl.tensor as dpt

a_np = np.full(3, 429)
a.dtype
#int32

a = dpt.full(3, 429)
a.dtype
#int32

b_np = np.full(3, 4294967295)
b.dtype
#int64

b = dpt.full(3, 4294967295)
RuntimeError: Unable to cast Python instance to C++ type
```

ndgrigorian:
Resolves #1230 

This PR adjusts how integer overflow and underflow is handled by ``dpctl.tensor.full``.

Previously, when an integer that would overflow the type of array in the result of ``full`` was passed in as ``fill_value``, an exception would be thrown by pybind11.

Now, ``_to_scalar`` is used to get fill value back after it has been converted to a Numpy scalar, which brings the value into the range of the type and allows pybind11 to unpack the scalar.

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -1101,6 +1101,8 @@ def full(
         fill_value = int(fill_value.real)
     elif fill_value_type is complex and np.issubdtype(dtype, np.floating):
         fill_value = fill_value.real
+    elif fill_value_type is int and np.issubdtype(dtype, np.integer):
+        fill_value = _to_scalar(fill_value, dtype)
 
     hev, _ = ti._full_usm_ndarray(fill_value, res, sycl_queue)
     hev.wait()
--- a/dpctl/tests/test_usm_ndarray_ctor.py
+++ b/dpctl/tests/test_usm_ndarray_ctor.py
@@ -1477,6 +1477,19 @@ def test_full_strides():
     assert np.array_equal(dpt.asnumpy(X), Xnp)
 
 
+def test_full_gh_1230():
+    q = get_queue_or_skip()
+    dtype = "i4"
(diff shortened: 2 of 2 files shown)


antonwolfy:
I've tried to test the build package and the behavior is still different and looks incorrect:
```
numpy.full(3, 4294967295)
Out: array([4294967295, 4294967295, 4294967295], dtype=int64)

dpt.full(3, 4294967295)
Out: usm_ndarray([-1, -1, -1], dtype=int32)
```

antonwolfy:
> > I've tried to test the built Win package and the behavior is still different and looks incorrect:
> > ```
> > numpy.full(3, 4294967295)
> > Out: array([4294967295, 4294967295, 4294967295], dtype=int64)
> > 
> > dpt.full(3, 4294967295)
> > Out: usm_ndarray([-1, -1, -1], dtype=int32)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > looks `fill_value` was just silently casted to `int32`, but expected `int64`.
> 
> Yes, this is to keep our conformity to the array API spec, which requires the output type be the default integer type.
> 
> Any possible changes to the default integer type on Windows will be made in a separate PR.

Ok, I see. Thank you @ndgrigorian! I have no more comments. LGTM!

antonwolfy (approved):

PR #1268, opened 2023-06-29
Implements negative, positive, pow, and square

ndgrigorian:
This pull request implements elementwise functions ``dpctl.tensor.negative``, ``dpctl.tensor.positive``, ``dpctl.tensor.pow``, and ``dpctl.tensor.square``.


dpctl/tensor/libtensor/include/kernels/elementwise_functions/pow.hpp
```diff
@@ -0,0 +1,269 @@
...
+        std::disjunction<tu_ns::is_complex<argT1>, tu_ns::is_complex<argT2>>>;
+    using supports_vec =
+        std::negation<std::disjunction<tu_ns::is_complex<argT1>,
+                                       tu_ns::is_complex<argT2>,
+                                       std::is_integral<argT1>,
+                                       std::is_integral<argT2>>>;
+
+    resT operator()(const argT1 &in1, const argT2 &in2)
+    {
+        return std::pow(in1, in2);
```
antonwolfy, line 65: `std::pow()` convert integer inputs to double, see [note](https://en.cppreference.com/w/cpp/numeric/math/pow):
```
Otherwise, if num1 and/or num2 has type double or an integer type, then std::pow(num1, num2) has the same effect as std::pow(static_cast<double>(num1), static_cast<double>(num2)).
```
and as result it will cause a crash on Iris Xe, since no fp64 support there.
We faced the same in dpnp and were trying to resolve by PR [#1463](https://github.com/IntelPython/dpnp/pull/1463).

ndgrigorian: Thank you for pointing this out.

I'll add Vlad's ipow implementation before converting this to ready for review.

dpctl/tensor/libtensor/include/kernels/elementwise_functions/pow.hpp
```diff
@@ -0,0 +1,292 @@
...
+                                       std::is_integral<argT1>,
+                                       std::is_integral<argT2>>>;
+
+    resT operator()(argT1 in1, argT2 in2)
+    {
+        if constexpr (std::is_integral_v<argT1> || std::is_integral_v<argT2>) {
+            if constexpr (std::is_signed_v<argT2>) {
+                if (in2 < 0) {
+                    // invalid; return 0
+                    return resT(0);
```
vlad-perevezentsev, line 70: I think it would be better if you raised an error here as implemented in numpy
```
import numpy

numpy.power(numpy.array([2]),-2)
ValueError: Integers to negative integer powers are not allowed.
```

ndgrigorian: Raising an error inside of a kernel is not something we've been doing inside of a kernel (parallel, some memory might already be changed, etc.)

We did not raise inside of kernels when implementing integer advanced indexing for this reason, so I think it should be left this way.

vlad-perevezentsev (approved):
LGTM!
thanks @ndgrigorian

PR #1271, opened 2023-06-30
dpctl.tensor.floor_divide fixed for signed 0 output

ndgrigorian:
On some devices, ``sycl::floor`` and ``std::floor`` would drop the sign of 0.

This PR resolves those cases and adds a test.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/elementwise_functions/floor_divide.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/elementwise_functions/floor_divide.hpp
@@ -60,24 +60,25 @@ struct FloorDivideFunctor
 
     resT operator()(const argT1 &in1, const argT2 &in2)
     {
-        auto tmp = in1 / in2;
-        if constexpr (std::is_integral_v<decltype(tmp)>) {
-            if constexpr (std::is_unsigned_v<decltype(tmp)>) {
-                return (in2 == argT2(0)) ? resT(0) : tmp;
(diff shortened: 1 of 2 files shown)


dpctl/tensor/libtensor/include/kernels/elementwise_functions/floor_divide.hpp
```diff
@@ -52,62 +52,60 @@ namespace tu_ns = dpctl::tensor::type_utils;
...
-                                     ((rem_tmp[i] < 0) != (in2[i] < 0)));
-                        tmp[i] -= corr;
+            for (int i = 0; i < vec_sz; ++i) {
+                if (in2[i] == argT2(0)) {
+                    res[i] = resT(0);
+                }
+                else {
+                    res[i] = in1[i] / in2[i];
+                    if constexpr (std::is_signed_v<resT>) {
+                        auto mod = in1[i] % in2[i];
```
oleksandr-pavlyk, line 94: This performs a second division. How about `mod = in1[i] - res[i] * in2[i]` instead?

ndgrigorian: I will try this out as well, I'd like to see how the performance looks.

I have also seen sources suggest that the remainder is, in some cases, a byproduct of division and the compiler can optimize these operations when nearby.

i.e., [here](https://en.cppreference.com/w/cpp/numeric/math/div) under notes

oleksandr-pavlyk (approved):
LGTM @ndgrigorian. Thank you

PR #1272, opened 2023-07-04
Implements logaddexp and hypot

ndgrigorian:
This PR implements logaddexp and hypot.

To implement these functions, an issue with type promotion for binary functions has been addressed: 

``logaddexp`` and ``hypot`` both act like compositions of unary functions with binary functions and type promotion logic for binary functions has assumed pure binary functions, resulting in behaviors such as:

```

In [2]: x = dpt.ones(1, dtype="i2")

In [3]: dpt.log(dpt.add(dpt.exp(x), dpt.exp(x)))
Out[3]: usm_ndarray([1.6931472], dtype=float32)

In [4]: np.logaddexp(dpt.asnumpy(x), dpt.asnumpy(x))
Out[4]: array([1.6931472], dtype=float32)

In [5]: dpt.logaddexp(x, x)
Out[5]: usm_ndarray([1.6931472])
```
where the dtype of the last array is ``float64``.


oleksandr-pavlyk:
@ndgrigorian The issue with type promotion has already been addressed in one of my commits to this branch.

dpctl/tensor/libtensor/include/kernels/elementwise_functions/logaddexp.hpp
```diff
@@ -0,0 +1,224 @@
...
+    using supports_sg_loadstore = typename std::negation<
+        std::disjunction<tu_ns::is_complex<argT1>, tu_ns::is_complex<argT2>>>;
+    using supports_vec = typename std::negation<
+        std::disjunction<tu_ns::is_complex<argT1>, tu_ns::is_complex<argT2>>>;
+
+    resT operator()(const argT1 &in1, const argT2 &in2)
+    {
+        resT max = std::max<resT>(in1, in2);
+        resT min = std::min<resT>(in1, in2);
+        return max + std::log(resT(1) + std::exp(min - max));
```
oleksandr-pavlyk, line 67: Perhaps use `std::log1p` instead.
```suggestion
        return max + std::log1p(std::exp(min - max));
```

dpctl/tensor/libtensor/include/kernels/elementwise_functions/logaddexp.hpp
```diff
@@ -0,0 +1,224 @@
...
+    sycl::vec<resT, vec_sz> operator()(const sycl::vec<argT1, vec_sz> &in1,
+                                       const sycl::vec<argT2, vec_sz> &in2)
+    {
+        sycl::vec<resT, vec_sz> res;
+        auto diff = in1 - in2;
+
+#pragma unroll
+        for (int i = 0; i < vec_sz; ++i) {
+            resT max = std::max<resT>(in1[i], in2[i]);
+            res[i] = max + std::log(resT(1) + std::exp(std::abs(diff[i])));
```
oleksandr-pavlyk, line 80:
```suggestion
            res[i] = max + std::log1p(std::exp(std::abs(diff[i])));
```

oleksandr-pavlyk (approved):
I think this PR is ready to go in. Thank you @ndgrigorian

PR #1281, opened 2023-07-13
Created a temporary copy in case of overlap for unary function

antonwolfy:
The PR proposes to allocate a temporary buffer rather than to raise an exception in case when the memory overlapping is detected between `in` and `out` arrays in a call of unary function.

The changed is intended to have the below code example as a valid:
```
import dpctl.tensor as dpt

a = dpt.arange(10, dtype='f4')
_ = dpt.sqrt(a, out=a)
```

The approach with temporary buffer was chosen as a start point while integrating support of the use case, since it is easiest way to implement this.
The next step might be to add separate kernels to handle in-place unary operations (when both `in` and `out` arrays point to the same memory) if the performance gain would be sensible.

### DIFF
--- a/dpctl/tests/elementwise/test_square.py
+++ b/dpctl/tests/elementwise/test_square.py
@@ -97,3 +97,29 @@ def test_square_special_cases(dtype):
             rtol=tol,
             equal_nan=True,
         )
+
+
+@pytest.mark.parametrize("dtype", ["f2", "f4", "f8", "c8", "c16"])
+def test_square_out_overlap(dtype):
+    q = get_queue_or_skip()
+    skip_if_dtype_not_supported(dtype, q)
+
+    X = dpt.linspace(0, 35, 60, dtype=dtype, sycl_queue=q)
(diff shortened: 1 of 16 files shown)


dpctl/tests/elementwise/test_square.py
```diff
@@ -97,3 +97,25 @@ def test_square_special_cases(dtype):
...
+
+    Y = dpt.sqrt(X, out=X)
+    assert Y is X
+    assert np.allclose(dpt.asnumpy(X), Xnp)
+
+    Ynp = np.sqrt(Xnp, out=Xnp[::-1])
+    Y = dpt.sqrt(X, out=X[::-1])
+    assert Y is not X
+    assert np.allclose(dpt.asnumpy(X), Xnp)
+    assert np.allclose(dpt.asnumpy(Y), Ynp)
```
oleksandr-pavlyk, line 121: I think this test should either use `dpt.square` or be removed from this file.

npolina4 (approved):

oleksandr-pavlyk (approved):

PR #1252, opened 2023-06-15
Update workaround to Klocwork

vlad-perevezentsev:
This PR updates the previous workaround for Klocwork.

- [X] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to an issue with a reproducer?
- [ ] Have you tested your changes locally for CPU and GPU devices?
- [ ] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you opening the PR as a draft?

PR #1289, opened 2023-07-18
Revert "Update w/a to Klocwork (#1252)"

oleksandr-pavlyk:
This reverts commit f15ec81912b0e9a4f97bbf6d18ad18e9f291032e.

This workaround, if needed, should only have been applied conditionally. 
Anyhow, we transition to use Coverity now, so this workaround is no longer necessary.


conda-recipe/build.sh
```diff
@@ -4,7 +4,6 @@
 # by DPC++ compiler conda packages. Will need to be added to DPC++ compiler
 # activation scripts.
 export LDFLAGS="$LDFLAGS -Wl,-rpath,$PREFIX/lib"
```
antonwolfy, line 6: Isn't that also relating to the W/A and should be removed? At least the above comment claims regarding that.

oleksandr-pavlyk:
Thank you for catching this @antonwolfy .  I have removed setting of the `LDFLAGS` and the comment.

vlad-perevezentsev (approved):
thanks @oleksandr-pavlyk

Issue #1293, opened 2023-07-20
dpt.sum with "keepsdims" returns incorrect shape in case one or more axis is equal to 0


antonwolfy:
The problem comes from [gh-1487](https://github.com/IntelPython/dpnp/issues/1487):

```python
import dpctl.tensor as dpt
import numpy

a = dpt.ones((0, 10))
dpt.sum(a, keepdims=True).shape 
# Out: ()

b = numpy.ones((0, 10))
numpy.sum(a, keepdims=True).shape 
# Out: (1,)
```

PR #1294, opened 2023-07-20
Fix gh-1293 for sum over zero-size array with keepdims set

oleksandr-pavlyk:
Closes gh-1293

This PR adds handling of `keepdims=True` case for special-case early exit for zero-size input.

### DIFF
--- a/dpctl/tensor/_reduction.py
+++ b/dpctl/tensor/_reduction.py
@@ -123,6 +123,10 @@ def sum(arr, axis=None, dtype=None, keepdims=False):
 
     res_usm_type = arr.usm_type
     if arr.size == 0:
+        if keepdims:
+            res_shape = res_shape + (1,) * red_nd
+            inv_perm = sorted(range(nd), key=lambda d: perm[d])
+            res_shape = tuple(res_shape[i] for i in inv_perm)
         return dpt.zeros(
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_reduction.py
```diff
@@ -123,6 +123,10 @@ def sum(arr, axis=None, dtype=None, keepdims=False):
 
     res_usm_type = arr.usm_type
     if arr.size == 0:
+        if keepdims:
+            res_shape = res_shape + (1,) * red_nd
+            inv_perm = sorted(range(nd), key=lambda d: perm[d])
```
antonwolfy, line 128: shouldn't we move that above the `if` to reduce code duplication, since it looks common and without any dependency on logic below?

antonwolfy (approved):
I've tested with `dpnp` from [#1488](https://github.com/IntelPython/dpnp/pull/1488) and all tests are passing now. Thank you @oleksandr-pavlyk !

dpctl/tensor/_reduction.py
oleksandr-pavlyk: Perhaps in a separate PR. We are talking about deduplicating two lines of code at the expense of making the code more complex, since unexpanded `res_shape` is needed in the call to `sum_over_axis`. 

Thank you for the review!

PR #1300, opened 2023-07-24
Boolean indexing performance

oleksandr-pavlyk:
`dpctl` used to utilize int64 data type temporary to compute running sum of flattened mask array. For smaller arrays, where int32 would suffice, this resulted in suboptimal performance. 

This PR adds support for int32 in advanced Boolean indexing implementation function, and changes orchestrating Pyton code to allocated temporary of int32 data type if the number of elements in the mask array is less than a threshold.

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -31,6 +31,8 @@ __doc__ = (
     ":class:`dpctl.tensor.usm_ndarray`."
 )
 
+int32_t_max = 2147483648
+
 
 def _copy_to_numpy(ary):
     if not isinstance(ary, dpt.usm_ndarray):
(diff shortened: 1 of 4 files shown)


oleksandr-pavlyk:
For
 
```
import dpctl.tensor as dpt
mask = dpt.ones((4096, 4096), dtype="?")
x = dpt.ones(mask.shape, dtype="f4")
%time y = x[mask]
```

The timing dropped from around 60ms to around 40ms on Iris Xe.

oleksandr-pavlyk:
Irrespective of the absolute gains in performance this change is good, since it reduces memory traffic.

It is likely that the tests in the test suite only exercise int32 temporaries. I can not even create array with size 2**32 on my laptop, but we should perhaps try this on a server-grade GPU, such as IDC beta.  Regardless, it may affect C++ code coverage numbers once we have that going.

ndgrigorian (approved):

PR #1311, opened 2023-07-28
MAINT: changes to test_log.py to avoid RuntimeErrors

oleksandr-pavlyk:
Changed tests to avoid `RuntimeErrors` issued by NumPy for numerical exceptions

Replaced feeding zeros into `dpt.log` where it is not essentially for the property tested. 

When testing special values, hard code expected results, rather than recompute them with `np.log`.

### DIFF
--- a/dpctl/tests/elementwise/test_log.py
+++ b/dpctl/tests/elementwise/test_log.py
@@ -31,8 +31,8 @@ def test_log_out_type(dtype):
     q = get_queue_or_skip()
     skip_if_dtype_not_supported(dtype, q)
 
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
@antonwolfy For some reason using fixture `suppress_invalid_numpy_warnings` did not work like it does in "test_sqrt.py".

Perhaps "test_sqrt.py" could use similar type of changes.

ndgrigorian (approved):
This is a good change.

Similar changes should be propagated to other tests with warnings.

PR #1315, opened 2023-07-31
Fix missing string format in isdtype

fcharras:
Fix a small mistake in `dpt.isdtype` that prevent wrong user input to be correctly displayed in the error message:

> TypeError: Expected instance of `dpt.dtype`, got {dtype_}


oleksandr-pavlyk (approved):
Thank you @fcharras, appreciate your contribution!

PR #1317, opened 2023-07-31
Implement tensor.atan2 and tensor.signbit

oleksandr-pavlyk:
This PR implements `dpctl.tensor.atan2` per array API [specs](https://data-apis.org/array-api/latest/API_specification/generated/array_api.atan2.html). 

Since special values stipulations require us to distinguish between `-0.0` and `+0.0` elements, which `dpctl.tensor.equal` considers equal, this PR also implements `dpctl.tensor.signbit` used to in tests of `dpctl.tensor.atan2`.


vtavana (approved):
LGTM! Thank you!

ndgrigorian (approved):

Issue #1322, opened 2023-08-02
dpt.nonzero returns array of int32 dtype


antonwolfy:
After merging #1300, the result array type of `dpt.nonzero` is changed:
```python
a = dpt.ones(10, dtype="i8")
dpt.nonzero(a)[0].dtype
# Out: dtype('int32')
```
The result dtype depends now on array size and may vary between `int32` and `int64` on the same platform:
```python
sz = numpy.iinfo("i4").max
a = dpt.ones(sz, dtype="i8")
dpt.nonzero(a)[0].dtype
# Out: dtype('int32')

sz = numpy.iinfo("i4").max + 1
a = dpt.ones(sz, dtype="i8")
dpt.nonzero(a)[0].dtype
# Out: dtype('int64')
```

The python array API states [that](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.nonzero.html?highlight=nonzero):
> The returned array must have the default array index data type.

and [here](https://data-apis.org/array-api/2022.12/API_specification/data_types.html#default-data-types) is about the default array index type:
> a default array index data type (either int32 or int64).
...
The default array index data type may be int32 on 32-bit platforms, but the default should be int64 otherwise.

PR #1323, opened 2023-08-02
Nonzero regression

oleksandr-pavlyk:
Fixes regression in output dtype of `tensor.nonzero` and closes gh-1322.

Changes in gh-1300 made data type of indexes arrays was the same as dtype of temporary cumulative sum array, which incidentally worked before the change. 


antonwolfy (commented):
I've checked with full dpnp tests scope and it's successfully passing now. Thank you @oleksandr-pavlyk

PR #1328, opened 2023-08-07
Fix some array api test cases

ndgrigorian:
This PR fixes an array API test case for ``logaddexp`` that fails for mixed nan and number operands.

It also applies a previous fix used for ``astype`` with ``order="K"`` to ``copy``.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/elementwise_functions/logaddexp.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/elementwise_functions/logaddexp.hpp
@@ -28,6 +28,7 @@
 #include <CL/sycl.hpp>
 #include <cstddef>
 #include <cstdint>
+#include <limits>
 #include <type_traits>
 
 #include "utils/offset_utils.hpp"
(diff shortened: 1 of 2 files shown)


dpctl/tensor/libtensor/include/kernels/elementwise_functions/logaddexp.hpp
```diff
@@ -55,14 +56,15 @@ using dpctl::tensor::type_utils::vec_cast;
...
-        std::disjunction<tu_ns::is_complex<argT1>, tu_ns::is_complex<argT2>>>;
+    using supports_sg_loadstore = std::true_type;
+    using supports_vec = std::true_type;
 
     resT operator()(const argT1 &in1, const argT2 &in2)
     {
         resT max = std::max<resT>(in1, in2);
+        if (std::isnan(max) || std::isinf(max)) {
+            return max;
+        }
```
oleksandr-pavlyk, line 67:
```suggestion
        if (std::isnan(max)) {
            return std::numeric_limits<resT>::quiet_NaN();
        } else if (std::isinf(max)) {
            return max;
        }
```

oleksandr-pavlyk:
I have pushed a change to factor scalar implementation out and fixed it to ensure `logaddexp(-inf,-inf)=-ing`.

PR #1329, opened 2023-08-08
Fix boolean reductions for #1327

### LINKED ISSUE #1327: Incorrect behavior of `dpctl.tensor.all` and  `dpctl.tensor.any` with `axis=0` parameter
The `dpctl.tensor.all` function returns an incorrect result with the parameter `axis=0` and `dimension >=3`  of the input ndarray.

Running on GPU

```
import dpctl.tensor as dpt
import numpy

a = numpy.arange(24).reshape(2, 3, 4) - 10

a.dtype
>dtype('int64')

dpt_a = dpt.asarray(a)

numpy.all(a,axis=0)
>array([[ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True, False,  True]])

dpt.all(dpt_a,axis=0)
>usm_ndarray([[ True, False, False, False],
              [False, False, False, False],
              [False, False, False, False]])

# at the same for dtype=int32

# floating types and boolean work correctly 

a = a.astype('float32')
dpt_a = dpt.asarray(a)

numpy.all(a,axis=0)
>array([[ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True, False,  True]])

dpt.all(dpt_a,axis=0)
>usm_ndarray([[ True,  True,  True,  True],
              [ True,  True,  True,  True],
              [ True,  True,  True,  True]])

```

Running on CPU for integer and floating types throws ` Segmentation fault`
for boolean type incorrect result 

```
import dpctl.tensor as dpt
import numpy

a = numpy.arange(24).reshape(2, 3, 4) - 10
dpt_a = dpt.asarray(a)
dpt.all(dpt_a,axis=0)
> Segmentation fault

dpt_a = dpt.astype(dpt_a, 'float32')
dpt.all(dpt_a,axis=0)
> Segmentation fault

dpt_a = dpt.astype(dpt_a, 'bool')
> usm_ndarray([[ True,  True,  True,  True],
              [ True,  True,  True,  True],
              [False, False,  True, False]])


# for ndim<3 works

a = numpy.arange(6).reshape(2, 3) - 4
dpt_a = dpt.asarray(a)
dpt.all(dpt_a,axis=0)
> usm_ndarray([ True, False,  True])

```

The `dpctl.tensor.any` function does not work on CPU (`Segmentation fault`) with the parameter `axis=0` and `dimension >=3`  of the input ndarray for all dtypes. 
```
import dpctl.tensor as dpt

dpt_a = dpt.ones((2, 3, 4))
dpt.any(dpt_a,axis=0)
Segmentation fault (core dumped)

```

ndgrigorian:
Closes #1327 

This PR fixes a typo in the Python binding common to both boolean reductions (any, all).

The binding was calling the implementations of these functions with ``dst_nd`` rather than ``iter_nd``.

``dst_nd`` and ``iter_nd`` often ended up being the same for small-dimensional arrays, which masked the problem.

### DIFF
--- a/dpctl/tensor/libtensor/source/boolean_reductions.hpp
+++ b/dpctl/tensor/libtensor/source/boolean_reductions.hpp
@@ -235,7 +235,8 @@ py_boolean_reduction(dpctl::tensor::usm_ndarray src,
         std::get<2>(iter_red_metadata_packing_triple_);
 
     py::ssize_t *iter_shape_and_strides = packed_shapes_and_strides;
-    py::ssize_t *red_shape_stride = packed_shapes_and_strides + (3 * iter_nd);
+    py::ssize_t *red_shape_stride =
+        packed_shapes_and_strides + 3 * simplified_iter_shape.size();
 
     std::vector<sycl::event> all_deps;
     all_deps.reserve(depends.size() + 1);
@@ -244,7 +245,7 @@ py_boolean_reduction(dpctl::tensor::usm_ndarray src,
     all_deps.push_back(copy_metadata_ev);
 
     auto red_ev =
-        fn(exec_q, dst_nelems, red_nelems, src_data, dst_data, dst_nd,
+        fn(exec_q, dst_nelems, red_nelems, src_data, dst_data, iter_nd,
            iter_shape_and_strides, iter_src_offset, iter_dst_offset,
            simplified_red_nd, red_shape_stride, red_src_offset, all_deps);
 
--- a/dpctl/tests/test_usm_ndarray_utility_functions.py
+++ b/dpctl/tests/test_usm_ndarray_utility_functions.py
@@ -148,3 +148,20 @@ def test_arg_validation_boolean_reductions(func):
         func(d)
     with pytest.raises(AxisError):
         func(x, axis=-3)
+
+
+def test_boolean_reductions_3d_gh_1327():
+    get_queue_or_skip()
+
+    size = 24
+    x = dpt.reshape(dpt.arange(-10, size - 10, 1, dtype="i4"), (2, 3, 4))
+    res = dpt.all(x, axis=0)
+    res_np = np.full(res.shape, True, dtype="?")
+    res_np[2, 2] = False
+
+    assert (dpt.asnumpy(res) == res_np).all()
+
+    x = dpt.ones((2, 3, 4, 5), dtype="i4")
+    res = dpt.any(x, axis=0)
+
+    assert (dpt.asnumpy(res) == np.full(res.shape, True, dtype="?")).all()


dpctl/tensor/libtensor/source/boolean_reductions.hpp
```diff
@@ -244,7 +245,7 @@ py_boolean_reduction(dpctl::tensor::usm_ndarray src,
     all_deps.push_back(copy_metadata_ev);
 
     auto red_ev =
-        fn(exec_q, dst_nelems, red_nelems, src_data, dst_data, dst_nd,
+        fn(exec_q, dst_nelems, red_nelems, src_data, dst_data, iter_nd,
```
oleksandr-pavlyk, line 248: If `iter_nd` is not the same as `simplified_iter_shape.size()`, perhaps `iter_nd` must be assigned this value. I would think this parameter is what tells the code how many elements of `iter_shape_and_strides` it can access, and changes on lines 238-239 seem to imply it is `3 * simplifier_iter_shape.size()`, not `3 * iter_nd`.

oleksandr-pavlyk: I tried applying change 

```diff
diff --git a/dpctl/tensor/libtensor/source/boolean_reductions.hpp b/dpctl/tensor/libtensor/source/boolean_reductions.hpp
index e6d14c8d4..413816f74 100644
--- a/dpctl/tensor/libtensor/source/boolean_reductions.hpp
+++ b/dpctl/tensor/libtensor/source/boolean_reductions.hpp
@@ -235,8 +235,9 @@ py_boolean_reduction(dpctl::tensor::usm_ndarray src,
         std::get<2>(iter_red_metadata_packing_triple_);

     py::ssize_t *iter_shape_and_strides = packed_shapes_and_strides;
+    iter_nd = static_cast<int>(simplified_iter_shape.size());
     py::ssize_t *red_shape_stride =
-        packed_shapes_and_strides + 3 * simplified_iter_shape.size();
+        packed_shapes_and_strides + 3 * iter_nd;

     std::vector<sycl::event> all_deps;
     all_deps.reserve(depends.size() + 1);
``` 

to this PR. Tests are passing too:

```python
import numpy as np, dpctl.tensor as dpt
def check(x_dpt, x_np, axis):
    return np.array_equal(dpt.asnumpy(dpt.all(x_dpt, axis=axis)), np.all(x_np, axis=axis))
def repeat(n):
    res = []
    axes_list = [0, 1, 2, 3, (0,1), (0,2), (0,3), (1,2), (1,3), (0,1,2), (0,1,3), (0,2,3), (1,2,3), (0,1,2,3)]
    for i in range(n):
        x_np = np.random.choice([True, False], size=120).reshape((2,3,4,5))
        x = dpt.asarray(x_np)
        res.append(all([check(x, x_np, a) for a in axes_list]))
    return all(res)

repeat(1000)
# outputs True
```

ndgrigorian: ``simplified_iter_shape.size()`` and ``iter_nd`` should be the same, as ``iter_nd`` is what is passed to ``simplify_iteration_space``.

The change to the packing size was only to align structurally with the ``sum`` reduction as much as possible. I can revert it to ``3 * iter_nd``.

oleksandr-pavlyk: Yes, you are right. The `simplified_iteration_space` modifies the first argument by reference.  Then please disregard my comment. LGTM!

PR #1332, opened 2023-08-09
Fix expm1 for complex special cases

ndgrigorian:
This PR changes the logic of ``expm1`` for complex data types to better account for special cases (``NaN``s, ``inf``s, etc.). These cases were handled identically to Numpy, but did not conform to the array API.

In addition, this PR slips in changes to silence the remaining warnings in the test suite which were being raised by the elementwise tests.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/elementwise_functions/expm1.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/elementwise_functions/expm1.hpp
@@ -85,7 +85,8 @@ template <typename argT, typename resT> struct Expm1Functor
                         return in;
                     }
                     else {
-                        return (x * resT{std::cos(y), std::sin(y)});
+                        return (resT{std::copysign(x, std::cos(y)),
+                                     std::copysign(x, std::sin(y))});
                     }
                 }
                 else {
--- a/dpctl/tensor/libtensor/include/kernels/elementwise_functions/expm1.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/elementwise_functions/expm1.hpp
@@ -85,19 +85,19 @@ template <typename argT, typename resT> struct Expm1Functor
                         return in;
                     }
                     else {
-                        return (std::numeric_limits<realT>::infinity() *
-                                    resT{std::cos(y), std::sin(y)} -
-                                realT(1));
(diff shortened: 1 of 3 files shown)


dpctl/tensor/libtensor/include/kernels/elementwise_functions/expm1.hpp
```diff
@@ -73,6 +74,45 @@ template <typename argT, typename resT> struct Expm1Functor
...
+                    }
+                }
+                else {
+                    // negative infinity cases
+                    if (!std::isfinite(y)) {
+                        return resT{-1, 0};
+                    }
+                    else {
+                        return (realT(0) * resT{std::cos(y), std::sin(y)} -
+                                realT(1));
```
oleksandr-pavlyk, line 100: Perhaps it is better to do:

```suggestion
                        return resT{ realT(-1), std::copysign(realT(0), std::sin(y))};
```

dpctl/tensor/libtensor/include/kernels/elementwise_functions/expm1.hpp
```diff
@@ -73,6 +74,45 @@ template <typename argT, typename resT> struct Expm1Functor
...
+                if (x > realT(0)) {
+                    // positive infinity cases
+                    if (!std::isfinite(y)) {
+                        return resT{x, std::numeric_limits<realT>::quiet_NaN()};
+                    }
+                    else if (y == realT(0)) {
+                        return in;
+                    }
+                    else {
+                        return (x * resT{std::cos(y), std::sin(y)});
```
oleksandr-pavlyk, line 88: I think we should do instead, to ensure that `NaN` does not come up in product computations:

```suggestion
                        return resT{std::copysign(x, std::cos(y)), std::copysign(x, std::sin(y))});
```

ndgrigorian: Yes, this makes sense. I'll make this change.

oleksandr-pavlyk: Other than this change, the PR looks good. I have built and tested it on Iris Xe and PVC.

oleksandr-pavlyk (approved):

PR #1338, opened 2023-08-11
Fixed tests from_dlpack and _from_dlpack_strided for USM-host allocations

oleksandr-pavlyk:
Since USM-host allocation, unlike USM-device and USM-shared allocations, is not associated with any particular device, only with the host, it is unreasonable to expect that ``sycl::get_pointer_device(ptr, ctx)`` will return the same device that is stored in the `sycl_queue` attribute of `dpctl.tensor.usm_ndarray` instance.

The tests are adjusted to bypass that check for USM-host allocations.

### DIFF
--- a/dpctl/tests/test_usm_ndarray_dlpack.py
+++ b/dpctl/tests/test_usm_ndarray_dlpack.py
@@ -127,9 +127,11 @@ def test_from_dlpack(shape, typestr, usm_type):
         Y = dpt.from_dlpack(X)
         assert X.shape == Y.shape
         assert X.dtype == Y.dtype
-        assert X.sycl_device == Y.sycl_device
         assert X.usm_type == Y.usm_type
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
SYCL spec for `sycl::malloc_host` can be found [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_host_allocation_functions). 

The `sycl::get_pointer_device(const void *ptr, const sycl::context &syclContext)` is [explicitly documented](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_unified_shared_memory_pointer_queries) so: "If `ptr` points within a host USM allocation for the context `syclContext`, returns the first device in `syclContext`."

ndgrigorian (approved):

PR #1340, opened 2023-08-13
dpctl.tensor.abs fixed for signed zeros

ndgrigorian:
This PR fixes the behavior of ``abs`` for signed zeros.

Previous behavior:
```

In [2]: dpt.abs(dpt.asarray(-0.))
Out[2]: usm_ndarray(-0.)
```

Now:
```

In [2]: dpt.abs(dpt.asarray(-0.))
Out[2]: usm_ndarray(0.)
```


ndgrigorian:
Superseded by #1343

PR #1341, opened 2023-08-14
Shift tuple support when axis=None for dpctl.tensor.roll

vlad-perevezentsev:
This PR suggests adding support for the case when `shift is tuple` and `axis=None` for `dpctl.tensor.roll` .
The previous implementation raised `TypeError` exception cause `copy_usm_ndarray_for_reshape` expects shift as an integer.
```
import dpctl.tensor as dpt

a = dpt.ones((5,2))
dpt.roll(a,shift=(1,2))

TypeError: _copy_usm_ndarray_for_reshape(): incompatible function arguments. The following argument types are supported:
    1. (src: dpctl::tensor::usm_ndarray, dst: dpctl::tensor::usm_ndarray, shift: int, sycl_queue: dpctl.SyclQueue, depends: List[dpctl.SyclEvent] = []) -> Tuple[dpctl.SyclEvent, dpctl.SyclEvent]

```

Additionally, the PR adds a check for the equality of the `source` and `destination`  parameters to return the input array `X` without having to move its axes in `dpctl.tensor.moveaxis`.

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -885,6 +885,9 @@ def moveaxis(X, source, destination):
             "the same number of elements"
         )
 
+    if source == destination:
+        return X
+
     ind = [n for n in range(X.ndim) if n not in source]
 
     for src, dst in sorted(zip(destination, source)):
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -426,6 +426,9 @@ def roll(X, shift, axis=None):
     if not isinstance(X, dpt.usm_ndarray):
         raise TypeError(f"Expected usm_ndarray type, got {type(X)}.")
     if axis is None:
+        # get the combined shift value for all axes
+        if type(shift) is tuple:
+            shift = sum(shift)
         res = dpt.empty(
             X.shape, dtype=X.dtype, usm_type=X.usm_type, sycl_queue=X.sycl_queue
         )
--- a/dpctl/tests/test_usm_ndarray_manipulation.py
+++ b/dpctl/tests/test_usm_ndarray_manipulation.py
@@ -590,6 +590,8 @@ def test_roll_empty():
     "data",
     [
         [2, None],
+        [(0, 1), None],
+        [(-1, 0), None],
         [-2, None],
         [2, 0],
         [-2, 0],
@@ -617,6 +619,8 @@ def test_roll_1d(data):
     "data",
     [
         [1, None],
+        [(2, 1), None],
+        [(-1, 2), None],
         [1, 0],
         [1, 1],
         [1, ()],


oleksandr-pavlyk:
Array API spec for `xp.roll` [stipulates](https://data-apis.org/array-api/latest/API_specification/generated/array_api.roll.html) that

>  If `shift` is a tuple, then `axis` **must** be a tuple of the same size, and each of the given `axes` must be shifted by the corresponding element in `shift`.

I realize that implemented behavior aligns with how NumPy currently behaves, but `dpnp.roll` could implement that convenience at its end.

dpctl/tensor/_manipulation_functions.py
```diff
@@ -426,6 +426,9 @@ def roll(X, shift, axis=None):
     if not isinstance(X, dpt.usm_ndarray):
         raise TypeError(f"Expected usm_ndarray type, got {type(X)}.")
     if axis is None:
+        # get the combined shift value for all axes
+        if type(shift) is tuple:
+            shift = sum(shift)
```
oleksandr-pavlyk, line 431: I am afraid this is the right implementation. Providing correct implementation requires writing a new kernel.

antonwolfy:
> Array API spec for `xp.roll` [stipulates](https://data-apis.org/array-api/latest/API_specification/generated/array_api.roll.html) that
> 
> > If `shift` is a tuple, then `axis` **must** be a tuple of the same size, and each of the given `axes` must be shifted by the corresponding element in `shift`.
> 
> I realize that implemented behavior aligns with how NumPy currently behaves, but `dpnp.roll` could implement that convenience at its end.

Then dpctl has to properly handle the use case and to raise a proper exception, since currently it tries to copy and fails:
```python
a = dpt.ones(10)

dpt.roll(a, (2, 1, 3), axis=None)
Cell In[5], line 1
----> 1 dpt.roll(a, (2, 1, 3), axis=None)

File ~/miniconda3/envs/dpnp_dev/lib/python3.9/site-packages/dpctl/tensor/_manipulation_functions.py:432, in roll(X, shift, axis)
    428 if axis is None:
    429     res = dpt.empty(
    430         X.shape, dtype=X.dtype, usm_type=X.usm_type, sycl_queue=X.sycl_queue
    431     )
--> 432     hev, _ = ti._copy_usm_ndarray_for_reshape(
    433         src=X, dst=res, shift=shift, sycl_queue=X.sycl_queue
    434     )
    435     hev.wait()
    436     return res

TypeError: _copy_usm_ndarray_for_reshape(): incompatible function arguments. The following argument types are supported:
    1. (src: dpctl::tensor::usm_ndarray, dst: dpctl::tensor::usm_ndarray, shift: int, sycl_queue: dpctl.SyclQueue, depends: List[dpctl.SyclEvent] = []) -> Tuple[dpctl.SyclEvent, dpctl.SyclEvent]

Invoked with: kwargs: src=usm_ndarray([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.], dtype=float32), dst=usm_ndarray([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.], dtype=float32), shift=(2, 1, 3), sycl_queue=<dpctl.SyclQueue at 0x7f60f7fe4d40>
```

oleksandr-pavlyk:
Superseded by gh-1380

PR #1342, opened 2023-08-14
``dpctl.tensor.where`` output preserves memory order of inputs

ndgrigorian:
This PR adjusts the behavior of ``dpctl.tensor.where`` to preserve the memory layout of its inputs. This improves the access pattern of the kernel.

Performance of old behavior:
```

In [2]: dt = dpt.int32

In [3]: sh = (8160, 8160)

In [4]: ar1 = dpt.ones(sh, dtype=dt, order="C")[:4080, ::-2].mT

In [5]: ar2 = dpt.zeros(sh, dtype=dt, order="C")[:4080, ::-2].mT

In [6]: condition = dpt.zeros(sh, dtype=dt, order="C")[:4080, ::-2].mT

In [7]: res = dpt.where(condition, ar1, ar2)
CPU times: user 143 ms, sys: 53.4 ms, total: 196 ms
Wall time: 220 ms

In [8]: %time res = dpt.where(condition, ar1, ar2)
CPU times: user 48.3 ms, sys: 67.6 ms, total: 116 ms
Wall time: 119 ms

In [9]: %time res = dpt.where(condition, ar1, ar2)
CPU times: user 81.5 ms, sys: 31.1 ms, total: 113 ms
Wall time: 114 ms
```

New behavior:
```
In [9]: %time res = dpt.where(condition, ar1, ar2)
CPU times: user 26.9 ms, sys: 198 µs, total: 27.1 ms
Wall time: 27.8 ms

In [10]: %time res = dpt.where(condition, ar1, ar2)
CPU times: user 16.1 ms, sys: 13.8 ms, total: 29.9 ms
Wall time: 30.9 ms

In [11]: %time res = dpt.where(condition, ar1, ar2)
CPU times: user 23.3 ms, sys: 1.43 ms, total: 24.7 ms
Wall time: 26.7 ms
```

which is an improvement of ~4x in certain cases.

``_empty_like_triple_orderK`` is introduced for this purpose and a test was added to ensure that the output strides are as expected.

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -427,12 +427,16 @@ def _empty_like_triple_orderK(X1, X2, X3, dt, res_shape, usm_type, dev):
     nd1 = X1.ndim
     nd2 = X2.ndim
     nd3 = X3.ndim
-    if nd1 > nd2 and nd1 > nd3 and X1.shape == res_shape:
-        return _empty_like_orderK(X1, dt, usm_type, dev)
-    elif nd1 < nd2 and nd3 < nd2 and X2.shape == res_shape:
-        return _empty_like_orderK(X2, dt, usm_type, dev)
-    elif nd1 < nd3 and nd2 < nd3 and X3.shape == res_shape:
-        return _empty_like_orderK(X3, dt, usm_type, dev)
+    if X1.shape == res_shape and X2.shape == res_shape and len(res_shape) > nd3:
+        return _empty_like_pair_orderK(X1, X2, dt, res_shape, usm_type, dev)
+    elif (
+        X2.shape == res_shape and X3.shape == res_shape and len(res_shape) > nd1
+    ):
+        return _empty_like_pair_orderK(X2, X3, dt, res_shape, usm_type, dev)
+    elif (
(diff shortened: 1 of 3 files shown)


dpctl/tensor/_copy_utils.py
```diff
@@ -351,7 +351,9 @@ def _empty_like_orderK(X, dt, usm_type=None, dev=None):
         )
     st = list(X.strides)
     perm = sorted(
-        range(X.ndim), key=lambda d: builtins.abs(st[d]), reverse=True
+        range(X.ndim),
+        key=lambda d: builtins.abs(st[d] if X.shape[d] > 1 else 0),
```
oleksandr-pavlyk, line 355: Nit: 

```suggestion
        key=lambda d: builtins.abs(st[d]) if X.shape[d] > 1 else 0,
```

oleksandr-pavlyk (approved):
I think this looks good to go in @ndgrigorian. I had a small nitpick, but it just a matter of style.

dpctl/tensor/_copy_utils.py
ndgrigorian: Nice catch

Issue #1279, opened 2023-07-13
dpctl.tensor.abs() does not work on Iris Xe on Windows


npolina4:
```python
import dpctl.tensor as dpt
a = dpt.ones(10, dtype='c8')
dpt.abs(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\npolina\devel\miniconda3\envs\dpnp_new\envs\dpnp1\lib\site-packages\dpctl\tensor\_elementwise_common.py", line 101, in __call__
    ht, _ = self.unary_fn_(x, out, sycl_queue=exec_q)
RuntimeError: Required aspect fp64 is not supported on the device
```

oleksandr-pavlyk:
This problem is Windows specific, and can not be reproduced on Linux.

PR #1343, opened 2023-08-15
Fix gh-1279, implement tensor.allclose

oleksandr-pavlyk:
Fixes gh-1279.

1. Cleans up operator implementations in "dpctl/tensor/_usmarray.pyx" and removes unused static method to improve coverage
2. Provides private methods implementing `dpctl.tensor.abs` and `dpctl.tensor.sqrt` for complex types. This closes gh-1279.
3. Implements, and adds tests to `dpctl.tensor.allclose`.


oleksandr-pavlyk:
@ndgrigorian This PR superseded gh-1340, as it also ensures that `dpctl.tensor.abs(dpctl.tensor.asarray([-0.0]))` gives "+0.0".

oleksandr-pavlyk:
Verified that after the changes dpctl is able to execute problematic inputs on machine with Iris Xe running Windows:

```
(fix-gh-1279) C:\Users\opavlyk\gh>python -c "import dpctl.tensor as dpt; x = dpt.ones(10, dtype=dpt.complex64); y = dpt.abs(x); print(y)"
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]

(fix-gh-1279) C:\Users\opavlyk\gh>python -c "import dpctl.tensor as dpt; x = dpt.ones(10, dtype=dpt.complex64); y = dpt.sqrt(x); print(y)"
[1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j]
```

dpctl/tensor/_testing.py
```diff
@@ -0,0 +1,144 @@
...
+    check3 = dpt.all(r1[mr] == r2[mr])
+    if not check3:
+        return check3
+    m = dpt.isfinite(r1)
+    mv1 = r1[m]
+    mv2 = r2[m]
+    check4 = dpt.all(
+        dpt.abs(mv1 - mv2)
+        < atol + rtol * dpt.maximum(dpt.abs(mv1), dpt.abs(mv2))
+    )
```
ndgrigorian, line 94: According to Numpy, these conditions should be ``<=``

See here
```

In [2]: atol = 1e-5

In [3]: x = dpt.zeros(10, dtype="f4")

In [4]: y = dpt.asarray(atol, dtype="f4")

In [5]: dpt.allclose(x, y, atol=atol, rtol=0)
Out[5]: usm_ndarray(False)

In [6]: x_np, y_np = dpt.asnumpy(x), dpt.asnumpy(y)

In [7]: np.allclose(x_np, y_np, atol=atol, rtol=0)
Out[7]: True

In [8]: dpt.abs(x - y) < (atol + 0 * dpt.maximum(dpt.abs(x), dpt.abs(y)))
Out[8]:
usm_ndarray([False, False, False, False, False, False, False, False,
             False, False])

In [9]: dpt.abs(x - y) <= (atol + 0 * dpt.maximum(dpt.abs(x), dpt.abs(y)))
Out[9]:
              True,  True])
```

ndgrigorian: Also, I think it would be a good idea to add a test for the behavior of the ``atol`` and ``rtol`` parameters, much like the above.

dpctl/tensor/_testing.py
```diff
@@ -0,0 +1,144 @@
...
+    return check4
+
+
+def _allclose_others(r1, r2):
+    return dpt.all(r1 == r2)
+
+
+def allclose(a1, a2, atol=1e-5, rtol=1e-8, equal_nan=False):
+    """allclose(a1, a2, atol=1e-5, rtol=1e-8)
```
ndgrigorian, line 103: The tolerance defaults are reversed compared to Numpy's defaults, which are ``atol=1e-8`` and ``rtol=1e-5``.

ndgrigorian:
There is a compelling case for another change here:

https://github.com/numpy/numpy/issues/10161

so the implementation would become

``dpt.abs(x - y) <= dpt.maximum(rtol * dpt.maximum(dpt.abs(x), dpt.abs(y)), atol)``

We aren't nearly as burdened by backwards compatibility, so it's interesting to consider, as the incorrect cases pointed out in this issue are incorrect here, too.

```
In [11]: rtol, atol = 1e-4, 2e-5

In [12]: a = dpt.asarray(0.142253)

In [13]: b = dpt.asarray(0.142219)

In [14]: dpt.allclose(a, b, atol=atol, rtol=rtol)
Out[14]: usm_ndarray(True)

In [15]: dpt.abs(a - b) <= dpt.maximum(rtol * dpt.maximum(dpt.abs(a), dpt.abs(b)), atol)
Out[15]: usm_ndarray(False)
```

Downside would be another call to ``maximum``.

oleksandr-pavlyk:
> Downside would be another call to `maximum`.

It won't be a downside, we would just trade a call to `dpt.add` with a call to `dpt.maximum`.

oleksandr-pavlyk:
@ndgrigorian I have pushed the change to use `abs(a-b) <= max(atol, rtol*max(abs(a), abs(b))` formula. I have also reflected that in the `tensor.allclose` docstring.

dpctl/tensor/_testing.py
oleksandr-pavlyk: Fixed, thanks!

dpctl/tensor/_testing.py
oleksandr-pavlyk: Test added, and `<` replaced with `<=`.

ndgrigorian:
@oleksandr-pavlyk It makes sense to change other calls to ``std::abs`` in dpctl be changed to use this implementation?

And if so, maybe it should be in ``math_utils.hpp``. It may be better off in another PR to not let this one get too bloated, though.

oleksandr-pavlyk:
I do not think so, but even if we do, we should do it in a separate PR.

ndgrigorian:
@oleksandr-pavlyk 
The ``sqrt`` implementation seems to have some issues with very large values and very small values.

```

In [2]: val_max = dpt.finfo(dpt.complex128).max

In [3]: x = dpt.asarray(complex(val_max, val_max), dtype="c16")

In [4]: dpt.sqrt(x)
Out[4]: usm_ndarray(inf+0.j)

In [5]: x_np = np.asarray(complex(val_max, val_max), dtype="c16")

In [6]: np.sqrt(x_np)
Out[6]: (1.4730945569055655e+154+6.101757441282702e+153j)

In [7]: val_min = dpt.finfo(dpt.complex128).min

In [8]: x = dpt.asarray(complex(val_min, val_min), dtype="c16")

In [9]: dpt.sqrt(x)
Out[9]: usm_ndarray(0.-infj)

In [10]: x_np = np.asarray(complex(val_min, val_min), dtype="c16")

In [11]: np.sqrt(x_np)
Out[11]: (6.101757441282702e+153-1.4730945569055655e+154j)
```

oleksandr-pavlyk:
@ndgrigorian I pushed the change to normalize the scale of the arguments in `csqrt` and rescale the result appropriately, but use of these scaling functions causes failure to offload on Windows to recur. 

Perhaps the solution is to use sych functions from `sycl` namespace rather than from `std::` namespace.

ndgrigorian:
@oleksandr-pavlyk 

Since you removed other superfluous ``iostream`` includes in this PR, can you also remove the one in ``positive.hpp``?

oleksandr-pavlyk:
Reconfirmed that the latest build resolves gh-1279 on Windows machine with Iris Xe integrated GPU:

```
(gh-1279) C:\Users\user\repos>python -c "import dpctl.tensor as dpt; x = dpt.ones(10, dtype='c8'); y = dpt.sqrt(x); print(y)"
[1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j]

(gh-1279) C:\Users\user\repos>python -c "import dpctl.tensor as dpt; x = dpt.ones(10, dtype='c8'); y = dpt.abs(x); print(y)"
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]

(gh-1279) C:\Users\user\repos>python -m dpctl -f
Platform  0 ::
    Name        Intel(R) OpenCL
    Version     OpenCL 3.0 WINDOWS
    Vendor      Intel(R) Corporation
    Backend     opencl
    Num Devices 1
      # 0
        Name                11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
        Version             2023.16.6.0.28_042959
        Filter string       opencl:cpu:0
Platform  1 ::
    Name        Intel(R) OpenCL HD Graphics
    Version     OpenCL 3.0
    Vendor      Intel(R) Corporation
    Backend     opencl
    Num Devices 1
      # 0
        Name                Intel(R) Iris(R) Xe Graphics
        Version             31.0.101.4032
        Filter string       opencl:gpu:0
Platform  2 ::
    Name        Intel(R) Level-Zero
    Version     1.3
    Vendor      Intel(R) Corporation
    Backend     ext_oneapi_level_zero
    Num Devices 1
      # 0
        Name                Intel(R) Iris(R) Xe Graphics
        Version             1.3.24931
        Filter string       level_zero:gpu:0

(gh-1279) C:\Users\user\repos>conda list dpctl
# packages in environment at C:\Users\user\Miniconda3\envs\gh-1279:
#
# Name                    Version                   Build  Channel
dpctl                     0.14.6dev3      py310h81563b4_34    file:///C:/Users/user/repos/channel
```

ndgrigorian (approved):
This LGTM, thank you @oleksandr-pavlyk

Issue #1350, opened 2023-08-16
Copy usm_ndarray fails with order='K'


antonwolfy:
Below example is not working when both `copy=True` and `order='K'` are passed to `dpt.asarray`:
```python
a = dpt.ones((2, 4, 3))
b = dpt.permute_dims(a, (2, 0, 1))
c = dpt.asarray(b, copy=True, order='K')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[31], line 1
----> 1 c = dpt.asarray(b, copy=True, order='K')

File ~/miniconda3/envs/dpnp_dev/lib/python3.9/site-packages/dpctl/tensor/_ctors.py:580, in asarray(obj, dtype, device, copy, usm_type, sycl_queue, order)
    578 # handle instance(obj, usm_ndarray)
    579 if isinstance(obj, dpt.usm_ndarray):
--> 580     return _asarray_from_usm_ndarray(
    581         obj,
    582         dtype=dtype,
    583         copy=copy,
    584         usm_type=usm_type,
    585         sycl_queue=sycl_queue,
    586         order=order,
    587     )
    588 if hasattr(obj, "__sycl_usm_array_interface__"):
    589     ary = _usm_ndarray_from_suai(obj)

File ~/miniconda3/envs/dpnp_dev/lib/python3.9/site-packages/dpctl/tensor/_ctors.py:198, in _asarray_from_usm_ndarray(usm_ndary, dtype, copy, usm_type, sycl_queue, order)
    196 eq = dpctl.utils.get_execution_queue([usm_ndary.sycl_queue, copy_q])
    197 if eq is not None:
--> 198     hev, _ = ti._copy_usm_ndarray_into_usm_ndarray(
    199         src=usm_ndary, dst=res, sycl_queue=eq
    200     )
    201     hev.wait()
    202 else:

ValueError: Destination array can not accomodate all the elements of source array.

# while with copy=False or order='C' all works fine:
c = dpt.asarray(b, copy=False, order='K')
c is b
# Out: True

c = dpt.asarray(b, copy=True, order='C')
c.flags.c_contiguous
# Out: True

dpctl.__version__
# Out: '0.14.6dev3+4.g9afdb5802'
```

PR #1351, opened 2023-08-16
Fix bad order=K code logic in tensor.asarray 

oleksandr-pavlyk:
Closes gh-1350

`dpctl.tensor.asarray` implementation of `order='K'` keyword processing was replaced with call to tested `_empty_like_orderK` utility to fix the issue reported in gh-1350.

Few routines had to be shuffled to avoid import failure due to circular import dependencies.

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -23,6 +23,8 @@ import dpctl.memory as dpm
 import dpctl.tensor as dpt
 import dpctl.tensor._tensor_impl as ti
 import dpctl.utils
+from dpctl.tensor._copy_utils import _empty_like_orderK
+from dpctl.tensor._data_types import _get_dtype
 from dpctl.tensor._device import normalize_queue_device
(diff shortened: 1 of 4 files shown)


dpctl/tensor/_ctors.py
```diff
@@ -106,6 +90,21 @@ def _array_info_sequence(li):
     return (n,) + dim, dt, device
 
 
+def _c_contig_strides(sh):
```
ndgrigorian, line 93: Is this meant to be used somewhere or is it a leftover from something?

oleksandr-pavlyk: I am going to remove it. I was used in the initial variant of the fix which was replaced with the call to `_empty_like_orderK`.

oleksandr-pavlyk: I removed the code and force-pushed the changes.

ndgrigorian (approved):

antonwolfy (approved):
The issue is resolved, no new problem observed by dpnp tests scope. Thank you @oleksandr-pavlyk !

PR #1352, opened 2023-08-16
Fully enable ``usm_ndarray`` in-place arithmetic operators

ndgrigorian:
This PR completes the implementation of in-place arithmetic operators.

With current behavior, in-place operations simply call the standard implementation and then set the original memory equal to the result. This sometimes leads to unnecessary memory allocation.

By reworking the ``out`` keyword, either input operand array can now be used as the output array, and copies are only required for casting and where race conditions must be considered (i.e., when the ``out`` array overlaps only some of an input).


oleksandr-pavlyk (commented):
I think this is ready to go in, with follow-up PRs to add kernels for more in-place operations (divide, bitwise operations, etc.)

oleksandr-pavlyk (approved):
Thank you @ndgrigorian . Useful change!

PR #1355, opened 2023-08-19
usm_ndarray.real and usm_ndarray.imag now set flags correctly

ndgrigorian:
The ``real`` and ``imag`` properties of ``usm_ndarray`` were setting the flags of the returned view incorrectly.

In most cases, the output was not contiguous, but the input was, which caused most functions to behave incorrectly on views into the real or imaginary parts of complex arrays.

This PR corrects that, and also allows ``real`` and ``imag`` to work on 0-dimensional arrays.

### DIFF
--- a/dpctl/tensor/_usmarray.pyx
+++ b/dpctl/tensor/_usmarray.pyx
@@ -662,7 +662,7 @@ cdef class usm_ndarray:
         """
         if self.nd_ < 2:
             raise ValueError(
-                "array.mT requires array to have at least 2-dimensons."
+                "array.mT requires array to have at least 2 dimensions."
             )
         return _m_transpose(self)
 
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk (approved):
Good catch, @ndgrigorian !

It is reminds me that it is possible to set up real/imag values of complex arrays:

```python
import dpctl.tensor as dpt

sp_vals = dpt.asarray([dpt.nan, -dpt.nan, dpt.inf, -dpt.inf, -1.0, -0.0, 0.0, 1.0], dtype="f4")
x = dpt.empty((sp_vals.shape[0], sp_vals.shape[0]), dtype="c8")
x.real[...] = sp_vals[:, dpt.newaxis]
x.imag[...] = sp_vals[dpt.newaxis, :]
```

PR #1366, opened 2023-08-24
Conversion from raw to multi_ptr should be done with address_space_cast

oleksandr-pavlyk:
We used `sycl::multi_ptr` constructor instead of
`sycl::address_space_cast` previsously, and change in https://github.com/KhronosGroup/SYCL-Docs/pull/432 introduced `sycl::access:decorated::legacy` as the default which is deprecated in SYCL 2020 standard which highlighted the problem.

In using `sycl::address_space_cast` we specify
`sycl::access::decorated::yes`.


oleksandr-pavlyk:
Even before this change, we should have been using now deprecated `sycl::make_ptr` instead of `multi_ptr` constructor.

Merging.

PR #1369, opened 2023-08-25
Addressed "Very high" Coverity issue: dereferencing null pointer

oleksandr-pavlyk:
Ensured that `create_property_list` always returns an not-null unique pointer by creating an default-constructed `property_list` for the fall-through.

With this change we no longer need two branches for call to `sycl::queue` constructor, since `propList` is always available.

### DIFF
--- a/libsyclinterface/source/dpctl_sycl_queue_interface.cpp
+++ b/libsyclinterface/source/dpctl_sycl_queue_interface.cpp
@@ -121,7 +121,7 @@ bool set_kernel_arg(handler &cgh,
     return arg_set;
 }
 
-static std::unique_ptr<property_list> create_property_list(int properties)
+std::unique_ptr<property_list> create_property_list(int properties)
 {
     std::unique_ptr<property_list> propList;
     int _prop = properties;
--- a/libsyclinterface/source/dpctl_sycl_queue_interface.cpp
+++ b/libsyclinterface/source/dpctl_sycl_queue_interface.cpp
@@ -121,7 +121,7 @@ bool set_kernel_arg(handler &cgh,
     return arg_set;
 }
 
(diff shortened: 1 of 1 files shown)


ndgrigorian (approved):

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -121,7 +121,7 @@ bool set_kernel_arg(handler &cgh,
     return arg_set;
 }
 
-std::unique_ptr<property_list> create_property_list(int properties)
+static std::unique_ptr<property_list> create_property_list(int properties)
```
diptorupd, line 124: Why is the function changed to `static`?

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -143,6 +143,9 @@ std::unique_ptr<property_list> create_property_list(int properties)
         propList =
             std::make_unique<property_list>(sycl::property::queue::in_order());
     }
+    else {
```
diptorupd, line 146: I was not aware SYCL runtime allows an empty property_list. The change is good.

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
oleksandr-pavlyk: I sued static in C sense, as this function is not public, but it is already in the anonymous namespace, so going to revert addition of `static` keyword here.

oleksandr-pavlyk, line 124:
```suggestion
std::unique_ptr<property_list> create_property_list(int properties)
```

diptorupd (approved):

PR #1370, opened 2023-08-26
Fixes boolean indexing for strided masks

ndgrigorian:
This PR fixes some edge cases to boolean indexing that were not working as expected.

The cumulative sum used to get the count of set elements in the mask was misbehaving, especially for cases with negative strides (and most noticeably, the 1D case). This seems to have been caused by stride simplification can change the order elements are traversed in, which caused incorrect results.

Additionally, the ``offset`` parameter was also unused in when the contiguous kernel was called after simplification.

A new stride simplification method called ``compact_iteration_space`` was introduced to fix the problem.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/boolean_advanced_indexing.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/boolean_advanced_indexing.hpp
@@ -424,7 +424,6 @@ typedef size_t (*mask_positions_strided_impl_fn_ptr_t)(
     size_t,
     const char *,
     int,
-    py::ssize_t,
     const py::ssize_t *,
     char *,
     std::vector<sycl::event> const &);
@@ -434,7 +433,6 @@ size_t mask_positions_strided_impl(sycl::queue q,
                                    size_t n_elems,
                                    const char *mask,
                                    int nd,
-                                   py::ssize_t input_offset,
                                    const py::ssize_t *shape_strides,
                                    char *cumsum,
                                    std::vector<sycl::event> const &depends = {})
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk:
It was a mistake to use `simplify_iteration_space_1` here. Not only will negative strides cause trouble, but so will F-contiguous mask arrays.

However I don't think removing simplification altogether is the right solution. We could use `simplify_iteration_space_2` using C-contiguous strides for the shape of the input mask array for the second input.

ndgrigorian:
> It was a mistake to use `simplify_iteration_space_1` here. Not only will negative strides cause trouble, but so will F-contiguous mask arrays.
> 
> However I don't think removing simplification altogether is the right solution. We could use `simplify_iteration_space_2` using C-contiguous strides for the shape of the input mask array for the second input.

I've tested this out and it does also resolve the problem.

Would it be worthwhile then to write an abbreviated version of ``simplify_iteration_space``, something like ``simplify_iteration_space_by_c_contig_strides``, which removes logic that will never be true for a C-contig second input and doesn't have an unused strides vector and offset as byproducts?

oleksandr-pavlyk:
> Would it be worthwhile then to write an abbreviated version of `simplify_iteration_space`, something like `simplify_iteration_space_by_c_contig_strides`, which removes logic that will never be true for a C-contig second input and doesn't have an unused strides vector and offset as byproducts?

Yes, I was thinking about this as well. It is a good idea. I'd suggest naming it `compact_iteration_space` (restricted simplification).

ndgrigorian:
@oleksandr-pavlyk 
I've added an implementation for ``compact_iteration_space`` to the PR.

oleksandr-pavlyk (approved):

PR #1380, opened 2023-08-30
Add roll kernels

oleksandr-pavlyk:
This PR supersedes gh-1341.

This PR removes `shift` parameter from `_copy_usm_ndarray_for_reshape` and introduces `_copy_usm_ndarray_for_roll_1d` (for `axis=None`) and `_copy_usm_ndarray_for_roll_nd` functions where `shifts` argument is expected to be a tuple specifying shift size for every axis.

This dedicated function `_copy_usm_ndarray_for_roll_nd` is used to implement multi-axis `roll` function instead of a sequence of concurrent copy operations.

This PR will make input `dpt.roll(x, shift=(1,1))` raise a `TypeError` exception (since array API spec mandates that when shift is a tuple, an axis must also be a tuple).


ndgrigorian (approved):
@oleksandr-pavlyk 
I've added a test just to cover a couple of lines in the coverage CI.

I tested out the branch and reviewed the code, it looks good to me.
With the changes to the indexers, we may want to go back and comb through other kernels, make sure there isn't some unnecessary casting. But maybe that can wait for when we replace ``py::ssize_t`` with ``ssize_t``.

PR #1387, opened 2023-09-03
Change roll kernels to copy few elements per kernel.

oleksandr-pavlyk:
This PR adds batching (copying more than one element per work-item).


oleksandr-pavlyk:
This change does not bring significant performance gain, so closing.

PR #1398, opened 2023-09-11
Remove python_extension_module

ZzEeKkAa:
`python_extension_module` raises errors when building with cmake>=3.27. To workaround it, use new way to include python into CMake and use `python_add_library` instead of `add_library`.


oleksandr-pavlyk (approved):

oleksandr-pavlyk:
I rebased the branch, and added a commit to remove specifying minimum required version of Python (as we should not prevent users from building with older Python and supported NumPy). Good to go in. Thank you for contributing @ZzEeKkAa !

PR #1399, opened 2023-09-11
Check in of generic reduction templates and some reductions

ndgrigorian:
This pull request implements the necessary meta-programming utilities for reductions with arbitrary binary operators, templates for generic reductions, as well as the new functions ``dpctl.tensor.max``, ``dpctl.tensor.min``, ``dpctl.tensor.argmax``, and ``dpctl.tensor.argmin``.

These meta-programming utilities, such as ``Identity``, make it possible to define unique aspects of a reduction such that it may be passed to a generic functor for computation.


dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
```diff
@@ -66,6 +131,172 @@ size_t choose_workgroup_size(const size_t nelems,
...
+    static constexpr std::complex<T> value{std::numeric_limits<T>::infinity(),
+                                           std::numeric_limits<T>::infinity()};
+};
+
+// Plus
+
+template <typename T, class Op>
+using IsPlus = std::bool_constant<
+    std::is_same_v<Op, sycl::plus<T>> || std::is_same_v<Op, sycl::plus<void>> ||
+    std::is_same_v<Op, std::plus<T>> || std::is_same_v<Op, std::plus<T>>>;
```
oleksandr-pavlyk, line 280:
```suggestion
template <typename T, class Op>
using IsPlus = std::bool_constant<
    std::is_same_v<Op, sycl::plus<T>> || std::is_same_v<Op, sycl::plus<void>> ||
    std::is_same_v<Op, std::plus<T>>>;
```

Duplicated condition. For the curiosity, why are we testing for `sycl::plus<void>`? Is it a blanket cover-all test for all types?

oleksandr-pavlyk: I see that the same test is being used in `sycl` headers. 🤷

ndgrigorian: It's done in the headers. I believe it's intended to test for the ``void`` specialization of the ``std`` functions (as the ``sycl`` variants are simply the function objects from the STL ``functional`` header), which deduces the return and parameter types (see: https://en.cppreference.com/w/cpp/utility/functional/plus_void)

These can be safely removed for our custom ops, which don't have those overloads, so I will go ahead and do that as well.

oleksandr-pavlyk (approved):
I think this PR is ready to go in. Subsequent development can be done in new PRs. Thank you @ndgrigorian !

PR #1407, opened 2023-09-19
Exposing libtensor as C API

chudur-budur:
This PR exposes different APIs from `libtensor` as C functions. 

List of functions to be exposed (tentative):

- [ ] [`dpctl::tensor::py_internal::usm_ndarray_linear_sequence_affine()`](https://github.com/IntelPython/dpctl/blob/1277ba691a039313b06e41deef99d64b918d3d2b/dpctl/tensor/libtensor/source/linear_sequences.hpp#L47)
- [ ] [`dpctl::tensor::py_internal::usm_ndarray_linear_sequence_step()`](https://github.com/IntelPython/dpctl/blob/1277ba691a039313b06e41deef99d64b918d3d2b/dpctl/tensor/libtensor/source/linear_sequences.hpp#L41)

---

### DIFF
--- a/dpctl/CMakeLists.txt
+++ b/dpctl/CMakeLists.txt
@@ -49,7 +49,8 @@ elseif(UNIX)
         "-D_FORTIFY_SOURCE=2 "
         "-Wformat "
         "-Wformat-security "
-#       "-fno-strict-overflow "    # no-strict-overflow is implied by -fwrapv
+
+        # "-fno-strict-overflow "    # no-strict-overflow is implied by -fwrapv
         "-fno-delete-null-pointer-checks "
         "-fwrapv "
     )
@@ -82,11 +83,12 @@ add_custom_target(_build_time_create_dpctl_include ALL
     COMMAND ${CMAKE_COMMAND} -E make_directory ${DPCTL_INCLUDE_DIR}/syclinterface
     COMMAND ${CMAKE_COMMAND} -E make_directory ${DPCTL_INCLUDE_DIR}/syclinterface/Support
     COMMAND ${CMAKE_COMMAND} -E make_directory ${DPCTL_INCLUDE_DIR}/syclinterface/Config
-    DEPENDS DPCTLSyclInterface
+    DEPENDS DPCTLSyclInterface DPCTLTensorInterface
 )
 
 set(_copied_header_files)
 file(GLOB _syclinterface_h ${CMAKE_SOURCE_DIR}/libsyclinterface/include/*.h)
+
 foreach(hf ${_syclinterface_h})
     get_filename_component(_header_name ${hf} NAME)
     set(_target_header_file ${DPCTL_INCLUDE_DIR}/syclinterface/${_header_name})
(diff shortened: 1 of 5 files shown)


dpctl/CMakeLists.txt
```diff
@@ -82,23 +83,25 @@ add_custom_target(_build_time_create_dpctl_include ALL
     COMMAND ${CMAKE_COMMAND} -E make_directory ${DPCTL_INCLUDE_DIR}/syclinterface
     COMMAND ${CMAKE_COMMAND} -E make_directory ${DPCTL_INCLUDE_DIR}/syclinterface/Support
     COMMAND ${CMAKE_COMMAND} -E make_directory ${DPCTL_INCLUDE_DIR}/syclinterface/Config
-    DEPENDS DPCTLSyclInterface
+    DEPENDS DPCTLSyclInterface DPCTLTensorInterface
```
chudur-budur, line 86: @oleksandr-pavlyk I want the headers copied before `DPCTLTensorInterface` target starts, how do I do that?

libtensorinterface/include/dpctl_libtensor_linear_sequences_interface.h
```diff
@@ -0,0 +1,21 @@
...
+#include "dpctl_sycl_types.h"
+
+DPCTL_C_EXTERN_C_BEGIN
+
+typedef struct DPCTLSyclEventPair {
+    DPCTLSyclEventRef first;
+    DPCTLSyclEventRef second;
+} DPCTLSyclEventPair;
+
+DPCTLSyclEventPair
```
diptorupd, line 13: Quick comment based on a very superficial review of the changes.

I assume the plan is to call this function from LLVM IR inside a dpex-generated module? If so, returning a `struct` is a not recommended at all. There is a potential to run into all sort of platform-specific ABI issues. Clang on x86 will convert the type signature to return void and add a return value pointer. It might be different on other ABI.

Moreover, I am against creating these C wrappers for C++ `std` classes. I did that for `std::vector` in libsyclinterface to adhere to a pure C API, as we were not sure how the API will be used. If I were writing `dpctl` today I see no need to do it that way as the only place the `std::vector` wrapper is used is in the Cython interface and a better approach will be to make the Cython interface pure C++.

chudur-budur: What would be a better alternative? Should I return an opaque pointer? Or just keep it `void`? Becuase I don't think we are using that returned pair of events anywhere anyway. They are just used to check if those events were finished or not.

diptorupd: We should have the two event objects created by caller and passed to the function as pointers. They will be assigned in the function.

libtensorinterface/include/dpctl_libtensor_linear_sequences_interface.h
```diff
@@ -0,0 +1,21 @@
...
+    DPCTLSyclEventRef second;
+} DPCTLSyclEventPair;
+
+DPCTLSyclEventPair 
+DPCTL_usm_ndarray_linear_sequence_affine_int(uint64_t start, 
+                                            uint64_t end, 
+                                            char *data;
+                                            uint64_t include_endpoint, 
+                                            const DPCTLSyclQueueRef QRef,
+                                            void *events);
```
diptorupd, line 19: The size of the `events` vector should also be passed in as an argument.

diptorupd:
@chudur-budur as with the corresponding numba-dpex changes, the need to expose tensor constructors as a C API library does not exist any more. As such closing this PR.

PR #1410, opened 2023-09-20
Both PYTHON_EXECUTABLE and Python_EXECUTABLE needed to be checked in FindDpctl.cmake

chudur-budur:
In `cmake/FindDpctl.cmake` we need to check both `PYTHON_EXECUTABLE` and `Python_EXECUTABLE`.

### DIFF
--- a/cmake/FindDpctl.cmake
+++ b/cmake/FindDpctl.cmake
@@ -1,15 +1,15 @@
-#.rst:
+# .rst:
 #
 # Find the include directory for ``dpctl_capi.h``, ``dpctl4pybind11.hpp``.
 #
 # This module sets the following variables:
 #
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
Could you please motivate this change with an example where the current script breaks?

oleksandr-pavlyk:
Feel free to close this PR if it is difficult to find an example where `dpctl` has issues building without this change and where the changes fixes the issue.

chudur-budur:
> Feel free to close this PR if it is difficult to find an example where `dpctl` has issues building without this change and where the changes fixes the issue.

Okay I am closing it for now, I couldn't manage time to reproduce the problem, will do it later.

PR #1411, opened 2023-09-20
Use sycl ext oneapi experimental for complex

oleksandr-pavlyk:
This PR applies functions from `sycl::ext:;oneapi::experimental` extension to evaluate trigonometric, hyperbolic unary functions and their inverses.


ndgrigorian:
@oleksandr-pavlyk 
I've spotted a few cases where ``std`` functions are still used on complex values.

``std::log``
https://github.com/IntelPython/dpctl/blob/71b85aba1611968285608e1455f5a7ecf67e669c/dpctl/tensor/libtensor/include/kernels/elementwise_functions/acos.hpp#L108

https://github.com/IntelPython/dpctl/blob/71b85aba1611968285608e1455f5a7ecf67e669c/dpctl/tensor/libtensor/include/kernels/elementwise_functions/acosh.hpp#L115

https://github.com/IntelPython/dpctl/blob/71b85aba1611968285608e1455f5a7ecf67e669c/dpctl/tensor/libtensor/include/kernels/elementwise_functions/asin.hpp#L125

https://github.com/IntelPython/dpctl/blob/71b85aba1611968285608e1455f5a7ecf67e669c/dpctl/tensor/libtensor/include/kernels/elementwise_functions/asinh.hpp#L111

``std::sinh``
https://github.com/IntelPython/dpctl/blob/71b85aba1611968285608e1455f5a7ecf67e669c/dpctl/tensor/libtensor/include/kernels/elementwise_functions/sinh.hpp#L84

oleksandr-pavlyk:
@ndgrigorian ready to merge, please take a look

dpctl/tensor/libtensor/include/kernels/elementwise_functions/sqrt.hpp
```diff
@@ -69,12 +70,17 @@ template <typename argT, typename resT> struct SqrtFunctor
...
+#ifdef USE_SYCL_FOR_COMPLEX_TYPES
+            using realT = typename argT::value_type;
+
+            return exprm_ns::sqrt(exprm_ns::complex<realT>(in));
+#else
+#ifdef _WINDOWS
+            return std::sqrt(in);
+#else
+            return std::sqrt(in);
+#endif
```
ndgrigorian, line 82:
```suggestion
#else
            return std::sqrt(in);
#endif
```
Unless I am misunderstanding something, it seems the ``#ifdef _WINDOWS`` condition is unnecessary here.

It also seems that the custom ``csqrt`` variants are now unused, and it may be possible to remove them.

oleksandr-pavlyk: Thanks for catching this @ndgrigorian . It should be `#ifdef _WINDOWS` then use `csqrt` instead of `std::sqrt`. Will push the change.

dpctl/tensor/CMakeLists.txt
```diff
@@ -188,12 +188,20 @@ foreach(_src_fn ${_no_fast_math_sources})
   )
 endforeach()
 if (UNIX)
-  set_source_files_properties(
-    ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/elementwise_functions/abs.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/elementwise_functions/sqrt.cpp
-    PROPERTIES COMPILE_DEFINITIONS "USE_STD_ABS_FOR_COMPLEX_TYPES;USE_STD_SQRT_FOR_COMPLEX_TYPES")
+  set(_compiler_definitions "USE_SYCL_FOR_COMPLEX_TYPES")
+else()
+  set(_compiler_definitions "USE_SYCL_FOR_COMPLEX_TYPES")
```
ndgrigorian, line 193: This conditional seems redundant. Maybe it's not supposed to use "USE_SYCL_FOR_COMPLEX_TYPES" if not Unix?

oleksandr-pavlyk: It is supposed to be used on both platforms, but the `if` exists because different set of preprocessor variables used to be set on Windows and Linux. I can simplify the cmake script.

dpctl/tensor/libtensor/include/kernels/elementwise_functions/isfinite.hpp
```diff
@@ -46,6 +46,7 @@ namespace isfinite
 
 namespace py = pybind11;
 namespace td_ns = dpctl::tensor::type_dispatch;
+namespace exprm_ns = sycl::ext::oneapi::experimental;
```
ndgrigorian, line 49: Superfluous line

oleksandr-pavlyk: Fixed

ndgrigorian (approved):
LGTM, thank you @oleksandr-pavlyk ! Nice to see the internal CI passing again.

PR #1412, opened 2023-09-20
Changes to elementwise function tests that use complex data types

ndgrigorian:
This PR simplifies the `test_[func]_out_overlap` testing for elementwise funcs in dpctl by removing most of these tests, which were redundant, leaving only a test for overlap in `test_abs.py`.

This PR also modifies `pyproject.toml` and `conftest.py` to add a new pytest marker and complementary option, `broken_complex`, which skips some tests by default due to known issues on some platforms (esp. Windows).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ omit = [
 
 [tool.pytest.ini.options]
 markers = [
-    "broken_complex: mark a test that is skipped due to complex implementation",
+    "broken_complex: mark a test that is skipped on Windows due to complex implementation",
 ]
 minversion = "6.0"
 norecursedirs= [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,9 @@ omit = [
 ]
 
 [tool.pytest.ini.options]
+markers = [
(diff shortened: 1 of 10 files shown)


oleksandr-pavlyk:
If the other PR where elementwise tess were added, I skipped these test only on windows. Can it be supported?

ndgrigorian:
> If the other PR where elementwise tess were added, I skipped these test only on windows. Can it be supported?

I've pushed a change that should do just that.

pyproject.toml
```diff
@@ -33,6 +33,9 @@ omit = [
 ]
 
 [tool.pytest.ini.options]
+markers = [
+    "broken_complex: mark a test that is skipped due to complex implementation",
```
oleksandr-pavlyk, line 37:
```suggestion
    "broken_complex: mark a test that is skipped on Windows due to complex implementation",
```

oleksandr-pavlyk (approved):
Module a nit in the description of `broken_complex` marker, this is good. Thank you @ndgrigorian

oleksandr-pavlyk (approved):

PR #1415, opened 2023-09-21
call operator of ProjFunctor must be const

oleksandr-pavlyk:
Also changed the implementation of ProjFunctor to make compiler's job easier.

`const` is required for this struct to be usable in kernel if it is created on the host and passed to the kernel as parameter.


ndgrigorian (approved):

ndgrigorian:
@oleksandr-pavlyk similar to my PR, you may want to update `pyproject.toml` again to reflect `broken_complex` skips on all platforms

```
"broken_complex: mark a test that is skipped on Windows due to complex implementation"
```

ndgrigorian (approved):
Windows CI is still red for some complex64 trig function tests, but we can add another PR to skip those.

This LGTM, thank you @oleksandr-pavlyk

PR #1426, opened 2023-10-01
Implement product

oleksandr-pavlyk:
This is PR builds on top of work of #1399 and implements `_prod_over_axis` and the satellite associated type query function.

```

In [2]: import dpctl.tensor._tensor_impl as ti

In [4]: x = dpt.linspace(1, 2, num=32, dtype="f4")

In [5]: x
Out[5]:
usm_ndarray([1.       , 1.032258 , 1.0645161, 1.0967741, 1.1290323,
             1.1612903, 1.1935483, 1.2258065, 1.2580645, 1.2903225,
             1.3225806, 1.3548386, 1.3870968, 1.4193548, 1.4516128,
             1.483871 , 1.516129 , 1.548387 , 1.5806451, 1.6129032,
             1.6451613, 1.6774192, 1.7096773, 1.7419355, 1.7741935,
             1.8064516, 1.8387096, 1.8709677, 1.9032258, 1.9354837,
             1.9677418, 2.       ], dtype=float32)

In [6]: z = dpt.empty(tuple(), dtype="f4")

In [7]: ti._prod_over_axis(src=x, trailing_dims_to_reduce=1, dst=z, sycl_queue=x.sycl_queue)[0].wait()

In [8]: z
Out[8]: usm_ndarray(224213.45, dtype=float32)

In [9]: dpt.asnumpy(x).prod()
Out[9]: 224213.47
```

The product runs into problems for complex types, and does not give the expected output for integral types, so this is a work-in-progress.

### DIFF
--- a/dpctl/tensor/_reduction.py
+++ b/dpctl/tensor/_reduction.py
@@ -149,7 +149,7 @@ def sum(x, axis=None, dtype=None, keepdims=False):
     Args:
         x (usm_ndarray):
             input array.
-        axis (Optional[int, Tuple[int,...]]):
+        axis (Optional[int, Tuple[int, ...]]):
             axis or axes along which sums must be computed. If a tuple
             of unique integers, sums are computed over multiple axes.
             If `None`, the sum is computed over the entire array.
@@ -210,10 +210,10 @@ def prod(x, axis=None, dtype=None, keepdims=False):
     Args:
         x (usm_ndarray):
             input array.
-        axis (Optional[int, Tuple[int,...]]):
-            axis or axes along which sums must be computed. If a tuple
-            of unique integers, sums are computed over multiple axes.
-            If `None`, the sum is computed over the entire array.
+        axis (Optional[int, Tuple[int, ...]]):
+            axis or axes along which products must be computed. If a tuple
+            of unique integers, products are computed over multiple axes.
+            If `None`, the product is computed over the entire array.
             Default: `None`.
         dtype (Optional[dtype]):
             data type of the returned array. If `None`, the default data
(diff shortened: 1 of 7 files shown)


ndgrigorian:
@oleksandr-pavlyk In testing out a ``prod`` implementation, I ran into the same issue with integers. It seems to only happen for 64-bit integers on GPU, CPU worked fine. I believe it's a bug. I didn't test it for 64-bit floats, but I suspect it may have the same problem.

The workaround that I used was a ``Multiplies`` struct that returns ``x * y``.

dpctl/tensor/_reduction.py
```diff
@@ -202,6 +202,67 @@ def sum(x, axis=None, dtype=None, keepdims=False):
...
+    """
+    return _reduction_over_axis(
+        x,
+        axis,
+        dtype,
+        keepdims,
+        ti._prod_over_axis,
+        ti._prod_over_axis_dtype_supported,
+        _default_reduction_dtype,
+        _identity=0,
```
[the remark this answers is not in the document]
oleksandr-pavlyk: My bad. Fix is coming

dpctl/tensor/libtensor/source/reduction_over_axis.hpp
```diff
@@ -79,15 +80,24 @@ inline bool check_atomic_support(const sycl::queue &exec_q,
     return supports_atomics;
 }
 
+template <bool return_value>
+bool fixed_decision(const sycl::queue &, sycl::usm::alloc)
```
ndgrigorian, line 84: With this, it will probably be possible to refactor ``argmin`` / ``argmax`` to use the same ``py_reduction_over_axis`` binding

oleksandr-pavlyk: This is a dynamic check. Things are likely to fall apart during compilation as compiler will still need to generate code for atomic implementation. Worth a try though.

ndgrigorian (approved):
@oleksandr-pavlyk 
I pushed a small change with a tweak to some docstrings.

I think this is good to go in.

PR #1430, opened 2023-10-06
Fix host_task dependency not being passed to repeat 1d

ndgrigorian:
This PR fixes a bug in repeat where host_task event dependencies for allocating shapes and strides on the device were not always being passed to the kernel, which could cause the program to crash.

[See this CI run](https://github.com/IntelPython/dpctl/actions/runs/6426525416/job/17451716888)


oleksandr-pavlyk (approved):
:+1: for catching this. LGTM!

Issue #1434, opened 2023-10-10
dpctl.tensor.sum function raises an error: SyclAsynchronousError


npolina4:
Device: opencl:cpu
OS: Linux
Occurrence frequency: periodically
```python
import dpctl.tensor as dpt
x = dpt.arange(10, dtype=dpt.float32)
axes = [None, 0, (0,)]
for axis in axe:
    y = dpt.sum(a, axis=axis, dtype=dpt.complex128)
```
```python
dpctl/_sycl_queue.pyx:126: in dpctl._sycl_queue.default_async_error_handler
    ???
E   dpctl._sycl_queue.SyclAsynchronousError: 1

The above exception was the direct cause of the following exception:
y = dpt.sum(
/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/dpctl/tensor/_reduction.py:193: in sum
    return _reduction_over_axis(
/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/dpctl/tensor/_reduction.py:139: in _reduction_over_axis
    dpctl.SyclEvent.wait_for(host_tasks_list)
E   SystemError: <built-in function wait_for> returned a result with an error set
```

oleksandr-pavlyk:
See https://github.com/IntelPython/dpnp/actions/runs/6475840967/job/17584595766?pr=1585

oleksandr-pavlyk:
Corrected example:

```python
import dpctl.tensor as dpt

x = dpt.arange(10, dtype=dpt.float32, device='cpu')

for rep in range(16356):
    print(rep)
    axes = [None, 0, (0,)]
    for axis in axes:
        y = dpt.sum(x, axis=axis, dtype=dpt.complex128)
```

I could not reproduce the crash though.

ndgrigorian:
I also could not reproduce the crash.

Similar crashes, however, do seem to be appearing in the CI, in various tests, and always where 'wait_for' is being used.
See:
https://github.com/IntelPython/dpctl/actions/runs/6458849200/job/17534736280
https://github.com/IntelPython/dpctl/actions/runs/6461852353/job/17543529221
https://github.com/IntelPython/dpctl/actions/runs/6461789466/job/17543428471
(last is a Windows case)

PR #1436, opened 2023-10-12
Do not use wait and throw

oleksandr-pavlyk:
This PR arises from improve upon behavior noted in gh-1434

- Static method `dpctl.SyclEvent.wait_for` now uses `DPCTLEvent_Wait` instead of `DPCTLEvent_WaitAndThrow`, aligning with `dpctl.SyclEvent.wait` and `dpctl.SyclQueue.wait` methods.
- `dpctl.SyclAsyncException` has been removed
- Constructor calls to `DPCTLQueue_Create` and `DPCTLContext_Create` now pass `NULL` function pointer for the async handler function
- The `dpctl._sycl_queue.default_async_handler` function has been removed.

---

### DIFF
--- a/dpctl/_sycl_context.pyx
+++ b/dpctl/_sycl_context.pyx
@@ -48,7 +48,6 @@ from ._backend cimport (  # noqa: E211
     error_handler_callback,
 )
 from ._sycl_device cimport SyclDevice
-from ._sycl_queue cimport default_async_error_handler
 from ._sycl_device import SyclDeviceCreationError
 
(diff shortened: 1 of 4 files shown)


oleksandr-pavlyk:
Thange from using `event::wait_and_throw` to `event::wait` will result in asynchronous errors accumulating and will be printed out to `cerr` during context destruction.

The change in this PR is the correct thing to do.

ndgrigorian (approved):
In that case, LGTM

PR #1444, opened 2023-10-17
Implements ``dpctl.tensor.clip``

ndgrigorian:
This pull request implements the function ``clip``, for clipping elements of an array to a range.
Clip permits both Python scalar and ``dpctl.tensor.usm_ndarray`` arguments for its ``min`` and ``max`` arguments.

An example of the new functionality:
```

In [2]: x = dpt.arange(-2, 8, dtype="i4")

In [3]: dpt.clip(x, 0, 5)
Out[3]: usm_ndarray([0, 0, 0, 1, 2, 3, 4, 5, 5, 5], dtype=int32)
```

### DIFF
--- a/dpctl/tensor/CMakeLists.txt
+++ b/dpctl/tensor/CMakeLists.txt
@@ -50,6 +50,7 @@ set(_tensor_impl_sources
     ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/elementwise_functions.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/repeat.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/reduction_over_axis.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/clip.cpp
 )
 
 set(python_module_name _tensor_impl)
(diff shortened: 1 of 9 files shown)


oleksandr-pavlyk:
Encountered `TypeError`:

```
In [1]: import dpctl, dpctl.tensor as dpt, dpctl.utils as du

In [2]: x = dpt.arange(10)

In [3]: dpt.clip(x, 0, 10)
Out[3]: usm_ndarray([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

In [4]: dpt.clip(x, 0, 5)
Out[4]: usm_ndarray([0, 1, 2, 3, 4, 5, 5, 5, 5, 5])

In [5]: dpt.clip(x, 0, dpt.inf)
```

ndgrigorian:
> Encountered `TypeError`:
> 
> ```
> In [1]: import dpctl, dpctl.tensor as dpt, dpctl.utils as du
> 
> In [2]: x = dpt.arange(10)
> 
> In [3]: dpt.clip(x, 0, 10)
> Out[3]: usm_ndarray([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
> 
> In [4]: dpt.clip(x, 0, 5)
> Out[4]: usm_ndarray([0, 1, 2, 3, 4, 5, 5, 5, 5, 5])
> 
> In [5]: dpt.clip(x, 0, dpt.inf)
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> ```

This is currently the expected behavior, as the output is expected to be the same type as the input, and a float can't be cast to int.

antonwolfy:
Checked with existing dpnp tests around `clip` and all of them are passed.

oleksandr-pavlyk:
@ndgrigorian I think `ValueError` makes more sense in the example `dpt.bitwise_and(dpt.arange(10, dtype="i8"), dpt.inf)` as well. Thanks for bringing this up @ndgrigorian

ndgrigorian:
> @ndgrigorian I think `ValueError` makes more sense in the example `dpt.bitwise_and(dpt.arange(10, dtype="i8"), dpt.inf)` as well. Thanks for bringing this up @ndgrigorian

Making this change was more breaking than expected for elementwise functions. It will belong in a separate PR, as it will have to touch a lot of tests and unary functions. There seem to also be other cases of inappropriate dtype -> TypeError throughout the project. It may make sense to change those as well.

For now, that work is done on just clip. The PR is ready for review.

oleksandr-pavlyk (approved):

PR #1446, opened 2023-10-19
Implements ``dpctl.tensor.logsumexp`` and ``dpctl.tensor.reduce_hypot``

ndgrigorian:
This pull request proposes the additions of ``logsumexp`` and ``reduce_hypot`` into dpctl.

``reduce_hypot`` can be used to find the Euclidean distance in N-dimensional space
```

In [2]: x1 = dpt.full(10, 3, dtype="f4")

In [3]: x2 = dpt.ones(10, dtype="f4")

In [4]: d = dpt.reduce_hypot(x1 - x2)

In [5]: d
Out[5]: usm_ndarray(6.324554, dtype=float32)
```

``logsumexp`` can be used to sum probabilities stored as the logarithm of the probability.

This pull request also slips in a change to disable the use of ``sycl::atomic_ref`` for arithmetic reductions (``sum`` and ``prod``) of inexact data types due to concerns with numerical stability. To better support this change, dedicated kernels for tree reductions over contiguous inputs have been implemented.


ndgrigorian:
@oleksandr-pavlyk 
I've added the per-type per-function atomic support checks we've discussed.

oleksandr-pavlyk (approved):
LGTM! Thank you @ndgrigorian

PR #1449, opened 2023-10-22
Further splitting down elementwise functions

oleksandr-pavlyk:
Added missing includes of `"common_inplace.hpp"` and split elementwise functions into smaller files. 

Will continue until building succeeds.


oleksandr-pavlyk:
@ndgrigorian I think adding missing `#include "common_inplace.hpp"` did the trick, or maybe further splitting. I would continue splitting files to granularity of a single function, but then create `"elementwise_common.cpp"` file which exports initialization module that calls each initialization routine for each elementwise function.

ndgrigorian:
> @ndgrigorian I think adding missing `#include "common_inplace.hpp"` did the trick, or maybe further splitting. I would continue splitting files to granularity of a single function, but then create `"elementwise_common.cpp"` file which exports initialization module that calls each initialization routine for each elementwise function.

Great, I will go ahead and commit those changes to this branch, and after that we can merge this.

oleksandr-pavlyk:
@ndgrigorian I change `static inline` functions in `elementwise_functions.hpp`. I removed `static` qualifier and moved them into anonymous namespace per https://stackoverflow.com/questions/10876930/should-one-never-use-static-inline-function

ndgrigorian (approved):

PR #1462, opened 2023-11-01
Fix search reductions giving incorrect results for F-contiguous inputs

ndgrigorian:
This pull request restricts calls of contiguous variants for search reductions, now excluding cases where F-contiguous data is reduced to a single value.

These cases would return the index as if counting down columns rather than counting along the rows.


oleksandr-pavlyk (approved):

vtavana (approved):
Thank you, @ndgrigorian!
It works fine now.

>>> x=[[1, 2, 3],[6, 5, 4]]
>>> numpy.argmax(numpy.array(x, order='F'))
3
 
>>> dpt.argmax(dpt.asarray(x, order='F'))
usm_ndarray(3)

Issue #1461, opened 2023-10-30
Sensible performance degradation in `dpt.tensor.sum`


antonwolfy:
After merging #1446, `dpt.tensor.sum` became significantly slow (observing when running L2-norm benchmark for dpnp on PVC).
Before the PR:
```python
import dpctl, dpctl.tensor as dpt, numpy

dpctl.__version__
# Out: '0.15.1dev0+62.g2eba93eac'

sh = (134217728, 3)
dt = numpy.float32
a = dpt.ones(sh, dtype=dt)

%timeit _ = dpt.sum(a, axis=1, dtype=dt)
# 6.67 ms ± 9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit _ = dpt.sum(a, axis=1, dtype=dt)
# 6.64 ms ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The new times:
```python
import dpctl, dpctl.tensor as dpt, numpy

dpctl.__version__
# Out: '0.15.1dev0+63.g03fd73794'

sh = (134217728, 3)
dt = numpy.float32
a = dpt.ones(sh, dtype=dt)

%timeit _ = dpt.sum(a, axis=1, dtype=dt)
# 2.35 s ± 3.68 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit _ = dpt.sum(a, axis=1, dtype=dt)
# 2.35 s ± 6.04 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Devices info:
```bash
$ python -m dpctl -f
Platform  0 ::
    Name        Intel(R) OpenCL
    Version     OpenCL 3.0 LINUX
    Vendor      Intel(R) Corporation
    Backend     opencl
    Num Devices 1
      # 0
        Name                Intel(R) Xeon(R) Platinum 8469 CPU @2.00GHz
        Version             2023.16.6.0.22_223734
        Filter string       opencl:cpu:0
Platform  1 ::
    Name        Intel(R) OpenCL Graphics
    Version     OpenCL 3.0
    Vendor      Intel(R) Corporation
    Backend     opencl
    Num Devices 1
      # 0
        Name                Intel(R) Data Center GPU Max 1100
        Version             23.35.27191.25
        Filter string       opencl:gpu:0
Platform  2 ::
    Name        Intel(R) FPGA Emulation Platform for OpenCL(TM)
[...]

PR #1463, opened 2023-11-02
Improve performance of reduction for small number of elements to reduce for types where tree-reduction is needed

oleksandr-pavlyk:
1. Renamed misspelled variable
2. If `reduction_nelems` is small, use `SequentialReductionKernel` for tree-reductions as it is done for atomic reduction
3. Tweak scaling down logic for moderately-sized number of elements to reduce. Closes gh-1461

   We should also use `max_wg` if the `iter_nelems` is very small (one), since choosing `max_wg` for large` iter_nelems` may lead to under-utilization of GPU.

---

### DIFF
--- a/dpctl/tensor/_elementwise_common.py
+++ b/dpctl/tensor/_elementwise_common.py
@@ -39,6 +39,31 @@ from ._type_utils import (
 class UnaryElementwiseFunc:
     """
     Class that implements unary element-wise functions.
+
+    Args:
+        name (str):
+            Name of the unary function
+        result_type_resovler_fn (callable):
+            Function that takes dtype of the input and
+            returns the dtype of the result if the
+            implementation functions supports it, or
+            returns `None` otherwise.
+        unary_dp_impl_fn (callable):
+            Data-parallel implementation function with signature
+            `impl_fn(src: usm_ndarray, dst: usm_ndarray,
+             sycl_queue: SyclQueue, depends: Optional[List[SyclEvent]])`
+            where the `src` is the argument array, `dst` is the
+            array to be populated with function values, effectively
+            evaluating `dst = func(src)`.
+            The `impl_fn` is expected to return a 2-tuple of `SyclEvent`s.
+            The first event corresponds to data-management host tasks,
+            including lifetime management of argument Python objects to ensure
+            that their associated USM allocation is not freed before offloaded
+            computational tasks complete execution, while the second event
+            corresponds to computational tasks associated with function
+            evaluation.
+        docs (str):
+            Documentation string for the unary function.
     """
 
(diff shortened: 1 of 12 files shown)


antonwolfy:
The PR resolves the performance issue gh-1461, L2 norm benchmark is back to expected values. Thank you!

ndgrigorian:
@oleksandr-pavlyk 
I realized while looking through ``reductions.hpp`` that floating-point types are not included in ``TypePairSupportDataForCompReductionAtomic``, so even though we permit atomics for ``max`` and ``min`` on floating-point arrays, there aren't any functions to call in the table.

``float`` and ``double`` should be added to fix this.

dpctl/tensor/_elementwise_common.py
```diff
@@ -363,6 +411,56 @@ def _get_shape(o):
...
+    Args:
+        name (str):
+            Name of the unary function
+        result_type_resovle_fn (callable):
+            Function that takes dtypes of the input and
+            returns the dtype of the result if the
+            implementation functions supports it, or
+            returns `None` otherwise.
+        binary_dp_impl_fn (callable):
+            Data-parallel umplementation function with signature
```
ndgrigorian, line 424: Nit
```suggestion
            Data-parallel implementation function with signature
```

dpctl/tensor/_elementwise_common.py
```diff
@@ -363,6 +411,56 @@ def _get_shape(o):
...
+            The first event corresponds to data-management host tasks,
+            including lifetime management of argument Python objects to ensure
+            that their associated USM allocation is not freed before offloaded
+            computational tasks complete execution, while the second event
+            corresponds to computational tasks associated with function
+            evaluation.
+        docs (str):
+            Documentation string for the unary function.
+        binary_inplace_fn (callable, optional):
+            Data-parallel omplementation function with signature
```
ndgrigorian, line 440:
```suggestion
            Data-parallel implementation function with signature
```

ndgrigorian:
https://github.com/IntelPython/dpctl/blob/af28d981d513729a31349fe19cb1da10515ab849/dpctl/tensor/libtensor/source/tensor_ctors.cpp#L40
https://github.com/IntelPython/dpctl/blob/af28d981d513729a31349fe19cb1da10515ab849/dpctl/tensor/libtensor/source/tensor_ctors.cpp#L51

These includes in ``tensor_ctors.cpp`` should be unnecessary with these changes.

oleksandr-pavlyk:
> dpctl/dpctl/tensor/libtensor/source/tensor_ctors.cpp

Thank you @ndgrigorian, good catch!

oleksandr-pavlyk:
Closes gh-1461

ndgrigorian (approved):
Thank you @oleksandr-pavlyk, LGTM

PR #1464, opened 2023-11-03
Improves performance of search reductions for small numbers of elements

ndgrigorian:
This pull request adds ``SequentialSearchReduction`` functor to improve ``argmax`` and ``argmin`` for small arrays.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/reductions.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/reductions.hpp
@@ -3476,7 +3476,9 @@ public:
                             idx_val = static_cast<outT>(m);
                         }
                     }
-                    else if constexpr (std::is_floating_point_v<argT>) {
+                    else if constexpr (std::is_floating_point_v<argT> ||
+                                       std::is_same_v<argT, sycl::half>)
+                    {
                         if (val < red_val || std::isnan(val)) {
                             red_val = val;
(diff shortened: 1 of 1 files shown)


dpctl/tensor/libtensor/include/kernels/reductions.hpp
```diff
@@ -3400,6 +3400,125 @@ struct LogSumExpOverAxis0TempsContigFactory
...
+                        using dpctl::tensor::math_utils::greater_complex;
+                        if (greater_complex<argT>(val, red_val) ||
+                            std::isnan(std::real(val)) ||
+                            std::isnan(std::imag(val)))
+                        {
+                            red_val = val;
+                            idx_val = static_cast<outT>(m);
+                        }
+                    }
+                    else if constexpr (std::is_floating_point_v<argT>) {
```
oleksandr-pavlyk, line 3503:
```suggestion
                    else if constexpr (std::is_floating_point_v<argT> || std::is_same_v<argT, sycl::half>) {
```

ndgrigorian: Good catch. This change should be applied to the other search reduction functor as well, I'll go ahead and do that here.

oleksandr-pavlyk (approved):
Looks great @ndgrigorian , thank you!

PR #1465, opened 2023-11-04
Implements statistical functions ``mean``, ``std``, ``var``

### LINKED ISSUE #1456: dpctl does not return a zero-length array when no reduction is going to be done
The below example behaves differently between dcptl and numpy:
```python
import numpy, dpctl, dpctl.tensor as dpt

dpctl.__version__
# Out: '0.15.1dev0+63.g03fd73794'

a = numpy.ones((0, 2), dtype='f4')
numpy.min(a, axis=(1,))
# Out: array([], dtype=float32)

b = dpt.ones((0, 2), dtype='f4')
dpt.min(b, axis=(1,))
Cell In[15], line 1
----> 1 dpt.min(b, axis=(1,))

File ~/miniconda3/envs/dpnp_dev/lib/python3.9/site-packages/dpctl/tensor/_reduction.py:499, in min(x, axis, keepdims)
    474 def min(x, axis=None, keepdims=False):
    475     """min(x, axis=None, dtype=None, keepdims=False)
    476
    477     Calculates the minimum value of the input array `x`.
   (...)
    497             array has the same data type as `x`.
    498     """
--> 499     return _comparison_over_axis(x, axis, keepdims, ti._min_over_axis)

File ~/miniconda3/envs/dpnp_dev/lib/python3.9/site-packages/dpctl/tensor/_reduction.py:421, in _comparison_over_axis(x, axis, keepdims, _reduction_fn)
    419 res_usm_type = x.usm_type
    420 if x.size == 0:
--> 421     raise ValueError("reduction does not support zero-size arrays")
    422 if red_nd == 0:
    423     return x

ValueError: reduction does not support zero-size arrays
```
As I can see python Array API [notes](https://data-apis.org/array-api/latest/API_specification/generated/array_api.min.html) the use case as `implementation-defined`. Should dpctl align the behavior with numpy here?

ndgrigorian:
This pull request
1. Implements ``dpctl.tensor.mean``, ``dpctl.tensor.std``, and ``dpctl.tensor.var`` as top-level functions which leverage tensor submodules.

2. Resolves #1456 by adding logic to tree reduction implementations that populate the output array with the identity when reducing over axes which include a zero-size axis and by adding early exits for `py_reduction_over_axis`, `py_tree_reduction_over_axis`, and `py_search_over_axis` when the destination array is of zero size. 
This permitted removing `identity` as an argument to the general reduction Python API, as well as removing logic for zero-size array inputs.

3. Adjusts reduction Python API functions to return a copy of the input array, rather than the input array itself, when no reduction will be performed (i.e., `axis=()`)

4. Adds additional supported type combinations to `sum` and `prod` reductions. Prior to #1446 , reductions over `int` inputs with `float` outputs had dedicated kernels when using atomics to avoid a copy. This pull request adds these type combinations for tree reductions.

### DIFF
--- a/dpctl/tensor/_statistical_functions.py
+++ b/dpctl/tensor/_statistical_functions.py
@@ -1,0 +1,281 @@
+#                       Data Parallel Control (dpctl)
+#
+#  Copyright 2020-2023 Intel Corporation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from numpy.core.numeric import normalize_axis_tuple
+
+import dpctl
+import dpctl.tensor as dpt
+import dpctl.tensor._tensor_elementwise_impl as tei
+import dpctl.tensor._tensor_impl as ti
+import dpctl.tensor._tensor_reductions_impl as tri
+
+from ._reduction import _default_reduction_dtype
+
+
+def _var_impl(x, axis, correction, keepdims):
+    nd = x.ndim
+    if axis is None:
+        axis = tuple(range(nd))
+    if not isinstance(axis, (tuple, list)):
+        axis = (axis,)
+    axis = normalize_axis_tuple(axis, nd, "axis")
+    perm = []
+    nelems = 1
+    for i in range(nd):
+        if i not in axis:
+            perm.append(i)
+        else:
+            nelems *= x.shape[i]
+    red_nd = len(axis)
+    perm = perm + list(axis)
+    q = x.sycl_queue
+    inp_dt = x.dtype
+    res_dt = (
+        inp_dt
+        if inp_dt.kind == "f"
+        else dpt.dtype(ti.default_device_fp_type(q))
+    )
+    res_usm_type = x.usm_type
+
+    deps = []
+    host_tasks_list = []
+    if inp_dt != res_dt:
+        buf = dpt.empty_like(x, dtype=res_dt)
+        ht_e_buf, c_e1 = ti._copy_usm_ndarray_into_usm_ndarray(
+            src=x, dst=buf, sycl_queue=q
+        )
+        deps.append(c_e1)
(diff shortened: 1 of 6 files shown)


dpctl/tensor/libtensor/include/kernels/reductions.hpp
```diff
@@ -2727,32 +2773,44 @@ struct TypePairSupportDataForSumReductionTemps
...
+        td_ns::TypePairDefinedEntry<argTy, std::uint32_t, outTy, double>,
 
         // input int64_t
         td_ns::TypePairDefinedEntry<argTy, std::int64_t, outTy, std::int64_t>,
+        td_ns::TypePairDefinedEntry<argTy, std::int64_t, outTy, double>,
 
-        // input uint32_t
+        // input uint64_t
         td_ns::TypePairDefinedEntry<argTy, std::uint64_t, outTy, std::uint64_t>,
+        td_ns::TypePairDefinedEntry<argTy, std::uint64_t, outTy, double>,
```
oleksandr-pavlyk, line 2813: Should we also allow `float` accumulation type here for devices that do not support `float64`?

ndgrigorian: Yes, I went ahead and added it. We permit it for `reduce_hypot` and `logsumexp`, so it made sense to do it here.

dpctl/tensor/_statistical_functions.py
```diff
@@ -0,0 +1,379 @@
...
+    if keepdims:
+        res_shape = res_shape + (1,) * sum_nd
+        inv_perm = sorted(range(nd), key=lambda d: perm[d])
+        res = dpt.permute_dims(dpt.reshape(res, res_shape), inv_perm)
+
+    res_shape = res.shape
+    # in-place divide
+    nelems_arr = dpt.asarray(
+        nelems, dtype=res_dt, usm_type=res_usm_type, sycl_queue=q
+    )
```
oleksandr-pavlyk, line 262:
```suggestion
    # in-place divide
    den_dt = dpt.finfo(res_dt).dtype if res_dt.kind == "c" else res_dt
    nelems_arr = dpt.asarray(
        nelems, dtype=den_dt, usm_type=res_usm_type, sycl_queue=q
    )
```

oleksandr-pavlyk: This would avoid promoting `nelems` to complex type, fixing `nan`-propagation woes.

oleksandr-pavlyk (dismissed):
Other than the suggestion, this looks great @ndgrigorian ! Thank you for working on implementing these functions.

oleksandr-pavlyk (dismissed):
Thank you @ndgrigorian

oleksandr-pavlyk (approved):

PR #1469, opened 2023-11-06
Add array API inspection utilities to ``dpctl.tensor``

ndgrigorian:
This pull request proposes implementing ``dpctl.tensor.__array_namespace_info__``, an inspection utility coming soon to the array API specification.

In addition, this PR adds an ``__array_api_version__`` attribute [as per the specification](https://data-apis.org/array-api/latest/API_specification/generated/array_api.__array_api_version__.html#array_api.__array_api_version__) and experimentally removes the ``--ci`` flag from array API testing.

### DIFF
--- a/dpctl/tensor/_array_api.py
+++ b/dpctl/tensor/_array_api.py
@@ -1,0 +1,116 @@
+#                       Data Parallel Control (dpctl)
+#
+#  Copyright 2020-2023 Intel Corporation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


dpctl/tensor/_array_api.py
```diff
@@ -0,0 +1,118 @@
...
+            }
+        else:
+            return {
+                key: val
+                for key, val in self._all_dtypes.items()
+                if key not in ignored_types and _isdtype_impl(val, kind)
+            }
+
+    def devices(self):
+        return dpctl.get_devices()
```
oleksandr-pavlyk, line 118: It is possible to run in an environment where CUDA device is available, but `dpctl` was not compiled to offload to NVidia GPUs. 

Perhaps we should thin this list by ensuring that we can create an `dpt.arange` for this device and cache that list. 

It does not need to be done in this PR though

oleksandr-pavlyk:
The `__array_namespace_info__` class would benefit from a docstring, even if concise.

oleksandr-pavlyk (approved):
Thank you @ndgrigorian

Issue #1468, opened 2023-11-06
dpt.sum is not doing computation in requested `dtype` on Windows


oleksandr-pavlyk:
```
>>> import dpctl.tensor as dpt
>>> a = dpt.full((2, 3, 4), 123456789, dtype=dpt.int32)
>>> dpt.sum(a, dtype="f4")
usm_ndarray(-1.3320044e+09, dtype=float32)
>>> dpt.sum(dpt.astype(a, "f4"), dtype="f4")
usm_ndarray(2.962963e+09, dtype=float32)
```

Both outputs should agree. This is cause test failure observed in IntelPython/dpnp#1617

PR #1470, opened 2023-11-07
Fix for gh-1468 in arithmetic reduction when type promotion is needed

oleksandr-pavlyk:
Closes gh-1468

This PR changes `_reduce_over_axis` when an implementation to reduce using requested `dtype` for given input type does not exist. The change casts input array into a temporary array of requested `dtype` and calls reduction on the temporary (silent assumption that implementation to reduce input array using the same input dtype is available).

### DIFF
--- a/dpctl/tensor/_reduction.py
+++ b/dpctl/tensor/_reduction.py
@@ -445,7 +445,7 @@ def _comparison_over_axis(x, axis, keepdims, _reduction_fn):
 
 
 def max(x, axis=None, keepdims=False):
-    """max(x, axis=None, dtype=None, keepdims=False)
+    """max(x, axis=None, keepdims=False)
 
     Calculates the maximum value of the input array `x`.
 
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_reduction.py
```diff
@@ -118,7 +118,7 @@ def _reduction_over_axis(
                 dpt.full(
                     res_shape,
                     _identity,
-                    dtype=_default_reduction_type_fn(inp_dt, q),
+                    dtype=dtype,
```
ndgrigorian, line 121: With this change, the call to ``astype`` is no longer necessary.

It also means that when ``logsumexp`` or ``reduce_hypot`` reduce over an empty axis or array (e.g., ``dpt.logsumexp(dpt.ones((1, 0, 1), dtype="i4"), axis=1, dtype="i4")``) you get ``OverflowError`` instead of silently casting the identity to the output type.

For now, the ``astype`` can be removed. I've experimented with removing this branch in #1465 too.

oleksandr-pavlyk: Yes, and it should not be `dtype`, it should be `res_dt`.

ndgrigorian (approved):
Thank you @oleksandr-pavlyk !

PR #1472, opened 2023-11-13
Changed dpctl.tensor.arange error from ValueError to TypeError for bool data type

npolina4:
Changed dpctl.tensor.arange error from ValueError to TypeError for bool data type for compatibility with numpy error

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -821,7 +821,7 @@ def arange(
     except TypeError:
         sh = 0
     if is_bool and sh > 2:
-        raise ValueError("no fill-function for boolean data type")
+        raise TypeError("no fill-function for boolean data type")
     res = dpt.usm_ndarray(
         (sh,),
         dtype=dt,


oleksandr-pavlyk:
What prompts this change? The change is also a departure from NumPy's behavior:

```

In [1]: import numpy as np

In [2]: np.arange(10, dtype="?")
Cell In[2], line 1
----> 1 np.arange(10, dtype="?")

ValueError: no fill-function for data-type.
```

npolina4:
Different error for NumPy 1.24.3
``` python
>>> import numpy as np
>>> np.arange(10, dtype="?")
  File "<stdin>", line 1, in <module>
TypeError: arange() is only supported for booleans when the result has at most length 2.
>>> np.__version__
'1.24.3'
```

oleksandr-pavlyk:
I see no reason to change behavior of `dpctl.tensor` here. I think `ValueError` is more appropriate than the `TypeError` in this instance. In fact, the [description text](https://docs.python.org/3/library/exceptions.html#TypeError) for `TypeError` explicitly states:

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

PR #1473, opened 2023-11-13
Fixed dpctl.tensor.result_type function for scalars

npolina4:
Added support for weak scalars data type.

### DIFF
--- a/dpctl/tensor/_clip.py
+++ b/dpctl/tensor/_clip.py
@@ -24,20 +24,23 @@ from dpctl.tensor._copy_utils import (
     _empty_like_triple_orderK,
 )
 from dpctl.tensor._elementwise_common import (
-    WeakBooleanType,
-    WeakComplexType,
-    WeakFloatingType,
-    WeakIntegralType,
     _get_dtype,
     _get_queue_usm_type,
     _get_shape,
-    _strong_dtype_num_kind,
     _validate_dtype,
-    _weak_type_num_kind,
 )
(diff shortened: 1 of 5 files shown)


dpctl/tensor/_manipulation_functions.py
```diff
@@ -727,12 +727,16 @@ def result_type(*arrays_and_dtypes):
...
-        for X in arrays_and_dtypes
-    ]
-
-    _supported_dtype(dtypes)
+    dtypes = []
+    for X in arrays_and_dtypes:
+        if isinstance(X, dpt.usm_ndarray):
+            dtypes.append(X.dtype)
+        elif np.isscalar(X):
+            dtypes.append(X)
```
oleksandr-pavlyk, line 735: This branch is not necessary, as scalars are treated in the fall-through clause:

```
In [1]: import numpy as np

In [2]: np.float32(2)
Out[2]: 2.0

In [3]: np.dtype(_)
Out[3]: dtype('float32')
```

oleksandr-pavlyk: I see that I not correct for Python scalars though.

ndgrigorian (approved):
Behavior for Numpy scalars seems worth noting
(the device does not support fp64)
```
In [9]: dpt.result_type(dpt.empty(0, dtype="f4"), np.float64(2))
Out[9]: dtype('float32')
```
But we usually treat them as Python scalars anyway, so I don't see a reason to change anything here.

Overall LGTM!

PR #1474, opened 2023-11-24
Adds element-wise functions `angle` and `reciprocal`

ndgrigorian:
This pull request implements two new element-wise functions, `dpctl.tensor.angle` and `dpctl.tensor.reciprocal`.

To support the inclusion of `reciprocal`, an acceptance function argument to the `UnaryElementwiseFunc` class is added to align the behavior of `reciprocal` and `divide`.

This pull request also slips in bug fixes to `usm_ndarray` -- namely, the `real` and `imag` properties.

`imag` was not allocating to the same queue as the original array for real-valued types, and would raise an exception for real-valued 0D arrays.

Both `real` and `imag` would return `None` for `float16` arrays as a result of incorrect logic in the typenum checks.

### DIFF
--- a/dpctl/tensor/_elementwise_common.py
+++ b/dpctl/tensor/_elementwise_common.py
@@ -28,7 +28,8 @@ from dpctl.utils import ExecutionPlacementError
 
 from ._copy_utils import _empty_like_orderK, _empty_like_pair_orderK
 from ._type_utils import (
-    _acceptance_fn_default,
+    _acceptance_fn_default1,
+    _acceptance_fn_default2,
     _all_data_types,
     _find_buf_dtype,
     _find_buf_dtype2,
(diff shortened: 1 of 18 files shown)


dpctl/tensor/_elementwise_common.py
```diff
@@ -28,7 +28,8 @@
 
 from ._copy_utils import _empty_like_orderK, _empty_like_pair_orderK
 from ._type_utils import (
-    _acceptance_fn_default,
+    _acceptance_fn_default1,
+    _acceptance_fn_default2,
```
oleksandr-pavlyk, line 32: Perhaps it is better to name these `_acceptance_fn_default_unary`, and `_acceptance_fn_default_binary` ?

dpctl/tensor/_usmarray.pyx
```diff
@@ -698,7 +699,8 @@ cdef class usm_ndarray:
         """ Returns imaginary component for arrays with complex data-types
         and returns zero array for all other data-types.
         """
-        if (self.typenum_ < UAR_CFLOAT):
+        # explicitly check for UAR_HALF, which is greater than UAR_CFLOAT
+        if (self.typenum_ < UAR_CFLOAT or self.typenum_ == UAR_HALF):
```
oleksandr-pavlyk, line 703: Good catch!

oleksandr-pavlyk (dismissed):
Module the renaming suggestion this PR is in perfect shape. The renaming can be addressed in a separate PR though.

dpctl/tensor/_elementwise_common.py
ndgrigorian: I agree, that's much clearer

ndgrigorian:
> Module the renaming suggestion this PR is in perfect shape. The renaming can be addressed in a separate PR though.

I noticed a typo in the `rsqrt` docstring, so I went ahead and added the acceptance function naming change as well.

oleksandr-pavlyk (approved):

PR #1483, opened 2023-12-06
Implementation of sort/argsort and set functions per array API spec

oleksandr-pavlyk:
Implemented data-parallel merge-sort algorithm, which powers array API operations `dpct.tensor.sort` and `dpctl.tensor.argsort`. Implemented set functions: `dpctl.tensor.unique_values`, `dpctl.tensor.unique_counts`, `dpctl.tensor.unique_inverse`, and `doctl.tensor.unique_all`.


ndgrigorian:
@oleksandr-pavlyk 
 Tests seem to pass using the asynchronous `unique_*` functions. Should I go ahead and remove `set_functions_sync.py` and transition to the asynchronous implementations?

oleksandr-pavlyk:
@ndgrigorian I have switched to asynchronous implementation per your feedback, but kept the synchronous one around just in case.

dpctl/tensor/_set_functions_async.py
```diff
@@ -94,7 +96,7 @@ def unique_values(x: dpt.usm_ndarray) -> dpt.usm_ndarray:
         )
         host_tasks.append(ht_ev)
         ht_ev, sort_ev = _sort_ascending(
-            src=fx,
+            src=tmp,
```
oleksandr-pavlyk, line 99: 👍 Good catch!

ndgrigorian (approved):
LGTM, thank you @oleksandr-pavlyk , very close to full array API coverage.

vtavana (approved):
Thank you, works fine for zero dimensional arrays and arrays with `nan`.

PR #1485, opened 2023-12-08
Work around sub_group load/store issues

oleksandr-pavlyk:
Work-around issue with sub_group::load, sub_group::store functions

These functions are from oneAPI sycl_ext_oneapi_group_load_store extension.
https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_group_load_store.asciidoc

The current implementation of these primitives seem to require certain pointer alignment.

This PR adds bool template parameter to contig functors for all elementwise functions,
and the elementwise_common_impl checks alignment all pointers. If any is not aligned to
the expected boundary, a generic implementation is used instead of the one using
`sg.load`/`sg.store`.


ndgrigorian (approved):
Assuming CI succeeds, LGTM

PR #1497, opened 2024-01-18
Do not use link-huge-device-code option

oleksandr-pavlyk:
This PR remove use of `-fsycl-link-huge-device-code` link option when building tensor_impl modules on Linux.


ndgrigorian:
According to the [llvm docs](https://intel.github.io/llvm-docs/UsersManual.html) the option was deprecated in favor of `-flink-huge-device-code`, and according to release notes this option is mostly identical. Would it make more sense to use the replacement?

oleksandr-pavlyk:
It is being removed not because of the deprecation warning, but because it is no longer necessary after splitting single monolithic `_tensor_impl` into a collection of smaller modules, and suspectedly makes linking of device code somewhat longer.

ndgrigorian (approved):
I see, then in that case, this LGTM.

PR #1501, opened 2024-01-24
Fix dpctl.tensor.isdtype function signature

ndgrigorian:
This pull request moves `isdtype` to `_type_utils.py` to make the function signature match what is required by the array API specification.

Additionally experiments with using the `--disable-deadline` when running array API tests, as some tests fail for exceeding the deadline.


oleksandr-pavlyk (approved):
Thank you @ndgrigorian

PR #1502, opened 2024-01-24
Fix generate-docs workflow breakage when run on PR closure

oleksandr-pavlyk:
Fix for generate documentation workflow. Only run "Publish docs" step if PR is not being closed.


ndgrigorian (approved):
Hopefully this will fix the problem

Issue #1489, opened 2023-12-16
dpctl could not properly handle NaN edges in `clip` function


antonwolfy:
dpctl raises an exception for either any edge as NaN or both edges as NaN passed to `dpctl.tensor.clip` function.
While NumPy accept any of below combinations without an issue:
```python
a = numpy.arange(7)

numpy.clip(a, numpy.nan, None)
numpy.clip(a, None, numpy.nan)
numpy.clip(a, numpy.nan, numpy.nan)
numpy.clip(a, -2, numpy.nan)
numpy.clip(a, numpy.nan, 10)
```
The below is example of errors raised by dpctl:
```python
dpctl.__version__
# Out: '0.15.1dev2+28.gebf94b04cf'

# create an array
a = dpt.arange(7)

# clip with any edge as NaN, where another is None
dpt.clip(a, numpy.nan, None)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 dpt.clip(a, numpy.nan, None)

File ~/miniconda3/envs/dpnp_dev/lib/python3.9/site-packages/dpctl/tensor/_clip.py:433, in clip(x, min, max, out, order)
    429     raise ValueError(
    430         "only one of `min` and `max` is permitted to be `None`"
    431     )
    432 elif max is None:
--> 433     return _clip_none(x, min, out, order, tei._maximum)
    434 elif min is None:
    435     return _clip_none(x, max, out, order, tei._minimum)

File ~/miniconda3/envs/dpnp_dev/lib/python3.9/site-packages/dpctl/tensor/_clip.py:249, in _clip_none(x, val, out, order, _binary_fn)
    247 _fp16 = sycl_dev.has_aspect_fp16
    248 _fp64 = sycl_dev.has_aspect_fp64
--> 249 if not _can_cast(val_dtype, res_dt, _fp16, _fp64):
    250     raise ValueError(
    251         f"function 'clip' does not support input types "
    252         f"({x_dtype}, {val_dtype}), "
    253         "and the inputs could not be safely coerced to any "
    254         "supported types according to the casting rule ''safe''."
    255     )
    257 orig_out = out

File ~/miniconda3/envs/dpnp_dev/lib/python3.9/site-packages/dpctl/tensor/_type_utils.py:103, in _can_cast(from_, to_, _fp16, _fp64)
[...]

PR #1505, opened 2024-01-24
`clip` permits `None` for `min` and `max`

ndgrigorian:
This pull request adjusts dpctl.tensor.clip to permit `min` and `max` to be `None` simultaneously, resulting in effectively a no-op.

In addition, makes a small adjustment to dpctl.tensor.copy, which would return a `TypeError` instead of raising it, and resolves https://github.com/IntelPython/dpctl/issues/1489 by fixing two typos in `_resolve_one_strong_one_weak_types`

### DIFF
--- a/dpctl/tensor/_clip.py
+++ b/dpctl/tensor/_clip.py
@@ -399,10 +399,12 @@ def clip(x, /, min=None, max=None, out=None, order="K"):
         x (usm_ndarray): Array containing elements to clip.
             Must be compatible with `min` and `max` according
             to broadcasting rules.
-        min ({None, usm_ndarray}, optional): Array containing minimum values.
+        min ({None, Union[usm_ndarray, bool, int, float, complex]}, optional):
+            Array containing minimum values.
             Must be compatible with `x` and `max` according
             to broadcasting rules.
-        max ({None, usm_ndarray}, optional): Array containing maximum values.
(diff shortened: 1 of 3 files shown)


dpctl/tensor/_clip.py
```diff
@@ -405,11 +402,9 @@ def clip(x, min=None, max=None, out=None, order="K"):
         min ({None, usm_ndarray}, optional): Array containing minimum values.
```
oleksandr-pavlyk, line 402: Should we document that `min` and `max` can actually be Python scalars?

```

In [1]: import dpctl.tensor as dpt, matmul as mm, dpctl

In [2]: x = dpt.remainder(dpt.arange(10**6, dtype="i4"), 171)

In [3]: y = dpt.clip(x, 0, 5)

In [4]: dpt.min(y), dpt.max(y)
Out[4]: (usm_ndarray(0, dtype=int32), usm_ndarray(5, dtype=int32))
```

ndgrigorian: Yes, we definitely should, good catch

oleksandr-pavlyk (approved):
LGTM! Nice to see coverage climb up like this. Thank you @ndgrigorian

antonwolfy (commented):
Works as expected with DPNP tests now. Thank you @ndgrigorian !

PR #1506, opened 2024-01-24
Add checks to ensure test suite passes even if no devices are available

oleksandr-pavlyk:
Adds checks for tests to fail if no device is available to DPC++ runtime. 

Checked using the following input on a machine with not CUDA card:

```
ONEAPI_DEVICE_SELECTOR=cuda:gpu python -m pytest -s dpctl/tests
```


ndgrigorian (approved):
LGTM!

PR #1508, opened 2024-01-25
Service improvements

oleksandr-pavlyk:
`dpctl/__main__.py` no longer triggers loading of libsycl library for queries

`python -m dpctl --includes` will no longer load SYCL runtime libraries.

Module file acquired three more options

      --include-dir
      --tensor-include-dir
      --library-dir

This would output absolute path to directory of include headers,
tensor include headers, and directory containing DPCTLSyclInterface
library.

`FindDpctl.cmake` has been renamed to `dpctl-config.cmake`. 

This PR also installs `dpctl-config.cmake` to `${PREFIX}/lib/cmake` folder, making `find_package(Dpctl)` work out of the box with `dpctl` installed in the Python environment. 

With local build of `dpctl`,  downstream projects can hint CMake to finding Dpctl package by
using standard mechanism:

```
-DDpctl_ROOT=$(python -m dpctl --cmakedir)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,10 @@ file(GLOB _cmake_scripts ${CMAKE_SOURCE_DIR}/cmake/*.cmake)
 install(FILES ${_cmake_scripts}
     DESTINATION dpctl/resources/cmake
 )
+install(FILES
+  ${CMAKE_SOURCE_DIR}/cmake/dpctl-config.cmake
+  DESTINATION lib/cmake/dpctl
+)
 
 if (DPCTL_GENERATE_DOCS)
     add_subdirectory(docs)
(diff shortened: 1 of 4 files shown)


antonwolfy:
I've checked `find_package(Dpctl)` out of the box, without defining any `DPCTL_MODULE_PATH` flag, and it works fine in dpnp.

ndgrigorian (approved):
I have tested building extensions with these changes, and using `-DDpctl_ROOT=$(python -m dpctl --cmakedir)`

It works for me out of the box, so LGTM

oleksandr-pavlyk:
It turns out `python -m dpctl --include-dir` does trip loading of `DPCTLSyclInterface` and `libsycl.so.*`. This is because `python -m dpctl` effectively does `import dpctl.__main__` and this imports `dpctl.__init__` which triggers loading of `libDPCTLSyclInterface`. 

To accomplish running queries like `--include-dir` without loading `libsycl` requires creating a sub-package, `dpctl_info`, and requiring users to do `python -m dpctl_info --include-dir` instead. This is a task for another PR though.

Issue #1512, opened 2024-01-26
tensor.asarray raises unexpected TypeError for out of bound inputs


oleksandr-pavlyk:
```
dpt.asarray(-9223372036854775809, dtype=dpt.uint8)
# raises TypeError: Numpy array of data type object is not supported. Please convert the input to an array with numeric data type.
```

Array API test suite expected `OverflowError`:

```
np.asarray(-9223372036854775809, dtype=dpt.uint8)
# raises OverflowError: Python int too large to convert to C long
```

This discrepancy is behind Array API test suite failure of `test_int_to_dtype`.

oleksandr-pavlyk:
@ndgrigorian

PR #1513, opened 2024-01-26
Fix array API test failure caused by issue reported in gh-1512

ndgrigorian:
This pull request adjusts `dpt.asarray` such that the `dtype` argument is passed to `np.asarray` call when called with a Python scalar as input

This guarantees the expected `OverflowError` is thrown.

Additionally, this PR adds tests for this behavior, for `copy=False` errors to `test_tensor_asarray`, and removes a superfluous `NotImplementedError` for Python sequences from `asarray`.

Closes gh-1512

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -632,17 +632,13 @@ def asarray(
                 usm_type=usm_type,
                 order=order,
             )
-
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk (approved):
Thank you @ndgrigorian , thanks for fixing gh-1512.

PR #1516, opened 2024-01-30
Decouple pybind11 and kernels

oleksandr-pavlyk:
Moved all uses of pybind11 types and functions out of `libtensor/include/kernels`. 

- Uses of `py::object` and casters were moved to respective `*.cpp` files.
- `dpctl::tensor::ssize_t` was defined
- Uses of `py::ssize_t` in function signatures replaced with of library's `ssize_t` type
- `#include "pybind11/pybind11.h"` removed (speeding up compilation time)
- Added `static_assert(std::is_same_v(py::ssize_t, dpctl::tensor:;ssize_t>)` in aome `*.cpp` files 
- `#include <CL/sycl.hpp>` has been replaced with `#include <sycl/sycl.hpp>`.

---


oleksandr-pavlyk:
I deliberately left `"libtensor/include/utils"` untouched, since some of those files work with `dpctl4pybind11` and with `pybind11` itself. 

Files in the kernel subfolder do not, so I think we should leave `memory_overlap` as is.

dpctl/tensor/libtensor/source/boolean_reductions.cpp
```diff
@@ -37,10 +37,13 @@
 #include "dpctl4pybind11.hpp"
 
 #include "kernels/boolean_reductions.hpp"
+#include "kernels/dpctl_tensor_types.hpp"
 #include "utils/type_utils.hpp"
 
 namespace py = pybind11;
 
+static_assert(std::is_same_v<py::ssize_t, dpctl::tensor::ssize_t>);
```
ndgrigorian, line 45: Is there a specific reason behind the selection of files for these static asserts?

oleksandr-pavlyk: The thinking was to make sure that compiling fails in case of type mismatch. I thought `static_assert` should be present in every Python extension module.

ndgrigorian: That makes sense. Would it perhaps be better then to move the asserts to the module `.cpp` files themselves?

So then `static_assert(std::is_same_v<py::ssize_t, dpctl::tensor::ssize_t>);` would appear in `tensor_elementwise.cpp` and `tensor_reductions.cpp` instead?

ndgrigorian (approved):
LGTM, thank you for tackling this @oleksandr-pavlyk !

PR #1527, opened 2024-02-05
Fixes setting of writable flag for views and writing to read-only arrays with `out` keyword

ndgrigorian:
This pull request makes changes to `_usmarray.pyx` to guarantee that the writable flag is handled correctly for views created via `real`, `imag`, `T`, `mT`, and shape-setting and indexing properties.

`__getitem__` now only adjusts the flag after all writing to the result has been performed, fixing the practice of setting elements to a read-only array.

Elementwise functions, `clip`, and `matmul` also now contain early (top-level) checks for the writable flag of a provided `out` array, fixing an edge case where the writable flag would be overridden by copying to the original array. This practice also prevents unnecessary computation and memory allocation.

### DIFF
--- a/dpctl/tensor/_clip.py
+++ b/dpctl/tensor/_clip.py
@@ -440,6 +440,9 @@ def clip(x, /, min=None, max=None, out=None, order="K"):
                     f"{type(out)}"
                 )
 
+            if not out.flags.writable:
+                raise ValueError("provided `out` array is read-only")
+
             if out.shape != x.shape:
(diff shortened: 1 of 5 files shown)


oleksandr-pavlyk (approved):
More tests are needed, and perhaps `usm_ndarray._set_writable_flag` could be implemented using newly added `_copy_writable` function, but this can be done in a separate PR. 

Thanks @ndgrigorian for finding and fixing this bug!

PR #1531, opened 2024-02-07
Add to source files compiled with no-fast-math

ndgrigorian:
This pull request adds where.cpp, copy_and_cast_usm_to_usm.cpp, and boolean reductions to the source files to be compiled with `-fno-fast-math`. This resolves issues with floating point NaNs being interpreted as false when cast to bool with a CUDA backend.

Additionally, factors out `boolean_reductions.cpp` and `boolean_reductions.hpp` into the reductions submodule.


ndgrigorian:
Examples of fixed functionality
```

In [2]: x = dpt.full(10, dpt.nan, dtype="f4", device="cuda")

In [3]: dpt.astype(x, "?")
Out[3]:
              True,  True])

In [4]: dpt.any(x)
Out[4]: usm_ndarray(True)

In [5]: dpt.where(x, dpt.asarray(1), dpt.asarray(0))
Out[5]: usm_ndarray([1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
```

oleksandr-pavlyk (approved):
Thank you for tracking the issue down and fixing it @ndgrigorian 

LGTM!

PR #1534, opened 2024-02-07
Make some offset indexer classes constexpr

oleksandr-pavlyk:
Making `NoOpIndexer` constructor and the call operator a `constexpr` allows compiler to completely inline its definition at earlier optimization stages.

Also replaced stray `py::ssize_t` with `ssize_t`.


ndgrigorian (approved):
LGTM, this is a smart change! I'd like to note here that the constexpr calls to the getters are enabled by the constexpr constructors.

PR #1546, opened 2024-02-13
Removed _gemm_batch_old_nm_impl, got rid of using _new_, _old_ in names

oleksandr-pavlyk:
This PR replaced uses of `_old_nm_` functions with `_new_nm_` functions, and then removed `GemmBatchFunctorThreadNM` and the `_old_nm_` unused functions. 

After this all `_new_` and `_old_` parts in the functions names were deleted.


ndgrigorian (approved):
I've tested this compared to master, I don't see a significant performance change.

LGTM!

PR #1550, opened 2024-02-15
Rewrite some properties to delete arrays before dealing w/ objects

oleksandr-pavlyk:
This PR changes implementations of two `SyclDevice` properties to ensure that returned array is deallocated before return Python object value is constructed.

MaxWorkGroupSizes2d, MaxWorkGroupSizes1d returns an array. Copy array elements into local variables, free array before packing those values into Python tuple.


ndgrigorian (approved):
LGTM!

PR #1554, opened 2024-02-16
Fix negative for boolean inputs

oleksandr-pavlyk:
@npolina4 pointed out that array API states it should not be supported for Boolean inputs.

There is no dedicated loop for type `bool`, but input `dtype` `bool `gets promoted to integral types, and `negative` function returns result for the promoted type.

The change makes `UnaryElementwiseFunc` instance for negative use acceptance functions, which raises an exception for promotion with `dpt.bool` as starting data type.

A test is added.


npolina4 (approved):
Thank you @oleksandr-pavlyk!

ndgrigorian (approved):
LGTM!

ndgrigorian (approved):

PR #1558, opened 2024-02-22
Allow setting local_accessors as kernel arguments in DPCTLQueue_Submit

diptorupd:

The `libsyclinterface` library provides a minimal C API interface to submit kernels to a `sycl::queue` and use the `sycl::handler` classes `set_arg` to set the arguments with which to launch a kernel.

The PR adds a way to create a `local_accessor` object from a set of required metadata: rank, dtype, shape, and then add the `local_accessor` as a kernel argument. To enable the changes, the DPCTLKernelArgType enum was expanded and necessary changes made to the queue interface.

The changes are still a work in progress and need test cases. The PR is opened to get early feedback and evaluate the feasibility of the approach.


oleksandr-pavlyk (dismissed):
Looks good to me, modulo a nit comment. Thank you @diptorupd

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -38,6 +38,100 @@
...
+            CGH.set_arg(IDX, la);                                              \
+            return true;                                                       \
+        }                                                                      \
+        case DPCTL_UNSIGNED_INT:                                               \
+        {                                                                      \
+            auto la = local_accessor<unsigned int, NDIM>(R, CGH);              \
+            CGH.set_arg(IDX, la);                                              \
+            return true;                                                       \
+        }                                                                      \
+        case DPCTL_UNSIGNED_INT8:                                              \
```
oleksandr-pavlyk, line 80: Is this different from `DPCTL_UNSIGNED_CHAR`? If these are aliases, we could never get good coverage

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -38,6 +38,100 @@
...
+            CGH.set_arg(IDX, la);                                              \
+            return true;                                                       \
+        }                                                                      \
+        case DPCTL_UNSIGNED_LONG_LONG:                                         \
+        {                                                                      \
+            auto la = local_accessor<unsigned long long, NDIM>(R, CGH);        \
+            CGH.set_arg(IDX, la);                                              \
+            return true;                                                       \
+        }                                                                      \
+        case DPCTL_SIZE_T:                                                     \
```
oleksandr-pavlyk, line 110: `size_t` is an alias for `unsigned long long`, so the same comment applies here.

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
diptorupd: Are you running coverage on `libsyclinterface`? https://coveralls.io/github/IntelPython/dpctl seems to indicate that `libsyclinterface` is not part of dpctl's coverage numbers. Moreover not all files of `libsyclinterface` are included.

oleksandr-pavlyk: It is: 

image

oleksandr-pavlyk: I see that many files from `libsyclinterface` are indeed absent from the report. I may have an idea why. I will open a PR to find out

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
diptorupd: In practice maybe, but not in theory. `size_t` equates to `SIZE_MAX` that is the largest size object on a given architecture. The only restriction is it should be at least 16 bit wide, its max is platform dependent. C99's `unsigned long long` is guaranteed to be at least 64 bit wide.

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
diptorupd: Well the aggregate may not be getting impacted if the overall lines of code for dpctl repo is not swayed much by libsycinterface. Even then it seems spurious.

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
diptorupd: @oleksandr-pavlyk would it make sense to switch over from the C99 types to the C++11 types only? That remove all support for `long`, `unsigned long` _etc._ and only support `[u]int*_t` types? 

As the only usage is from numba-dpex, there should not be any other impact. In numba-dpex the logic for supported kernel argument types is as follows:

```python
if ty == types.int32 or isinstance(ty, types.scalars.IntegerLiteral):
        # DPCTL_LONG_LONG
        val = context.get_constant(types.int32, 9)
    elif ty == types.uint32:
        # DPCTL_UNSIGNED_LONG_LONG
        val = context.get_constant(types.int32, 10)
    elif ty == types.boolean:
        # DPCTL_UNSIGNED_INT
        val = context.get_constant(types.int32, 5)
    elif ty == types.int64:
        # DPCTL_LONG_LONG
        val = context.get_constant(types.int32, 9)
    elif ty == types.uint64:
        # DPCTL_SIZE_T
        val = context.get_constant(types.int32, 11)
    elif ty == types.float32:
        # DPCTL_FLOAT
        val = context.get_constant(types.int32, 12)
    elif ty == types.float64:
        # DPCTL_DOUBLE
        val = context.get_constant(types.int32, 13)
    elif ty == types.voidptr or isinstance(ty, types.CPointer):
        # DPCTL_VOID_PTR
        val = context.get_constant(types.int32, 15)
    else:
        raise NotImplementedError
```

libsyclinterface/tests/test_sycl_queue_submit_local_accessor_arg.cpp
```diff
@@ -0,0 +1,364 @@
...
+using namespace dpctl::syclinterface;
+
+typedef struct MDLocalAccessorTy
+{
+    size_t ndim;
+    DPCTLKernelArgType dpctl_type_id;
+    size_t dim0;
+    size_t dim1;
+    size_t dim2;
+} MDLocalAccessor;
```
oleksandr-pavlyk, line 54: The need to define this in test file is a hint that we should have this `typedef` in a header file of libsyclinterface itself.

oleksandr-pavlyk (approved):
Looks good to me. Thank you @diptorupd  !

PR #1559, opened 2024-02-22
Cleans up elementwise function docstrings

ndgrigorian:
This pull request makes a number of changes to the docstrings of element-wise functions, with a focus on guaranteeing that the information in the docstrings is correct and removing typos.

Each function, when referencing expected data types of inputs, now refers to the types which will not be cast. For example, `dpt.divide` arguments now state that they expect floating-point data, as otherwise, the data will be copied and cast.

### DIFF
--- a/dpctl/tensor/_elementwise_funcs.py
+++ b/dpctl/tensor/_elementwise_funcs.py
@@ -35,7 +35,7 @@ Args:
         Input array, expected to have numeric data type.
     out ({None, usm_ndarray}, optional):
         Output array to populate.
-        Array have the correct shape and the expected data type.
+        Array must have the correct shape and the expected data type.
     order ("C","F","A","K", optional):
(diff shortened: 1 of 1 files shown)


dpctl/tensor/_elementwise_funcs.py
```diff
@@ -954,10 +964,12 @@
 # U20: ==== LOG         (x)
 _log_docstring = """
 log(x, out=None, order='K')
-Computes the natural logarithm element-wise.
+
+Computes the natural logarithm for each element `x_i` of input array `x`.
```
oleksandr-pavlyk, line 968: If we used "Test if each element ..." above, we should be consistent and use "Compute the natural logarithm ...". 

No need to change it in this PR, but we should make a note of this and apply changes eventually.

oleksandr-pavlyk (approved):
Thanks for tackling these changes @ndgrigorian 

This PR is good to go in.

PR #1560, opened 2024-02-23
x.__dlpack_device__() returns ID of the parent device

oleksandr-pavlyk:
`x.__dlpack_device__()` returns ID of the parent device, if allocation device is not a root device.

It gives ID (position in `sycl::device::get_devices()`) for the root device that the array allocation device descends from for sub-devices, or ID of the allocation device is that device is a root device.

This allows sharing through DLPack of allocation on sub-devices, as long as default platform context for the platform the device is from was used in allocation.


ndgrigorian (approved):
These changes LGTM, thank you @oleksandr-pavlyk !

PR #1574, opened 2024-02-29
Modularized test_main tests

oleksandr-pavlyk:
Also added code to work around a bug with CPU runtime, which may cause DPC++ RT to not see CPU device in the sub-process if the parent process has initialized CPU plugin.


ndgrigorian (approved):
LGTM!

PR #1581, opened 2024-03-05
Updated the DPCTLKernelArgType enum values to be aligned with C++11 types

diptorupd:

Previously, the `DPCTLKernelArgType` had a mix of C99 types (`long`, `long long`, *etc.*) and C++ types (`size_t`, `int8_t`). After these changes the enum values have been updated to only include the C++11 integer types: `[u]int[8|16|32|64]_t`, `float` and `double`.

A Python enum corresponding to the libsyclinterface C++ enum was also added to make it simpler to get the types from Python. It will help numba-dpex not have to hard code magic values in its code base.

The `DPCTLQueue_SubmitRange` and `DPCTLQueue_SubmitNdRange` functions were updated to silence compiler warnings.

Unit tests were added for verifying the submission of one-dimensional range kernels each with a different kernel argument type.

NOTE: `DPCTLQueue_SubmitNdRange` will not be tested as part of this PR. I will be be adding those tests as part of the local accessor support as a kernel argument PR.

New libsyclinterfrace unit tests to test for all possible kernel argument types, including a negative test.


libsyclinterface/dbg_build.sh
```diff
@@ -2,30 +2,31 @@
...
-INSTALL_PREFIX=`pwd`/../install
+INSTALL_PREFIX=$(pwd)/../install
 rm -rf ${INSTALL_PREFIX}
 
 cmake                                                       \
     -DCMAKE_BUILD_TYPE=Debug                                \
     -DCMAKE_C_COMPILER=icx                                  \
-    -DCMAKE_CXX_COMPILER=dpcpp                              \
+    -DCMAKE_CXX_COMPILER=icpx                               \
+    -DCMAKE_CXX_FLAGS=-fsycl                                \
```
oleksandr-pavlyk, line 14: Probably the right way of integrating SYCL and cmake would be to use `find_package(IntelSycl)`, but this could be done in a separate PR.

oleksandr-pavlyk (commented):
Thank you @diptorupd 

It would also be useful to document in the comment how multiple SPVs are compiled into a single one. 

Also, if one is on a machine without support for `double` type, IGC route of generation of the kernel for `double` won't work, how does one generate the kernel then?

oleksandr-pavlyk:
> Thank you @diptorupd
> 
> It would also be useful to document in the comment how multiple SPVs are compiled into a single one.
> 
> Also, if one is on a machine without support for `double` type, IGC route of generation of the kernel for `double` won't work, how does one generate the kernel then?

Clarifying my own misunderstanding. When C++ is compiled with default `-fsycl-split-device-code` value, kernel bundles are formed as per aspect required, so a kernel function would contain kernels specialized for all types but `float16` and `float64` and two more kernel bundles may be formed, one for `float16` and one for `float64`. 

This way, JIT-ting the generic bundle would work for all devices, regardless of the aspect supported. Hence no manual bundling of multiple SPV files is required.

The recipe to extract SPV from executable produced by DPC++ assumes IGC is available. An alternative approach is to use `SYCL_DUMP_IMAGES` environment variable [supported](https://intel.github.io/llvm-docs/EnvironmentVariables.html) by  oneAPI DPC++, which would dump all offload sections of fat binary to the current working directory. 

With these clarified, the PR is good to go in, although I'd still prefer if these notes were copied into the source of the test file.

oleksandr-pavlyk (dismissed):

oleksandr-pavlyk (approved):
Thank you @diptorupd , LTGM!

libsyclinterface/dbg_build.sh
diptorupd: These changes are in the helper script, let us fix in separate PR. I just needed to get the helper script working once more to develop on libsyclinterface.

dpctl/_sycl_queue.pyx
```diff
@@ -631,34 +631,28 @@ cdef class SyclQueue(_SyclQueue):
...
                 kargs[idx] = <void*><size_t>(ctypes.addressof(arg))
-                kargty[idx] = _arg_data_type._INT
-            elif isinstance(arg, ctypes.c_uint):
+                kargty[idx] = _arg_data_type._UINT8_T
+            elif isinstance(arg, ctypes.c_short):
                 kargs[idx] = <void*><size_t>(ctypes.addressof(arg))
-                kargty[idx] = _arg_data_type._UNSIGNED_INT
-            elif isinstance(arg, ctypes.c_uint8):
+                kargty[idx] = _arg_data_type._INT16_T
+            elif isinstance(arg, ctypes.c_ushort):
```
ZzEeKkAa, line 641: Sorry for being late, but why `if instance(arg, ....)` still uses old style types?

PR #1584, opened 2024-03-06
Implement array-API's searchsorted

oleksandr-pavlyk:
This PR implements `dpctl.tensor.searchsorted` per array API 2023.12 specification (https://data-apis.org/array-api/latest/API_specification/generated/array_api.searchsorted.html).

`tensor.searchsorted` supports all data-types, including complex types, to match what `tensor.sort` and `tensor.argsort` support. 

`tensor.searchsorted` is used in implementation of set functions `unique_inverse` and `unique_all` per outstanding FIXME note.

Test file added.


dpctl/tensor/_set_functions.py
```diff
@@ -403,23 +403,9 @@ def unique_inverse(x):
...
-        pos_next = pos + counts[i]
-        _dst = inv[pos:pos_next]
-        ht_ev, _ = _full_usm_ndarray(fill_value=i, dst=_dst, sycl_queue=exec_q)
-        pos = pos_next
-        host_tasks.append(ht_ev)
-    dpctl.SyclEvent.wait_for(host_tasks)
-    return UniqueInverseResult(
-        unique_vals, dpt.reshape(inv[unsorting_ids], x.shape)
-    )
+    inv = dpt.searchsorted(unique_vals, x)
```
ndgrigorian, line 407: We may want to also avoid unnecessary synchronizations here and in `unique_all`

dpctl/tensor/_searchsorted.py
```diff
@@ -0,0 +1,172 @@
...
+    else:
+        ht_ev, _ = _searchsorted_right(
+            hay=x1,
+            needles=x2,
+            positions=dst,
+            sycl_queue=q,
+            depends=[
+                ev,
+            ],
+        )
```
ndgrigorian, line 167: I'm not sure what's better here, but something to consider is that `side` could be passed to a `searchsorted` Python API, and the internals of the lambda could pick between the `_searchsorted_left` and `_searchsorted_right` tables when calling.

This approach worked well in `repeat` and kept the Python calls a little cleaner.

ndgrigorian (approved):
Previous comment was mostly food for thought. I've verified that tests pass on CUDA as well.

LGTM, thank you @oleksandr-pavlyk!

dpctl/tensor/_searchsorted.py
oleksandr-pavlyk: Fair point. I will try this out in a subsequent PR.

PR #1585, opened 2024-03-07
Start kernel_arg_type enums from 0

ZzEeKkAa:
Fixes https://github.com/IntelPython/numba-dpex/pull/1379 and https://github.com/IntelPython/dpctl/pull/1581 . Other enums must be revisited.

### DIFF
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -88,6 +88,155 @@ __all__ = [
 _logger = logging.getLogger(__name__)
 
 
+cdef class kernel_arg_type_attribute:
+    cdef str parent_name
+    cdef str attr_name
+    cdef int attr_value
+
+    def __cinit__(self, str parent, str name, int value):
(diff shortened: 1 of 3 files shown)


oleksandr-pavlyk:
The change itself seems fine, did you test whether numerical value of enum members in Python actually changed values because of this change? Should tests be added to verify numerical values of different enum members?

ZzEeKkAa:
I've tested in numba-dpex. Definitely we need to test this and other enums in following PRs.

oleksandr-pavlyk:
I have pushed a change to reimplement `kernel_arg_type` class using Cython, to have hard binding with values of enum in C.

This change would drop coverage, and requires adding more tests, feel free to add those, otherwise I'd push changes to add more tests a bit later.

ZzEeKkAa:
LGTM! Tested localy for dpex - it works!

diptorupd (approved):
Nice improvements, four years in the waiting!

PR #1586, opened 2024-03-08
Cmake incremental builds improvement

oleksandr-pavlyk:
Avoid installing header files which are formed during building steps. 

During installation the files are not changed, but timestamps get updated prompting incremental cmake build to do full build instead.

### DIFF
--- a/libsyclinterface/CMakeLists.txt
+++ b/libsyclinterface/CMakeLists.txt
@@ -280,7 +280,7 @@ set_target_properties(DPCTLSyclInterface
 
 if (SKBUILD)
     set(_lib_destination dpctl)
-    set(_include_destination dpctl/include/syclinterface)
+    set(_include_destination ${CMAKE_INSTALL_PREFIX}/unused)
 else()
     set(_lib_destination ${CMAKE_INSTALL_PREFIX}/lib)
     set(_include_destination ${CMAKE_INSTALL_PREFIX}/include)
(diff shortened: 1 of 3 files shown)


libsyclinterface/CMakeLists.txt
```diff
@@ -280,7 +280,7 @@ set_target_properties(DPCTLSyclInterface
 
 if (SKBUILD)
     set(_lib_destination dpctl)
-    set(_include_destination dpctl/include/syclinterface)
+    set(_include_destination ${CMAKE_INSTALL_PREFIX}/unused)
```
diptorupd, line 283: What is this change doing?

oleksandr-pavlyk: Avoiding that `syclinterface` header, which are copied to `dpctl/include/syclinterface` as part of build step be overwritten by the install. The install just puts them in `_skbuild/*/cmake-install/unused` folder, and `setuptools` does not copy them into the Python layout, and instead copies files from `dpctl/include/syclinterface` like intended.

diptorupd: Then why is this step needed all  together? Will it not work, if the line is completely removed?

oleksandr-pavlyk:
I suggest we merge this to improve incremental builds, and improvements from @ZzEeKkAa in gh-1587 could be added when ready.

oleksandr-pavlyk:
The gh-1587 is coming along nicely, so this PR is largely obsolete. Keeping it around until gh-1587 gets merged though.

PR #1587, opened 2024-03-08
Depend on the source headers directly

ZzEeKkAa:
Update CMake configuration and headers to  use better target dependencies. It will unblock incremental compilation and flexible dependency management.


oleksandr-pavlyk:
Increase in coverage is due to disappearence of Cython files from coverage.

oleksandr-pavlyk:
gh-1586 does not exhibit changes in coverage metrics despite also using Cython 3.0.9. 

I also checked that C++ code generated by Cython did not materially change between versions.

Thus the change is caused by build system reorg.

oleksandr-pavlyk:
This message emitted while processing `.coverage` database points to a problem:

```
Source file is not python /home/runner/work/dpctl/dpctl/dpctl/_diagnostics.pyx
```

ZzEeKkAa:
May it be because we don't copy headers/cpp files of cython to source directory and they are always in build directory now?

oleksandr-pavlyk:
> May it be because we don't copy headers/cpp files of cython to source directory and they are always in build directory now?

Yes, indeed. See http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html

I have pushed changes to copy CXX files into layout for coverage builds.

oleksandr-pavlyk (approved):
Thank you for getting this done @ZzEeKkAa ! Looks good to me now!

oleksandr-pavlyk:
Testing clean building with and without use of `lld`, using

```bash
/usr/bin/time -f "%P %M %E" python scripts/build_locally.py --verbose
```

outputs are, on my WSL machine,

```
with lld
641% 3953024 20:24.33

with default linker
632% 4048804 20:59.47
```

ndgrigorian (approved):
I have also tested building this branch locally and it LGTM

oleksandr-pavlyk:
I have checked locally that `dpnp` build continues working as expected.

antonwolfy:
I can also confirm that `dpnp` build works with that PR without any issue.

antonwolfy (approved):

PR #1589, opened 2024-03-10
Improve coverage sycl interface

oleksandr-pavlyk:
This PR extends `test_sycl_queue_submit` to use 1D range, 2D range, 3D range and use dependent events.

### DIFF
--- a/libsyclinterface/tests/test_sycl_device_aspects.cpp
+++ b/libsyclinterface/tests/test_sycl_device_aspects.cpp
@@ -97,7 +97,7 @@ auto build_gtest_values(const std::array<std::pair<T1, T2>, N> &params)
 auto build_params()
 {
     constexpr auto param_1 = get_param_list<const char *>(
-        "opencl:gpu", "opencl:cpu", "level_zero:gpu", "host");
+        "opencl:gpu", "opencl:cpu", "level_zero:gpu");
 
     constexpr auto param_2 =
         get_param_list<std::pair<const char *, sycl::aspect>>(
--- a/libsyclinterface/tests/test_sycl_queue_submit.cpp
+++ b/libsyclinterface/tests/test_sycl_queue_submit.cpp
@@ -256,13 +256,13 @@ struct TestQueueSubmitFP64 : public ::testing::Test
     std::ifstream spirvFile;
     size_t spirvFileSize_;
     std::vector<char> spirvBuffer_;
+    DPCTLSyclDeviceRef DRef = nullptr;
     DPCTLSyclQueueRef QRef = nullptr;
     DPCTLSyclKernelBundleRef KBRef = nullptr;
 
     TestQueueSubmitFP64()
     {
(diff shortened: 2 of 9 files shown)


oleksandr-pavlyk:
Coverage info: https://coveralls.io/builds/66187084 

image

oleksandr-pavlyk:
Additional changes pushed:

- Submit for float64 should only be exercised if device has needed aspect (otherwise coverage and tests can not be run on Iris Xe)
- submit local_accessor is now done for instances with ndim=1, 2, 3 getting coverage of `dpctl_sycl_interface.cpp` to over 90%
- Added test for `DPCTLQueue_Create` when context ref is nullptr, but device is not to cover few more lines


Changes to `tests/CMakeLists.txt`:
- Created target `run-c-api-tests` to run tests with environment variable set
- Created target `llvm-cov-report` to output per-file coverage summary and `llvm-cov-show` to output per-line coverage info (handly to detect which lines are not covered)
- Removed `llvm-cov` (it is now effectively `llvm-cov-report`)
- Each `llvm-cov-report`, `llvm-cov-show` and `llvm-genhtml` depend on `run-c-api-tests`

Changes to `dbg_build.sh`:
- Added lines to set `PATH` and `LLVM_TOOLS_HOME` env. variable so that cmake can find `llvm-cov` and other programs
- Replaced `make llvm-genhtml with `make llvm-cov-report` in the script.

libsyclinterface/tests/test_sycl_device_aspects.cpp
```diff
@@ -97,7 +97,7 @@ auto build_gtest_values(const std::array<std::pair<T1, T2>, N> &params)
 auto build_params()
 {
     constexpr auto param_1 = get_param_list<const char *>(
-        "opencl:gpu", "opencl:cpu", "level_zero:gpu", "host");
```
diptorupd, line 100: Not for this PR, but the design where a list of filter string is used is an anti-pattern. I got rid of a similar pattern in numba-dpex unit tests last year. All tests should run on default device and we can control the default device using the ONEAPI_DEVICE_SELECTOR environment variable.

libsyclinterface/tests/test_sycl_queue_submit.cpp
```diff
@@ -334,9 +356,11 @@ TEST_F(TestQueueSubmit, CheckForFloat)
 
 TEST_F(TestQueueSubmitFP64, CheckForDouble)
 {
-    submit_kernel<double>(QRef, KBRef, spirvBuffer_, spirvFileSize_,
-                          DPCTLKernelArgType::DPCTL_FLOAT64_T,
-                          "_ZTS11RangeKernelIdE");
+    if (DPCTLDevice_HasAspect(DRef, DPCTLSyclAspectType::fp64)) {
```
diptorupd, line 359: Good catch

diptorupd (approved):

PR #1595, opened 2024-03-18
Avoid superfluous copy of dpctl_capi struct

oleksandr-pavlyk:
This PR eliminates superfluous copying of `dpctl_capi` instance, even if it is later eliminated by the compiler. 

The copying would necessary in `static dpctl_capi api = lookup()` line, since `lookup` returns `dpctl_capi` instance by value.
The change is to make it return a reference.

Since `dpctl_capi` is a singleton, this change only eliminates a single copy, so consider it a technical debt cleanup change.

The PR also adds a default move assignment operator for `dpctl_capi`.


ndgrigorian (approved):
LGTM!

PR #1602, opened 2024-03-23
Implements accumulation functions in dpctl.tensor

ndgrigorian:
This pull request proposes implementation of `dpctl.tensor.cumulative_sum`, `dpctl.tensor.cumulative_prod`, and `dpctl.tensor.cumulative_logsumexp`.

`cumulative_sum` is already part of the array API standard and is implemented as per the spec. `cumulative_prod` and `cumulative_logsumexp` are implemented with a similar API, including an `include_initial` keyword argument.

### DIFF
--- a/dpctl/tensor/_reduction.py
+++ b/dpctl/tensor/_reduction.py
@@ -164,6 +164,7 @@ def _reduction_over_axis(
                 sycl_queue=q,
                 depends=[cpy_e],
             )
+            host_tasks_list.append(ht_e_red)
             ht_e_cpy2, _ = ti._copy_usm_ndarray_into_usm_ndarray(
                 src=tmp_res, dst=res, sycl_queue=q, depends=[r_e]
             )
--- a/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
@@ -678,7 +678,7 @@ sycl::event inclusive_scan_iter(sycl::queue &exec_q,
             outputT *local_scans = stack_elem.get_local_scans_ptr();
             size_t local_stride = stack_elem.get_local_stride();
 
-            nwiT constexpr updates_per_wi = n_wi;
+            constexpr nwiT updates_per_wi = n_wi;
             size_t update_nelems =
                 ceiling_quotient<size_t>(src_size, updates_per_wi);
 
@@ -730,7 +730,7 @@ sycl::event inclusive_scan_iter(sycl::queue &exec_q,
             outputT *local_scans = stack_elem.get_local_scans_ptr();
             size_t local_stride = stack_elem.get_local_stride();
 
-            nwiT constexpr updates_per_wi = n_wi;
+            constexpr nwiT updates_per_wi = n_wi;
             size_t update_nelems =
                 ceiling_quotient<size_t>(src_size, updates_per_wi);
 
--- a/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
@@ -113,11 +113,6 @@ public:
         return src_;
     }
 
-    const T *get_src_const_ptr() const
-    {
-        return src_;
-    }
-
     size_t get_size() const
     {
         return size_;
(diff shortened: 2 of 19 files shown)


dpctl/tensor/_reduction.py
```diff
@@ -164,6 +164,7 @@ def _reduction_over_axis(
                 sycl_queue=q,
                 depends=[cpy_e],
             )
+            host_tasks_list.append(ht_e_red)
```
oleksandr-pavlyk, line 167: Good catch!

dpctl/tensor/libtensor/include/kernels/accumulators.hpp
```diff
@@ -66,184 +68,312 @@ template <typename inputT, typename outputT> struct NonZeroIndicator
...
+                }
+                else {
+                    // shift input to the left by a single element relative to
+                    // output
+                    local_iscan[m_wi] =
+                        (i + m_wi < acc_nelems && i + m_wi > 0)
+                            ? transformer(
+                                  input[inp_iter_offset +
+                                        inp_indexer((s0 + s1 * (i + m_wi)) -
+                                                    1)])
```
oleksandr-pavlyk, line 281: Can `s0` and `s1` ever be zeros?

ndgrigorian: No, it shouldn't be possible, and especially not where `include_initial` is true.

They're first set (as constexpr) to 0 and 1 respectively, then later, `chunk_size - 1` and `chunk_size` if more than 1 group is necessary. Since `chunk_size` is `wg_size * n_wi`, they can never both be 0.

dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
```diff
@@ -154,6 +154,30 @@ T custom_reduce_over_group(const GroupT &wg,
...
+    sycl::group_barrier(wg, sycl::memory_scope::work_group);
+
+    if (wg.leader()) {
+        for (size_t i = 1; i < wgs; ++i) {
+            local_mem_acc[i] = op(local_mem_acc[i], local_mem_acc[i - 1]);
+        }
+    }
+
+    T accumulated_local_val = local_mem_acc[local_id];
+    sycl::group_barrier(wg, sycl::memory_scope::work_group);
```
oleksandr-pavlyk, line 176:
```suggestion
    if (wg.leader()) {
        T prev = local_mem_acc[0];
        for (size_t i = 1; i < wgs; ++i) {
            prev = op(local_mem_acc[i], prev]);
            local_mem_acc[i] = prev;
        }
    }
    
    // ensure all work-items see the same SLM that leader updated
    sycl::group_barrier(wg, sycl::memory_scope::work_group);
    T accumulated_local_val = local_mem_acc[local_id];
    // ensure all work-items updated the local variable
    sycl::group_barrier(wg, sycl::memory_scope::work_group);
```

oleksandr-pavlyk: I feel like it is more appropriate to move the second `group_barrier` out of this function to after its call site.

Use of private variable `scan_val` helps do more work in registers, reducing bandwidth of SLM access,

dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
```diff
@@ -154,6 +154,30 @@ T custom_reduce_over_group(const GroupT &wg,
...
 }
 
+template <typename T, typename GroupT, typename LocAccT, typename OpT>
+T custom_inclusive_scan_over_group(const GroupT &wg,
+                                   LocAccT local_mem_acc,
+                                   const T local_val,
+                                   const OpT &op)
+{
+    auto local_id = wg.get_local_id(0);
+    auto wgs = wg.get_local_range(0);
```
oleksandr-pavlyk, line 164:
```suggestion
    const std::uint32_t local_id = wg.get_local_id(0);
    const std::uint32_twgs = wg.get_local_range(0);
```

And the `for` loop private variable need to also be `std::uint32_t`.

oleksandr-pavlyk (approved):
LGMT! Thank you @ndgrigorian

PR #1610, opened 2024-03-27
Fixes `out` kwarg in `matmul` when axes are appended to inputs

ndgrigorian:
This PR corrects behavior of `matmul` to properly account for appending of axes to inputs where one or both arguments are vectors.

Previously, these cases would raise an error, and require an `out` array with axes appended by the user.


oleksandr-pavlyk (approved):
LGTM! Thank you @ndgrigorian !

PR #1614, opened 2024-03-27
Fix main script

oleksandr-pavlyk:
Output of `python -m dpctl --library` incorrectly output `-lSyclInterface`. It should have been `-lDPCTLSyclInterface`. 

This PR changes the output, adjusts docstrings to use correct library name and modifies test to assert that the expected library name occurs as a substring in the output.

This fix should be backported to 0.16.1


ndgrigorian (approved):
Break in examples is unrelated, this works as expected for me locally, so LGTM!

PR #1621, opened 2024-03-29
Update reduction data types for 2023.12 array API specification, update `__array_api_version__`

ndgrigorian:
This pull request updates reductions over floating point data to be performed in the input type by default as per the 2023.12 array API spec.

This pull request also changes `__array_api_version__` to 2023.12. As `cumulative_sum` is a requirement of the 2023.12 spec, this PR should be redirected to master after the base branch is merged.

### DIFF
--- a/dpctl/tensor/_accumulation.py
+++ b/dpctl/tensor/_accumulation.py
@@ -20,51 +20,12 @@ import dpctl
 import dpctl.tensor as dpt
 import dpctl.tensor._tensor_accumulation_impl as tai
 import dpctl.tensor._tensor_impl as ti
-from dpctl.tensor._type_utils import _to_device_supported_dtype
+from dpctl.tensor._type_utils import (
+    _default_accumulation_dtype,
+    _default_accumulation_dtype_fp_types,
+    _to_device_supported_dtype,
+)
 from dpctl.utils import ExecutionPlacementError
-
-
-def _default_accumulation_dtype(inp_dt, q):
-    """Gives default output data type for given input data
-    type `inp_dt` when accumulation is performed on queue `q`
-    """
-    inp_kind = inp_dt.kind
-    if inp_kind in "bi":
-        res_dt = dpt.dtype(ti.default_device_int_type(q))
-        if inp_dt.itemsize > res_dt.itemsize:
-            res_dt = inp_dt
-    elif inp_kind in "u":
-        res_dt = dpt.dtype(ti.default_device_int_type(q).upper())
-        res_ii = dpt.iinfo(res_dt)
-        inp_ii = dpt.iinfo(inp_dt)
-        if inp_ii.min >= res_ii.min and inp_ii.max <= res_ii.max:
-            pass
-        else:
-            res_dt = inp_dt
-    elif inp_kind in "fc":
-        res_dt = inp_dt
-
-    return res_dt
-
-
-def _default_accumulation_dtype_fp_types(inp_dt, q):
-    """Gives default output data type for given input data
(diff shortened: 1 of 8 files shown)


oleksandr-pavlyk:
I think the spec changed default output dtype of reduction operations only for the floating point types.

For integrals previous behavior is correct, but I will double-check.

oleksandr-pavlyk:
> I think the spec changed default output dtype of reduction operations only for the floating point input dtypes.
> 
> For integrals previous behavior is correct, but I will double-check.

Yes, the change to `dtype=None` behavior was made in data-apis/array-api#744, The PR works as specified though:

```
In [13]: def check(dt): return dpt.sum(dpt.zeros((10**3, 10**2), dtype=dt, device="cpu"), axis=0).dtype

In [14]: { inp:check(ty) for inp, ty in dpt.__array_namespace_info__().dtypes().items()}
Out[14]:
{'bool': dtype('int64'),
 'float32': dtype('float32'),
 'complex64': dtype('complex64'),
 'complex128': dtype('complex128'),
 'int8': dtype('int64'),
 'int16': dtype('int64'),
 'int32': dtype('int64'),
 'int64': dtype('int64'),
 'uint8': dtype('uint64'),
 'uint16': dtype('uint64'),
 'uint32': dtype('uint64'),
 'uint64': dtype('uint64')}
```

oleksandr-pavlyk:
The "array_api_tests/test_special_cases.py::test_unary[signbit(x_i is NaN) -> True]" failure seems like a bug in the test suite.

This is the output of ` print(case, ": ", x.shape, x.strides, x._element_offset, x.dtype, x.device, res.shape, res.dtype, res.device, x, res, idx, hex(int(x_v[idx])), in_, out, binary(in_))` added to the `test_unary` function:

```
x_i is NaN -> True :  () () 0 float32 Device(opencl:cpu:0) () bool Device(opencl:cpu:0) nan False () 0x7fc00000 nan 0.0 01111111110000000000000000000000
```

The "0x7fc00000" is hexadecimal view into float32, and it indeed does not have the sign bit set.

I filed https://github.com/data-apis/array-api-tests/issues/250

ndgrigorian:
> The "array_api_tests/test_special_cases.py::test_unary[signbit(x_i is NaN) -> True]" failure seems like a bug in the test suite.
> 
> This is the output of ` print(case, ": ", x.shape, x.strides, x._element_offset, x.dtype, x.device, res.shape, res.dtype, res.device, x, res, idx, hex(int(x_v[idx])), in_, out, binary(in_))` added to the `test_unary` function:
> 
> ```
> x_i is NaN -> True :  () () 0 float32 Device(opencl:cpu:0) () bool Device(opencl:cpu:0) nan False () 0x7fc00000 nan 0.0 01111111110000000000000000000000
> ```
> 
> The "0x7fc00000" is hexadecimal view into float32, and it indeed does not have the sign bit set.
> 
> I filed [data-apis/array-api-tests#250](https://github.com/data-apis/array-api-tests/issues/250)

Yes, there are a couple more cases too: the `copysign` failure also seems to be incorrect, both inputs appear to be zero but the output is expected to be NaN, and the `cumulative_sum` test failure comes from an assumption that `cumulative_sum(M)` for `M` size 1 and 1D should return an array with empty shape (i.e., a scalar).

The `prod` failure is a result of a test not yet updated for the 2023.12 spec.

If these failures are still around come time to merge this branch, I will introduce them to the array API test skips file.

oleksandr-pavlyk (approved):
For the a future PR, we could parametrize default dtype functions to take callables that return needed default type given `q`. In tests we can use mock functions to reach branches left out by CI runs.

PR #1628, opened 2024-04-01
Correction to `dpctl.tensor.tile` for scalar input and empty repetitions

### LINKED ISSUE #1627: `tile` does not handle empty repetitions and scalar input as per array API
When `tile` is called on a scalar array with empty repetitions, it returns a 1D array instead of 0D (scalar)

```
>>> import dpctl,dpctl.tensor as dpt, numpy as np
>>> dpt.tile(dpt.asarray(5), ()).shape
(1,)
 
>>> np.tile(np.array(5), ()).shape
()
```

ndgrigorian:
This PR fixes a bug in `tile` where if the input is 0D and `repetitions` is empty, instead of a copying a scalar input and returning a scalar, a 1D array of size 1 would be returned instead.

Per the [array API spec](https://data-apis.org/array-api/latest/API_specification/generated/array_api.tile.html) and Numpy, this is incorrect behavior.

Resolves #1627 

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -964,18 +964,6 @@ def tile(x, repetitions):
                 f"Expected tuple or integer type, got {type(repetitions)}."
             )
 
-    # case of scalar
-    if x.size == 1:
-        if not repetitions:
-            # handle empty tuple
-            repetitions = (1,)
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk (approved):
Look good to me! Thank you @ndgrigorian

PR #1629, opened 2024-04-01
Edit docstrings for `tile` and `repeat`, fix "usm_narray" typos

ndgrigorian:
This PR changes docstrings for `tile` and `repeat` to better describe behaviors of these functions, and fixes minor typos throughout dpctl where "usm_narray" was used instead of "usm_ndarray"

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -725,7 +725,7 @@ def repeat(x, repeats, /, *, axis=None):
         repeats (Union[int, Sequence[int, ...], usm_ndarray]):
             The number of repetitions for each element.
 
-            `repeats` must be broadcast compatible with `N` where `N` is
+            `repeats` must be broadcast-compatible with `N` where `N` is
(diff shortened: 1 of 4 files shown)


dpctl/tensor/_manipulation_functions.py
```diff
@@ -717,32 +717,38 @@ def swapaxes(X, axis1, axis2):
...
+    Repeat elements of an array on a per-element basis.
 
     Args:
         x (usm_ndarray): input array
 
         repeats (Union[int, Sequence[int, ...], usm_ndarray]):
             The number of repetitions for each element.
-            `repeats` is broadcast to fit the shape of the given axis.
+
+            `repeats` must be broadcast compatible with `N` where `N` is
```
oleksandr-pavlyk, line 728: Should it be "broadcast-compatible"?

ndgrigorian: Yes, that would be more correct.

oleksandr-pavlyk (approved):
Thank you for finding and fixing these @ndgrigorian !

PR #1630, opened 2024-04-02
Fix compiling for CUDA with 2024.1 compiler

ndgrigorian:
Using the 2024.1 compiler and building with `DPCTL_TARGET_CUDA=1` fails due to unrecognized symbols in elementwise functions like `fabs`, `fabsf`, `copysign`, and `floor` from the C++ std namespace.

This PR proposes replacing the non-working symbols with their Sycl namespace equivalents.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/elementwise_functions/cabs_impl.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/elementwise_functions/cabs_impl.hpp
@@ -1,0 +1,84 @@
+//===------- cabs_impl.hpp - Implementation of cabs  -------*-C++-*/===//
+//
+//                      Data Parallel Control (dpctl)
+//
+// Copyright 2020-2024 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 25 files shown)


ndgrigorian:
@oleksandr-pavlyk 
This PR may also be a good opportunity to test removing the work-around implemented in `proj.hpp` for the previous compiler.

oleksandr-pavlyk:
I have pushed a commit to remove the work-around. Local tests passed.

dpctl/tensor/libtensor/include/kernels/elementwise_functions/cabs_impl.hpp
```diff
@@ -0,0 +1,84 @@
...
+        return q_nan;
+    }
+    else if (std::isnan(y)) {
+        return q_nan;
+    }
+    else {
+#ifdef USE_SYCL_FOR_COMPLEX_TYPES
+        return exprm_ns::abs(exprm_ns::complex<realT>(z));
+#else
+        return std::hypot(std::real(z), std::imag(z));
```
oleksandr-pavlyk, line 76:
```suggestion
        return std::hypot(x, y);
```

oleksandr-pavlyk (approved):

PR #1645, opened 2024-04-18
Avoid sycl::queue copying in libtensor

oleksandr-pavlyk:
Changed signatures functions to take ``sycl::queue `` argument by reference, not by value.

Since ``sycl::queue`` contains a shared pointer inside, copying it involves operating atomic variable, and hence should be avoided if possible.


ndgrigorian (approved):
LGTM!

PR #1650, opened 2024-04-22
Fixes element-wise comparisons of mixed signed-unsigned integer inputs

ndgrigorian:
This pull request addresses issues in comparison functions with mixed kinds (signed and unsigned) by explicitly adding overloads for combinations of `uint64` and `int64` and adding a unique weak-type resolver function for the comparisons.

The resolver function guarantees that negative Python scalars are not used to initialize an array with an unsigned integer data type in the binary comparisons, which makes comparisons between negative Python scalars and unsigned integer arrays work as expected.

Old behavior:
```

In [2]: x = dpt.arange(10, dtype="u8")

In [3]: -1 < x
Out[3]:
              True,  True])
```

Now:
```
In [3]: -1 < x
Out[3]:
usm_ndarray([False, False, False, False, False, False, False, False,
             False, False])
```

Also slips in a change removing an unnecessary overload to `not_equal` for mixed float and complex float data types, which are safe for basic casting and not present in other elementwise comparisons.


ndgrigorian:
@oleksandr-pavlyk 
This PR does not yet address similar issues in `clip`, `minimum`, and `maximum`.

For `minimum` and `maximum` a similar solution for negative Python scalars is possible. `clip` may be a bit trickier.

dpctl/tensor/_elementwise_common.py
```diff
@@ -476,6 +481,22 @@ def get_type_promotion_path_acceptance_function(self):
...
+
+        Resolver influences what type the scalar will be
+        treated as prior to type promotion behavior.
+        The function takes 3 arguments:
+            o1_dtype - A class representing a Python scalar type or a dtype
+            o2_dtype - A class representing a Python scalar type or a dtype
+            sycl_dev - :class:`dpctl.SyclDevice` on which function evaluation
+                is carried out.
+
+        One of o1_dtype and o2_dtype must be a dtype
```
oleksandr-pavlyk, line 496:
```suggestion
        The function takes 3 arguments:
        
        Args:
            o1_dtype (object, dtype):
                A class representing a Python scalar type or a ``dtype``
            o2_dtype (object, dtype): 
                A class representing a Python scalar type or a ``dtype``
            sycl_dev (:class:`dpctl.SyclDevice`):
                Device on which function evaluation is carried out.

        One of ``o1_dtype`` and ``o2_dtype`` must be a ``dtype`` instance.
```

dpctl/tensor/libtensor/include/kernels/elementwise_functions/equal.hpp
```diff
@@ -151,6 +169,10 @@ template <typename T1, typename T2> struct EqualOutputType
                                         bool>,
         td_ns::
             BinaryTypeMapResultEntry<T1, std::int64_t, T2, std::int64_t, bool>,
+        td_ns::
+            BinaryTypeMapResultEntry<T1, std::uint64_t, T2, std::int64_t, bool>,
+        td_ns::
+            BinaryTypeMapResultEntry<T1, std::int64_t, T2, std::uint64_t, bool>,
```
oleksandr-pavlyk, line 175: Nice that we can get away with only instantiating int64/uint64 entries in the 4x4 support matrix!

dpctl/tensor/_type_utils.py
```diff
@@ -400,6 +400,72 @@ def _resolve_weak_types(o1_dtype, o2_dtype, dev):
...
 
 
+def _resolve_weak_types_comparisons(o1_dtype, o2_dtype, dev):
+    "Resolves weak data type per NEP-0050 for comparisons,"
+    "where result type is known to be `bool` and special behavior"
+    "is needed to handle mixed integer kinds"
+    if isinstance(
+        o1_dtype,
+        (WeakBooleanType, WeakIntegralType, WeakFloatingType, WeakComplexType),
+    ):
```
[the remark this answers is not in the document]
ndgrigorian: I agree, that would be a good change

ndgrigorian:
@oleksandr-pavlyk 
I've added tests and added the `_is_weak_dtype` utility function.

As of Numpy 2, Numpy will no longer permit converting Python integers to integer arrays where the Python integer is out-of-bounds for the integer array's data type (at least in ufuncs). This is how they resolve the issues with `minimum`, `maximum`, and `clip` with unsigned integer arrays and negative Python integers.

Imitating that behavior would align with the future of Numpy (and therefore should be acceptable to DPNP) but will take a little more work. Therefore, I will handle it in a separate PR, and this is ready for review pending CI success.

oleksandr-pavlyk (approved):
Thank you @ndgrigorian ! Looks good to me. Local testing worked well too.

PR #1384, opened 2023-09-01
Fix dpctl.tensor.put race condition

PiotrekB416:
As shown in #1360 `dpctl.tensor.put` has a race condition if `indices` has non-unique values
This PR solves this issue by removing unnecessary indices 

- [x] Have you provided a meaningful PR description?
- [x] Have you added a test, reproducer or referred to an issue with a reproducer?
- [ ] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you opening the PR as a draft?

Issue #1360, opened 2023-08-21
dpctl.tensor.put returns incorrect result 


vlad-perevezentsev:
The code below works incorrectly in dpctl when `dpctl.tensor.put` needs to rewrite the duplicate index of an element with a different value

```
import dpctl.tensor as dpt

x = dpt.ones((5,))
ind = dpt.asarray([0, 0])

dpt.put(x, ind, [10, 20])
x
> usm_ndarray([10.,  1.,  1.,  1.,  1.])

import numpy
a = numpy.ones((5,))
numpy.put(a, [0, 0], [10, 20])
a
> array([20.,  1.,  1.,  1.,  1.])

```

oleksandr-pavlyk:
`put` parallelizes assignments, and the present input results in race condition.

oleksandr-pavlyk:
gh-1653 adds a docstring note for ``dpt.put`` warning of the race condition, and provides sample code for resolving it for 1D case.

This addition resolves this issue, which we intend to closed in a few days.

oleksandr-pavlyk:
https://github.com/IntelPython/dpctl/pull/1653 adds a docstring note for dpt.put warning of the race condition, and provides sample code for resolving it for 1D case.

This addition resolves this issue, which we intend to closed in a few days.

ndgrigorian:
Closing this issue as resolved now that #1653 is in and #1384 is closed

PR #1653, opened 2024-04-23
Addresses gh-1360 with documentation

oleksandr-pavlyk:
Document race condition in put when indices array contain duplicated. Provide an example of resolving race condition for 1-d ``x``, ``indices``, and ``vals``.

This PR supersedes gh-1384, since support for `unique_only` can not be efficiently support for higher rank arrays with ``axes`` specified.

### DIFF
--- a/dpctl/tensor/_indexing_functions.py
+++ b/dpctl/tensor/_indexing_functions.py
@@ -136,12 +136,6 @@ def put(x, indices, vals, /, *, axis=None, mode="wrap"):
             The array the values will be put into.
         indices (usm_ndarray):
             One-dimensional array of indices.
-
-            Note that if indices are not unique, a race
-            condition will result, and the value written to
-            ``x`` will not be deterministic.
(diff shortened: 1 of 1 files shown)


dpctl/tensor/_indexing_functions.py
```diff
@@ -128,30 +132,76 @@ def put(x, indices, vals, /, *, axis=None, mode="wrap"):
...
+        x (usm_ndarray):
+            The array the values will be put into.
+        indices (usm_ndarray):
+            One-dimensional array of indices.
+
+            Note that if indices are not unique, a race
+            condition will result, and the value written to
+            ``x`` will not be deterministic.
+            :func:`dpctl.tensor.unique` can be used to
+            guarantee unique elements in ``indices``.
```
ndgrigorian, line 144: With this change, this note seems redundant. Perhaps it would be best to remove it.

ndgrigorian (approved):
This LGTM, good to have this example!

PR #1655, opened 2024-04-25
pybind11 type caster for `sycl::half`

ndgrigorian:
This pull request proposes removal of `PythonObjectUnboxer` in favor of a pybind11 type caster for `sycl::half`. This permits extensions to convert Python scalars to `sycl::half` with only the inclusion of `dpctl4pybind11.hpp`.


dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -671,6 +671,54 @@ struct type_caster<sycl::kernel_bundle<sycl::bundle_state::executable>>
...
+        PyObject *source = src.ptr();
+
+        if (convert || PyFloat_Check(source)) {
+            py_value = PyFloat_AsDouble(source);
+        }
+        else {
+            return false;
+        }
+
+        bool py_err = py_value == double(-1) && PyErr_Occurred();
```
oleksandr-pavlyk, line 699:
```suggestion
        bool py_err = (py_value == double(-1)) && PyErr_Occurred();
```

I'd prefer to keep the parenthesis in place, that way there is no guessing while reading the code

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -671,6 +671,54 @@ struct type_caster<sycl::kernel_bundle<sycl::bundle_state::executable>>
...
+        }
+        else {
+            return false;
+        }
+
+        bool py_err = (py_value == double(-1)) && PyErr_Occurred();
+
+        if (py_err) {
+            PyErr_Clear();
+            if (py_err && convert && (PyNumber_Check(source) != 0)) {
```
oleksandr-pavlyk, line 703:
```suggestion
            if (convert && (PyNumber_Check(source) != 0)) {
```

We checked thar ``py_err`` has value of `true`.

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -671,6 +671,54 @@ struct type_caster<sycl::kernel_bundle<sycl::bundle_state::executable>>
...
+                auto tmp = reinterpret_steal<object>(PyNumber_Float(source));
+                PyErr_Clear();
```
oleksandr-pavlyk, line 705: Do you know why this second call to ``PyErr_Clear`` is needed?

ndgrigorian: Yes, and it isn't necessary here. Casting Python complex to int specifically can fail, so it would be necessary for a more general caster. Here I believe it can be removed.

oleksandr-pavlyk (approved):

PR #1656, opened 2024-04-26
Expanded intel_device_info output

oleksandr-pavlyk:
Added `"free_memory"` for Free Global Memory on the device in bytes, `"memory_clock_rate"` for Maximum Memory Clock Rate in MHz, and `"memory_bus_width"` for bus width in bits.

Additions come from https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_device_info.md

### DIFF
--- a/dpctl/utils/src/device_queries.cpp
+++ b/dpctl/utils/src/device_queries.cpp
@@ -168,7 +168,7 @@ PYBIND11_MODULE(_device_queries, m)
           py::arg("device"));
 
     m.def("intel_device_info_free_memory", &py_intel_free_memory,
-          "Returns the memory avialble on the device in units of bytes.",
+          "Returns the memory available on the device in units of bytes.",
           py::arg("device"));
 
     m.def("intel_device_info_memory_clock_rate", &py_intel_memory_clock_rate,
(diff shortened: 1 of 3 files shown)


dpctl/utils/src/device_queries.cpp
```diff
@@ -136,4 +166,16 @@ PYBIND11_MODULE(_device_queries, m)
     m.def("intel_device_info_max_mem_bandwidth", &py_intel_max_mem_bandwidth,
           "Returns the maximum memory bandwidth in units of bytes/second.",
           py::arg("device"));
+
+    m.def("intel_device_info_free_memory", &py_intel_free_memory,
+          "Returns the memory avialble on the device in units of bytes.",
```
ndgrigorian, line 171:
```suggestion
          "Returns the memory available on the device in units of bytes.",
```
Nit

ndgrigorian (commented):
Since we verified this worked on PVC with `ZES_ENABLE_SYSMAN=1`, it may be a good idea to document that this must be set to get the `ext_intel_free_memory` descriptor.

ndgrigorian (approved):

PR #1658, opened 2024-05-01
Changes test_func_order tests for some unary functions

oleksandr-pavlyk:
The change is to iterate over permutations first, doing some work that does not depend on the order, and then iterate over orders.

This should save doing some computations 4 times (once for each order), now only done once.

The change shaves 4 seconds off of execution of entire test suite on CPU device (from 5 minutes 6 seconds to 5 minutes and 2 seconds).


ndgrigorian (approved):
LGTM!

PR #1660, opened 2024-05-02
Metadata improvements

ZzEeKkAa:
Changes in this PR:
- Move all supported configuration to `pyproject.toml` from `setup.py` and `setup.cfg` (latest one was removed).
- Populate additional fields to `project` and `build-system` sections of `pyproject.toml`.
- Use build packages from pyproject's `build-system` section in the conda `meta.yaml` to keep it in sync.
- Use package managed `versioneer`.

Problems:
- wheel package removes symbolic links

Checklist:


pyproject.toml
```diff
@@ -1,52 +1,141 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+  # TODO: keep in sync with [project.dependencies]
+  "wheel>=0.43",
+  "build>=1.1",
+  "setuptools>=63.0.0",
+  "scikit-build>=0.17.0",
+  "ninja>=1.11.1; platform_system!='Windows'",
```
oleksandr-pavlyk, line 9: We used `ninja` on Windows before, what is the reason for moving away from `ninja`? What build system replaces `ninja` on Windows?

pyproject.toml
```diff
@@ -1,52 +1,141 @@
...
+[project.optional-dependencies]
+coverage = ["Cython", "pytest", "pytest-cov", "coverage", "tomli"]
+docs = [
+  "Cython",
+  "sphinx",
+  "sphinx_rtd_theme",
+  "pydot",
+  "graphviz",
+  "sphinxcontrib-programoutput"
+]
```
oleksandr-pavlyk, line 70: This is missing a bunch of required packages. See `generate-docs.yml` for the list

pyproject.toml
ZzEeKkAa: On Windows Ninja is distributed with MSVC

pyproject.toml
ZzEeKkAa: Thank you! I just took this list from setup.py

conda-recipe/build.sh
```diff
@@ -11,18 +11,37 @@ export ICXCFG="$(pwd)/icpx_for_conda.cfg"
...
+export CMAKE_GENERATOR=Ninja
+# Make CMake verbose
+export VERBOSE=1
+
+CMAKE_ARGS="${CMAKE_ARGS} -DDPCTL_LEVEL_ZERO_INCLUDE_DIR=${PREFIX}/include/level_zero"
+
+# -wnx flags mean: --wheel --no-isolation --skip-dependency-check
+${PYTHON} -m build -w -n -x
+${PYTHON} -m wheel tags --remove --build "$GIT_DESCRIBE_NUMBER" \
+    --platform-tag manylinux2014_x86_64 dist/dpctl*.whl
```
oleksandr-pavlyk, line 27: I think we need `manylinux_2_28` or some such. CC @ekomarova

conda-recipe/build.sh
```diff
@@ -11,18 +11,37 @@ export ICXCFG="$(pwd)/icpx_for_conda.cfg"
...
+export CXX=icpx
+
+export CMAKE_GENERATOR=Ninja
+# Make CMake verbose
+export VERBOSE=1
+
+CMAKE_ARGS="${CMAKE_ARGS} -DDPCTL_LEVEL_ZERO_INCLUDE_DIR=${PREFIX}/include/level_zero"
+
+# -wnx flags mean: --wheel --no-isolation --skip-dependency-check
+${PYTHON} -m build -w -n -x
```
oleksandr-pavlyk, line 25: Can we use `python -m wheel unpack`, remove copies of `libDPCTLSyclInterface.so*` and pack it back?

ZzEeKkAa: Then numba_dpex won't link to those libraries for wheel install.

conda-recipe/build.sh
ZzEeKkAa: Updated. Set to pick up glibc version from `linux_headers` conda package version.

pyproject.toml
ZzEeKkAa: It is not in use for now. Get use of it could be added in future PRs, since this PR does not affect that functionality.

oleksandr-pavlyk:
Removal of work-around kludge in `setup.py` that ensured symbolic links created by `cmake` stay symbolic in conda package is a change int the right direction, as the ecosystem moves to new generation packaging framework. But the change turns symbolic links into hard copies of shared objects, ballooning the size of both the conda package and wheel package. 

I propose to modify `build.sh` to run `python -m wheel unpack wheel_dir`, remove duplicate copies of the library, e.g., `libDPCTLSyclInterface.so` and `libDPCTLSyclInterface.so.0.18` while recording their names in script variables. 

We would then `python -m wheel pack wheel_dir`, retag it, and run `pip install` to install into `$PREFIX`. We would them post-process the content of installed `site-packages/dpctl` to move actual `libDPCTLSyclInterface.so.0` to `libDPCTLSyclInterface.so.0.18` (we saved the filename earlier), and create symbolic links `libDPCTLSyclInterface.so.0 -> libDPCTLSyclInterface.so.0.18` and `libDPCTLSyclInterface.so -> libDPCTLSyclInterface.so.0`.

This way conda package would contain symbolic links as it did previously.

oleksandr-pavlyk:
Assuming my PR gh-1706 into this branch gets merged, the set of changes is a big step forward. Thank you @ZzEeKkAa for doing this work!

oleksandr-pavlyk (approved):
Thank you @ZzEeKkAa !

conda-recipe/build.sh
```diff
@@ -8,21 +8,69 @@ echo "--gcc-toolchain=${BUILD_PREFIX} --sysroot=${BUILD_PREFIX}/${HOST}/sysroot
...
+export soxxx_name=$(basename ${soxxx_full_path})
+export wheel_path=$(dirname $(dirname ${so_full_path}))
+
+# deal with hard copies
+${PYTHON} -m wheel pack ${wheel_path}
+
+rm -rf dpctl_wheel
+popd
+
+${PYTHON} -m wheel tags --remove --build "$GIT_DESCRIBE_NUMBER" \
```
ekomarova, line 50: What is the purpose of adding a build number here? In fact, the internal CI has its own logic for adding a build number

ekomarova, line 50: What is the purpose of adding a build number here? In fact, the internal CI has its own logic for adding a build number

conda-recipe/build.sh
```diff
@@ -8,21 +8,69 @@ echo "--gcc-toolchain=${BUILD_PREFIX} --sysroot=${BUILD_PREFIX}/${HOST}/sysroot
...
+
+# Recover symbolic links
+# libDPCTLSyclInterface.so.0 -> libDPCTLSyclInterface.so.0.17
+# libDPCTLSyclInterface.so -> libDPCTLSyclInterface.so.0
+mv ${libdir}/${sox_name} ${libdir}/${soxxx_name}
+ln -s ${libdir}/${soxxx_name} ${libdir}/${sox_name}
+ln -s ${libdir}/${sox_name} ${libdir}/${so_name}
+
+# Copy wheel package
+if [[ -v WHEELS_OUTPUT_FOLDER ]]; then
```
ekomarova, line 72: This condition is incorrect for conda building, since the variable WHEELS_OUTPUT_FOLDER is always declared (it can have zero or non-zero length in the case of wheels building), and even in the absence of wheel package, we will still get into this condition, which leads to the build problem on linux for conda. We should not check only the existence of WHEELS_OUTPUT_FOLDER , but that WHEELS_OUTPUT_FOLDER has a non-zero length, as it was earlier:
```
if [ -n "${WHEELS_OUTPUT_FOLDER}" ]; then
```

conda-recipe/build.sh
oleksandr-pavlyk: Public CI builds and upload wheels to dppy/label/dev channel. Adding build numbers here allows to distinguish between different wheels in the repo

conda-recipe/build.sh
oleksandr-pavlyk: I used `[[ -d ${WHEELS_OUTPUT_FOLDER}" ]]` instead which is even stronger. Not only should the string be non-empty, it must contain path to existing folder.

oleksandr-pavlyk (approved):

ekomarova (approved):
LGTM!
The problems with testing wheels are not related to this PR, and will be fixed later by infra team

PR #1665, opened 2024-05-06
Initialize `UAR_BYTE_` and `UAR_UBYTE_` in `dpctl_capi`

ndgrigorian:
This PR fixes two uninitialized typenums in `dpctl_capi` for `UAR_BYTE_` and `UAR_UBYTE_`.

It was noticed during extension writing that these variables are initialized to -1 and never set to the proper values.


oleksandr-pavlyk (approved):
Good catch!

PR #1666, opened 2024-05-06
Update wheel tag with pep600

ekomarova:
This PR is designed to fix the problem described here: [SAT-6873](https://jira.devtools.intel.com/browse/SAT-6873). I did some experiments and found out that using pep600 wheel tag prevents the wheel from being installed on non-supported platforms


oleksandr-pavlyk (approved):
LGTM! Thank you @ekomarova. 

I have not tested this change myself yet. I assume testing was part performed while working out particulars of this change.

PR #1667, opened 2024-05-09
Update dlpack to v1.0rc

ndgrigorian:
This PR bumps the version of dlpack by bringing in the new `dlpack.h` from v1.0rc.

As a result of changes introduced in v1.0rc
- `get_build_dlpack_version` now returns a tuple giving the major and minor DLPack version dpctl was built with
- support for the new `DLManagedTensorVersioned` struct has been added, including full support of the new `flags` attribute
- `__dlpack__` method of `usm_ndarray` has been updated to choose between `DLManagedTensorVersioned` and `DLManagedTensor` paths based on the new `max_version` keyword argument, and also now has `dl_device` and `copy` keyword arguments
- `from_dlpack` method has been updated with `device` and `copy` keyword arguments and now has a path for arrays supporting the new `__dlpack__` interface and `DLManagedTensorVersioned`. `from_dlpack` also honors the read-only bitmask that can now be present on a `DLManagedTensorVersioned` object
- an enumerator class for the DLPack device types has been added to `_usmarray.pyx`

This pull request also slips in a minor change to `usm_ndarray`'s constructor to honor the writable flag when `buffer` is another `usm_ndarray`

- [X] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to an issue with a reproducer?
- [X] Have you tested your changes locally for CPU and GPU devices?
- [X] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [X] If this PR is a work in progress, are you opening the PR as a draft?

PR #1670, opened 2024-05-10
Add more dlpack tests

oleksandr-pavlyk:
This PR adds tests to improve coverage in gh-1667

### DIFF
--- a/dpctl/tests/test_usm_ndarray_dlpack.py
+++ b/dpctl/tests/test_usm_ndarray_dlpack.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import collections
 import ctypes
 
 import pytest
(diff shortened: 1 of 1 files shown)


dpctl/tests/test_usm_ndarray_dlpack.py
```diff
@@ -234,3 +247,53 @@ def test_dlpack_from_subdevice():
...
+    del cap
+
+
+def test_versioned_dlpack_capsule():
+    try:
+        x = dpt.arange(100, dtype="i4")
+    except dpctl.SyclDeviceCreationError:
+        pytest.skip("No default device available")
+
+    cap = x.__dlpack__(max_version=(1, 0))
```
ndgrigorian, line 283: May be a good idea to replace `max_version=(1, 0)` calls with the result of `get_build_dlpack_version`, so they don't have to be changed in the future.

ndgrigorian: Actually, thinking about it more, never mind—it's fine like this. As long as the build version is higher than (1, 0) it will work, and this specifically tests the versioned capsule, which may be a subset of future functionality.

ndgrigorian (approved):
This LGTM! Thank you for the addition of tests to the branch @oleksandr-pavlyk

Issue #1654, opened 2024-04-24
Segmentation fault with dpctl.tensor.argsort on CPU device


antonwolfy:
The below example crashes with Segmentation fault while running on CPU device:
```python
import dpctl.tensor as dpt
from numpy.random import default_rng

N = 4 * 10**6
rng = default_rng(42)
a = rng.random((N,))

ia = dpt.asarray(a, device='cpu')
block_size = 65536

for i in range(0, len(ia), block_size):
    tmp_ia = ia[i : i + block_size]
    sorting_index = dpt.argsort(tmp_ia)
```
<details>
  <summary>The callstack from `gdb` output</summary>

  ### Backtrace
  ```bash
(gdb) bt
#0  0x00007fffac0554fc in typeinfo name for dpctl::tensor::kernels::sort_detail::merge_adjacent_blocks_from_temp_krn<std::pair<double, long>, dpctl::tensor::kernels::TupleComp<std::pair<double, long>, dpctl::tensor::py_internal::(anonymous namespace)::ExtendedRealFPLess<double> > > ()
#1  0x00007fff4c34f000 in ?? ()
#2  0x00007fff4c34f000 in ?? ()
#3  0x00007fff4c34f000 in ?? ()
#4  0x00007fff4c34f000 in ?? ()
#5  0x00007fff4c34f000 in ?? ()
#6  0x00007fff4c34f000 in ?? ()
#7  0x00007fff4c34f000 in ?? ()
#8  0x0000000000000040 in ?? ()
#9  0x0000000000000800 in ?? ()
#10 0x0000000000000800 in ?? ()
#11 0x0000000000000800 in ?? ()
#12 0x0000000000000800 in ?? ()
#13 0x0000000000000800 in ?? ()
#14 0x0000000000000800 in ?? ()
#15 0x0000000000000800 in ?? ()
#16 0x0000000000000800 in ?? ()
#17 0x0000000000000440 in ?? ()
#18 0x0000000000000000 in ?? ()
  ```

  ### Full backtrace
  ```bash
(gdb) thread apply all bt

Thread 34 (Thread 0x7fff4cc4f700 (LWP 19771)):
[...]

PR #1676, opened 2024-05-13
Fix for crash reported in gh-1654

oleksandr-pavlyk:
Closes gh-1654

The reason behind the crash was out of bound access to shared local memory accessor.

This also fixes a crash in `dpt.sort` on CUDA device for sorting of 256 elements of floating point numbers.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/sorting/sort.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/sorting/sort.hpp
@@ -67,7 +67,7 @@ void merge_impl(const std::size_t offset,
     const std::size_t local_size_2 = local_end_2 - local_start_2;
 
     const auto r_item_1 = in_acc[end_1 - 1];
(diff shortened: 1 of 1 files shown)


ndgrigorian (approved):
CUDA tests pass, so LGTM. Nice fix to get in this version.

oleksandr-pavlyk:
The example that was crashing on CUDA device:

```python
import dpctl.tensor as dpt
x = dpt.arange(256, dtype="i4", device="cuda:gpu")
dpt.sort(x, descending=True) # crash
```

antonwolfy (approved):
I checked locally DPNP tests relating to `sort` and `argsort` functions and they passed without any issue.
Also the gh-issue reported as #1676 isn't reproduced anymore locally for me with the proposed dpctl fix.

Thank you @oleksandr-pavlyk for resolving the issue!

PR #1678, opened 2024-05-15
`order` of output array for binary elementwise funcs, `clip`, and `matmul` defaults to C-contiguous

ndgrigorian:
This PR proposes a change to `clip`, `matmul`, and the `BinaryElementwiseFunc` class which makes `C` the default contiguity of outputs when `order="K"` and the input arrays must be cast.

This fixes a minor discrepancy where when multiple input arrays are both C- and F-contiguous and need to be cast, the output defaulted to F-contiguous instead of C-contiguous, creating unusual edge cases where type promotion (and therefore the selected device) became relevant to the output array's order, which should not be the case.

For example
```

In [2]: x1 = dpt.ones((10, 1), dtype="i4", device="cpu")

In [3]: x2 = dpt.ones((1, 5), dtype="f4", device="cpu")

In [4]: r = dpt.multiply(x1, x2)

In [5]: r.flags
Out[5]:
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  WRITABLE : True
```
The selected device's type promotion graph contains `float64`, leading to F-contiguous outputs.

after this change:
```

In [2]: x1 = dpt.ones((10, 1), dtype="i4", device="cpu")
x
In [3]: x2 = dpt.ones((1, 5), dtype="f4", device="cpu")

In [4]: r = dpt.multiply(x1, x2)

In [5]: r.flags
Out[5]:
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  WRITABLE : True
```

### DIFF
--- a/dpctl/tensor/_clip.py
+++ b/dpctl/tensor/_clip.py
@@ -791,17 +791,17 @@ def clip(x, /, min=None, max=None, out=None, order="K"):
 
         if order == "K":
             if (
-                x.flags.f_contiguous
-                and a_min.flags.f_contiguous
-                and a_max.flags.f_contiguous
-            ):
-                order = "F"
-            elif (
                 x.flags.c_contiguous
                 and a_min.flags.c_contiguous
                 and a_max.flags.c_contiguous
             ):
                 order = "C"
+            elif (
+                x.flags.f_contiguous
+                and a_min.flags.f_contiguous
+                and a_max.flags.f_contiguous
+            ):
+                order = "F"
         if order == "K":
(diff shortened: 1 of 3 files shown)


oleksandr-pavlyk:
This looks uncontroversial enough for me.

oleksandr-pavlyk (approved):
LGMT!

ndgrigorian:
@vtavana confirmed that dpctl test suite was unaffected by this change, I will go ahead and merge it

Internal CI failures also unrelated

PR #1680, opened 2024-05-16
Fix bug in basic slicing of empty arrays

oleksandr-pavlyk:
Reproducer provided by @ndgrigorian:

```python
import dpctl.tensor as dpt
x = dpt.ones((0, 10))
y = dpt.moveaxis(x, 1, 0)
y[1] # raises ValueError
```

Now this returns empty 1d array like `y[0]` does.

A test is added for both integral basic indexing,
and for simple slice.

---

### DIFF
--- a/dpctl/tensor/_slicing.pxi
+++ b/dpctl/tensor/_slicing.pxi
@@ -120,9 +120,11 @@ def _basic_slice_meta(ind, shape : tuple, strides : tuple, offset : int):
         sh0 = _slice_len(sl_start, sl_stop, sl_step)
         str0 = sl_step * strides[0]
         new_strides = strides if (sl_step == 1 or sh0 == 0) else (str0,) + strides[1:]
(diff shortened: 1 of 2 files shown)


ndgrigorian (approved):
This resolves the array API test failure. I've also tested it out in some edge cases and it seems to work as intended.

```

In [2]: x = dpt.arange(100, dtype="f4")

In [3]: x1 = dpt.reshape(x[2:], (7, 14))

In [4]: x1[-1:0, :].mT[2]
Out[4]: usm_ndarray([], dtype=float32)

In [5]: _.shape
Out[5]: (0,)

In [6]: test = x1[-1:0, :]

In [7]: test.shape
Out[7]: (0, 14)

In [8]: test.shape = (0, 100)

In [9]: test.mT[2]
Out[9]: usm_ndarray([], dtype=float32)

In [10]: test[:, 2]
Out[10]: usm_ndarray([], dtype=float32)
```

Windows CI break is unrelated...

PR #1681, opened 2024-05-16
Disable vectorized `bitwise_invert` for boolean inputs

ndgrigorian:
This PR proposes removing the `sycl::vec` overload of `bitwise_invert` for boolean inputs to fix unexpected behaviors.

Vectorized `bitwise_invert` on boolean inputs would corrupt the data in a subtle way. When casting this data to an integer data type, 255 would always be the result of casting `true`.
This is likely related to `operator!` on `sycl::vec` returning -1 for `true` and expecting the output type to be a signed integer type.


oleksandr-pavlyk (approved):
This looks good to me. Thank you @ndgrigorian ! I have added @npolina4 as a reviewer as original reporter of this issue.

npolina4 (approved):
Looks good to me. Thank you @ndgrigorian!

PR #1682, opened 2024-05-16
Removes malformed fast path for zero-size arrays in `repeat`

ndgrigorian:
This PR fixes a small bug in `dpctl.tensor.repeat` caught by the array API tests where calling `repeat` with `axis=None` on a zero-size array would result in an array with the shape of the input, rather than `x.shape == (0,)` as expected.

To fix this, the malformed fast path for zero-size arrays has been removed, as it was unnecessary: all calls in repeat are already special-cased in the C++ backend to return empty events for zero-size inputs, so any savings for this path were minimal.

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -838,9 +838,6 @@ def repeat(x, repeats, /, *, axis=None):
             f"got {type(repeats)}"
         )
 
-    if axis_size == 0:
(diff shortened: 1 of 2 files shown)


ndgrigorian:
@oleksandr-pavlyk 
I've added a test for the correct behavior.

oleksandr-pavlyk (approved):
LGTM!  Having tests has been helpful!

Issue #292, opened 2021-02-26
Support for aspects


1e-to:
Add aspects as a new enum class (DPCTLSyclAspect) to dpctl_sycl_enum_types.h
Maybe implemented as bool DPCTLDevice_Has(DPCTLSyclAspect A) in C API and cpdef has(aspect=*) in Cython?

```c
enum DPCTLSyclAspect
{
    cpu,
    gpu,
    accelerator,
    custom,
    emulated,
    host_debuggable,
    fp16,
    fp64,
    atomic64,
    image,
    online_compiler,
    online_linker,
    queue_profiling,
    usm_device_allocations,
    usm_host_allocations,
    usm_atomic_host_allocations,
    usm_shared_allocations,
    usm_atomic_shared_allocations,
    usm_system_allocations
};

```

PR #1691, opened 2024-05-22
Add support for aspect::emulated

oleksandr-pavlyk:
Add `dpctl.SyclDevice.has_aspect_emulated` now that DPC++ supports it, see gh-292

### DIFF
--- a/dpctl/tests/test_sycl_device.py
+++ b/dpctl/tests/test_sycl_device.py
@@ -163,13 +163,12 @@ list_of_supported_aspects = [
     "atomic64",
     "usm_atomic_host_allocations",
     "usm_atomic_shared_allocations",
+    "emulated",
 ]
 
 # SYCL 2020 spec aspects not presently
 # supported in DPC++, and dpctl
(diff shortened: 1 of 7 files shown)


dpctl/tests/test_sycl_device.py
```diff
@@ -163,13 +163,12 @@ def test_equal():
...
     "usm_atomic_shared_allocations",
+    "emulated",
 ]
 
 # SYCL 2020 spec aspects not presently
 # supported in DPC++, and dpctl
-list_of_unsupported_aspects = [
-    "emulated",
-]
+list_of_unsupported_aspects = []
```
ndgrigorian, line 171: @oleksandr-pavlyk 
Should this list be removed at this point?

oleksandr-pavlyk: This list feeds into a fixture that parametrizes test for unsupported aspects. I'd keep it in case spec adds an aspect which compiler does not yet support.

ndgrigorian (approved):
LGTM!

PR #1706, opened 2024-06-07
Implement manipulation of flattened SO files

oleksandr-pavlyk:
This PR modifies `build.sh` to post-process the wheel and remove duplicate shared objects produced by Python packaging that follows symbolic links.

Symbolic links are restored for conda after wheel is installed.

The check ``test_service.py::test_syclinterface`` was modified to ignore presence of duplicate copies of the library shared objects if `--no-sycl-interface-test` option was specified to pytest.

The option is used in 
- coverage collection script `scripts/get_coverage.py`
- in workflow which builds `dpctl` with nighly sycl bundle of intel/llvm.

-------

### DIFF
--- a/dpctl/tests/test_service.py
+++ b/dpctl/tests/test_service.py
@@ -151,13 +151,21 @@ def test_dev_utils():
             device.parent_device
 
 
-def test_syclinterface():
+@pytest.fixture
+def should_skip_syclinterface(request):
+    return request.config.getoption("--no-sycl-interface-test")
+
+
+def test_syclinterface(should_skip_syclinterface):
     install_dir = os.path.dirname(os.path.abspath(dpctl.__file__))
     paths = glob.glob(os.path.join(install_dir, "*DPCTLSyclInterface*"))
     if "linux" in sys.platform:
-        assert len(paths) > 1 and any(
-            [os.path.islink(fn) for fn in paths]
-        ), "All library instances are hard links"
+        if should_skip_syclinterface:
+            pass
+        else:
+            assert len(paths) > 1 and any(
+                [os.path.islink(fn) for fn in paths]
+            ), "All library instances are hard links"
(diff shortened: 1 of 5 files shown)


ZzEeKkAa:
@oleksandr-pavlyk should we just keep `libdpctlsyclinterface.so.0.18` And teach dependent projects to import just it?

dpctl/tests/test_service.py
```diff
@@ -151,13 +151,21 @@ def test_dev_utils():
             device.parent_device
 
 
-def test_syclinterface():
+@pytest.fixture
+def should_skip_syclinterface(request):
+    return request.config.getoption("--no-sycl-interface-test")
+
+
+def test_syclinterface(should_skip_syclinterface):
```
ZzEeKkAa, line 159: May be something like `@pytest.mark.skipif` would be cleaner?
https://docs.pytest.org/en/stable/how-to/skipping.html#id1

oleksandr-pavlyk: I think this change should be deferred to a later PR.

Essentially, it would replace `pass` with a call to `pytest.skip`.

oleksandr-pavlyk:
No, we must keep `so.0`.

In case of incompatibility, we must increment major version.

oleksandr-pavlyk:
We need both .DLL and .LIB. The lib are used for linking (they are essentially loaders of DLL).

oleksandr-pavlyk:
> No, we must keep `so.0`.
> 
> In case of incompatibility, we must increment major version.

For soversion-ed libraries, linker searches for `so.0` as evidenced either `ldd` or `readelf`. If `.so.0` is deleted in favor of keeping `so.0.18`, importing of dpctl would stop working

ZzEeKkAa (approved):
LGMT!

PR #1707, opened 2024-06-08
Remove USE_SYCL_FOR_COMPLEX_TYPES

oleksandr-pavlyk:
Always use SYCL namespace functions. Also replaced some std namespace functions with corresponding sycl namespace functions.


ndgrigorian:
Tested build with CUDA and it built successfully, so this LGTM.

Since I had to add pre-commit changes, can @antonwolfy or @vtavana please take a look at the changes and review this PR?

I still see some uses of `std::sin`, `std::cos`, etc. but I think these can be replaced in a follow-up PR.

oleksandr-pavlyk:
The last commit, although green in CI, no longer build for NVPTX64 target, crashing during build of `cumulative_logsumexp.cpp` file.

It is best to report this to the attention of compiler engineers as a standalone reproducer and revert the culprit.

ndgrigorian:
> The last commit, although green in CI, no longer build for NVPTX64 target, crashing during build of `cumulative_logsumexp.cpp` file.
> 
> It is best to report this to the attention of compiler engineers as a standalone reproducer and revert the culprit.
> 
> Crash log

Culprit appears to be `sycl::log1p`. I will revert the change to use `sycl::log1p` before this is merged, among other `std` function replacements.

ndgrigorian (approved):
With `sycl::log1p` to `std::log1p` change for `logaddexp` and remaining `std` functions changed, CUDA builds and tests seem to pass. I think the PR is ready.

antonwolfy (approved):
@ndgrigorian, `dpnp` tests passed with these changes without any new issue.

PR #1719, opened 2024-06-26
Add Python scalar support to `dpt.where`

ndgrigorian:
This PR proposes a change to `dpt.where` to add support for Python scalars, aligning it with `clip` and elementwise functions.

As has been pointed out by the array API community, an common use-case for `dpt.where` is supporting Python scalars. If the decision to permit them in other elementwise functions is made, `where` will likely follow suit.

Regardless of the outcome of this decision, it's sensible to support them here.

As the first argument (`condition`) does not participate in the type promotion in `where`, if both arguments are scalars, the scalars are treated as the default types for the respective kind.

### DIFF
--- a/dpctl/tensor/_search_functions.py
+++ b/dpctl/tensor/_search_functions.py
@@ -121,16 +121,17 @@ def where(condition, x1, x2, /, *, order="K", out=None):
             and otherwise yields from ``x2``.
             Must be compatible with ``x1`` and ``x2`` according
             to broadcasting rules.
-        x1 (usm_ndarray): Array from which values are chosen when
-            ``condition`` is ``True``.
+        x1 (Union[usm_ndarray, bool, int, float, complex]):
+            Array from which values are chosen when ``condition`` is ``True``.
             Must be compatible with ``condition`` and ``x2`` according
             to broadcasting rules.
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_search_functions.py
```diff
@@ -81,36 +151,90 @@ def where(condition, x1, x2, /, *, order="K", out=None):
         raise TypeError(
             "Expecting dpctl.tensor.usm_ndarray type, " f"got {type(condition)}"
         )
-    if not isinstance(x1, dpt.usm_ndarray):
-        raise TypeError(
-            "Expecting dpctl.tensor.usm_ndarray type, " f"got {type(x1)}"
+    if order not in ["K", "C", "F", "A"]:
```
antonwolfy, line 154: Do we need to update the docstring and to state scalars as possible type of input?

ndgrigorian: Yes, good point

antonwolfy (approved):
I checked the PR in dpnp tests and all tests are passed on GPU and CPU devices.

ndgrigorian:
CI failure is unrelated to this PR, I will go ahead and merge it

PR #1726, opened 2024-07-08
Remove 'template' from call to `get_kernel` method of kernel bundle in sorting

ndgrigorian:
This PR removes an unnecessary templating of the `get_kernel` method of the kernel bundle in `sort.hpp`.

As no template parameter was given, this not only did not do anything, but was causing nightly builds to fail.


antonwolfy (approved):
LGTM!

PR #1727, opened 2024-07-08
`extract` and `_extract_impl` now coerce USM type

### LINKED ISSUE #1723: Question about USM type of result array for `dpt.extract`
Due to some reason USM type of result array isn't coerced from USM types of inputs:
```python

dpctl.__version__
# Out: '0.18.0dev0+77.g26d34f565d'

 a = dpt.ones(2, usm_type='shared')
cond = dpt.asarray([True, False], usm_type='device')

dpt.extract(cond, a).usm_type
# Out: 'shared'

dpctl.utils.get_coerced_usm_type([a.usm_type, cond.usm_type])
# Out: 'device'
```
Normally, for other dpctl tensor functions, the USM type of result would be `'device'` for the same inputs as in above example.

ndgrigorian:
This PR changes `_extract_impl` in `_copy_utils` to coerce USM types to bring `extract` and boolean indexing in line with other operations in dpctl which take multiple arrays as input

Closes #1723 

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -707,6 +707,9 @@ def _extract_impl(ary, ary_mask, axis=0):
         raise TypeError(
             f"Expecting type dpctl.tensor.usm_ndarray, got {type(ary_mask)}"
         )
+    dst_usm_type = dpctl.utils.get_coerced_usm_type(
+        (ary.usm_type, ary_mask.usm_type)
+    )
     exec_q = dpctl.utils.get_execution_queue(
         (ary.sycl_queue, ary_mask.sycl_queue)
     )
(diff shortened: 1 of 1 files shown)


antonwolfy (approved):
Thank you @ndgrigorian 
All dpnp tests are passed with the PR.

oleksandr-pavlyk:
@ndgrigorian @antonwolfy 

I think that behavior was intentional. One can not expect the symmetry type promotion does not apply (data type of index array does not influence the type of the output).

antonwolfy:
@oleksandr-pavlyk
Then it looks a bit confusing and unclear why the behavior of `dpt.extract` differs from `dpt.where` for instance, where USM type of `condition` array influences on USM type of result array.
And also it's true for other dpctl functions from indexing routing, where the USM type of result array is building based on USM type of all input arrays.

Will be the change roll backed? And should the behavior be clarified in description of `dpt.extract` function then?

oleksandr-pavlyk:
I see. Internal consistency is important. Let's discuss this, decide one way or another and document the decision.

PR #1731, opened 2024-07-13
Enable building to target AMD GPUs 

ndgrigorian:
This PR modifies cmake scripts throughout dpctl to enable building for AMD. This is done by either setting the `DPCTL_TARGET_HIP` environment variable to the intended build architecture, or using `-DDPCTL_TARGET_HIP`.

`_dpctl_sycl_target_compile_options` and `_dpctl_sycl_target_link_options` cmake lists are used to prevent branching logic in later scripts.


oleksandr-pavlyk:
@ndgrigorian Regrettably use of `sycl::log1p` change necessary to enable compiling for AMD breaks compiling for CUDA.

Perhaps a preprocessor variable can be used to enable building for SPV/NVPTX or SPV/AMDGCN targets, but not for all three except after the bug gets fixed. It may be possible to write implementation of log1p to enable building for all three too.

ndgrigorian:
> @ndgrigorian Regrettably use of `sycl::log1p` change necessary to enable compiling for AMD breaks compiling for CUDA.
> 
> Perhaps a preprocessor variable can be used to enable building for SPV/NVPTX or SPV/AMDGCN targets, but not for all three except after the bug gets fixed. It may be possible to write implementation of log1p to enable building for all three too.

Yes, I only added the commit to make it convenient for the build failure to be reproduced.

Writing our own implementation is possible, too. I think that would be preferable, but on the other hand, it's a corner case to build for both CUDA and AMD.

oleksandr-pavlyk:
I think the issue is not with `sycl::log1p` vs `std::log1p`. Using one over another masks the underlying issue. With `sycl::log1p` used, it is possible to make compilation command succeed by manually removing `-fno-delete-null-pointer-checks` too. 

That said, I too was not able to produce a stand-alone reproducer.

ndgrigorian:
I have successfully built dpctl for both CUDA and HIP simultaneously using this branch on a machine with CUDA and ROCm installed (no AMD devices, however—only Intel and NVidia).

Errors came up while building, but I believe that these were caused by going OOM—errors showed up twice in `prod.cpp`, once in `sum.cpp`, and once in `copy_and_cast_usm_to_usm.cpp`. `prod`, `sum`, etc. seemed to work on both level-zero and CUDA without a problem, and tests passed. Even in verbose mode, the message wasn't especially helpful.

I have marked this PR as ready for review, the CUDA segfault is resolved.

ndgrigorian:
> @oleksandr-pavlyk @antonwolfy
> 
> I have successfully built dpctl for both CUDA and HIP simultaneously using this branch on a machine with CUDA and ROCm installed (no AMD devices, however—only Intel and NVidia).
> 
> Errors came up while building, but I believe that these were caused by going OOM—errors showed up twice in `prod.cpp`, once in `sum.cpp`, and once in `copy_and_cast_usm_to_usm.cpp`. `prod`, `sum`, etc. seemed to work on both level-zero and CUDA without a problem, and tests passed. Even in verbose mode, the message wasn't especially helpful.
> 
> I have marked this PR as ready for review, the CUDA segfault is resolved.

Worth noting that I tried building with both `DPCTL_TARGET_HIP=gfx1100` and `DPCTL_TARGET_HIP=gfx1030`.

oleksandr-pavlyk:
Is this the command to use ?  

```
python scripts/build_locally.py --verbose --cmake-opts="-DDPCTL_TARGET_HIP=gfx1030"
```

oleksandr-pavlyk (approved):
Looks good to me! Thank you @ndgrigorian . 

It makes sense to similarly allow `DPCTL_TARGET_CUDA` to take non-boolean values to specify architecture in a future.

PR #1732, opened 2024-07-13
Implements `dpctl.tensor.count_nonzero` and `dpctl.tensor.diff`

ndgrigorian:
This PR proposes introducing two new functions to dpctl which will be coming to the array API specification's next version: `count_nonzero` and `diff`.

* `count_nonzero` casts an array to `bool` before calling `sum`. As dedicated kernels for `sum` of `bool` into any integer type, the summation won't require any additional copies.

* `diff` recursively takes the `n`-th forward difference along a given axis of an array. `prepend` and `append` arguments (which can be arrays or scalars) are concatenated to the input array. We diverge from the array API by permitting scalars for `prepend` and `append` as well as not constraining them to the same type as the input array

### DIFF
--- a/dpctl/tensor/_reduction.py
+++ b/dpctl/tensor/_reduction.py
@@ -773,3 +773,15 @@ def argmin(x, /, *, axis=None, keepdims=False, out=None):
             default array index data type for the device of ``x``.
     """
     return _search_over_axis(x, axis, keepdims, out, tri._argmin_over_axis)
+
+
+def count_nonzero(x, /, *, axis=None, keepdims=False, out=None):
+    if x.dtype != dpt.bool:
+        x = dpt.astype(x, dpt.bool, copy=False)
+    return sum(
+        x,
+        axis=axis,
+        dtype=ti.default_device_index_type(x.sycl_device),
+        keepdims=keepdims,
+        out=None,
+    )
--- a/dpctl/tensor/_reduction.py
+++ b/dpctl/tensor/_reduction.py
@@ -776,6 +776,37 @@ def argmin(x, /, *, axis=None, keepdims=False, out=None):
 
 
 def count_nonzero(x, /, *, axis=None, keepdims=False, out=None):
+    """
+    Counts the number of elements in the input array ``x`` which are non-zero.
+
+    Args:
+        x (usm_ndarray):
+            input array.
+        axis (Optional[int, Tuple[int, ...]]):
+            axis or axes along which to count. If a tuple of unique integers,
+            the number of non-zero values are computed over multiple axes.
+            If ``None``, the number of non-zero values is computed over the
+            entire array.
+            Default: ``None``.
+        keepdims (Optional[bool]):
+            if ``True``, the reduced axes (dimensions) are included in the
+            result as singleton dimensions, so that the returned array remains
+            compatible with the input arrays according to Array Broadcasting
+            rules. Otherwise, if ``False``, the reduced axes are not included
+            in the returned array. Default: ``False``.
+        out (Optional[usm_ndarray]):
+            the array into which the result is written.
+            The data type of ``out`` must match the expected shape and data
+            type.
+            If ``None`` then a new array is returned. Default: ``None``.
+
+    Returns:
+        usm_ndarray:
+            an array containing the count of non-zero values. If the sum was
+            computed over the entire array, a zero-dimensional array is
+            returned. The returned array will have the default array index data
+            type.
+    """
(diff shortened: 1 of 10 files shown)


dpctl/tensor/_reduction.py
```diff
@@ -773,3 +773,46 @@ def argmin(x, /, *, axis=None, keepdims=False, out=None):
...
+            type.
+    """
+    if x.dtype != dpt.bool:
+        x = dpt.astype(x, dpt.bool, copy=False)
+    return sum(
+        x,
+        axis=axis,
+        dtype=ti.default_device_index_type(x.sycl_device),
+        keepdims=keepdims,
+        out=None,
```
oleksandr-pavlyk, line 817: Should this be `out=out`?

ndgrigorian: Definitely was supposed to be, good catch

dpctl/tensor/_utility_functions.py
```diff
@@ -144,3 +156,315 @@ def any(x, /, *, axis=None, keepdims=False):
...
+    # form slices and recurse
+    sl0 = tuple(
+        slice(None) if i != axis else slice(1, None) for i in range(x_nd)
+    )
+    sl1 = tuple(
+        slice(None) if i != axis else slice(None, -1) for i in range(x_nd)
+    )
+
+    for _ in range(n):
+        arr = arr[sl0] - arr[sl1]
```
oleksandr-pavlyk, line 468: For `n > 1` we can be more economical about allocating new arrays. We could allocate two temporary arrays to write into and recycle them in the loop.

ndgrigorian: The most recent commit accomplishes this by unrolling the first two steps of the loop when `n > 1` and storing those outputs in temporaries, reusing them.

dpctl/tests/test_tensor_diff.py
```diff
@@ -0,0 +1,352 @@
...
+def test_diff_axis():
+    get_queue_or_skip()
+
+    x = dpt.tile(
+        dpt.asarray([9, 12, 7, 17, 10, 18, 15, 9, 8, 8], dtype="i4"), (3, 4, 1)
+    )
+    x[:, ::2, :] = 0
+
+    for n in [1, 2, 3]:
+        res = dpt.diff(x, n=3, axis=1)
```
oleksandr-pavlyk, line 74: Did you mean `res = dpt.diff(x, n=n, axis=1)` ?

ndgrigorian: Yes, that's correct, I'll push a fix

dpctl/tests/test_tensor_diff.py
```diff
@@ -0,0 +1,352 @@
...
+    get_queue_or_skip()
+
+    dts = [
+        ("i1", "u1", "i8"),
+        ("i1", "u8", "u1"),
+        ("u4", "i4", "f4"),
+        ("i8", "c8", "u8"),
+    ]
+
+    for _dts in dts:
```
oleksandr-pavlyk, line 93: Maybe better to do `for dt0, dt1, dt2 in dts:` and used these named variables instead of `_dts[0]`, `_dts[1]` and `_dts[2]`.

dpctl/tests/test_tensor_diff.py
```diff
@@ -0,0 +1,352 @@
...
+    q3 = get_queue_or_skip()
+
+    ar1 = dpt.ones(1, dtype="i4", sycl_queue=q1)
+    ar2 = dpt.ones(1, dtype="i4", sycl_queue=q2)
+    ar3 = dpt.ones(1, dtype="i4", sycl_queue=q3)
+
+    assert_raises_regex(
+        ExecutionPlacementError,
+        "Execution placement can not be unambiguously inferred from input "
+        "arguments",
```
oleksandr-pavlyk, line 299: Same thing here. Checking exception type to be `ExecutionPlacementError` should be sufficient

oleksandr-pavlyk:
Please add `tensor.diff` and `tensor.count_nonzero` to docs.

ndgrigorian:
> Please add `tensor.diff` and `tensor.count_nonzero` to docs.

I will add `nextafter` too, since it was forgotten

oleksandr-pavlyk:
On the diff page, can the bullet be changed into an indent?

image

oleksandr-pavlyk (approved):
Thank you @ndgrigorian ! Ready to be merged

PR #1619, opened 2024-03-28
Overhaul of dpctl documentation

oleksandr-pavlyk:
This PR is a rewrite of `dpctl`'s documentation. Here is the link to [rendered docs](https://intelpython.github.io/dpctl/pulls/1619/index.html) based on sources in this PR.

1. It introduces Beginner's guide, User Guide, Reference Guide and contributor's guide
2. It does away with `generate_rst.py` and AutoAutosummary class, but uses Sphinx's autosummary extension to generate individual pages for classes, and array API functions.
3. It changes theme from read-the-docs to Furo
4. It modifies docstrings to fix sphinx issues
5. It modifies array API function signatures by using positional-only, keyword-only delimiters as done in the standard. 
    This required test changes and code changes. 

- [x] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [x] If this PR is a work in progress, are you opening the PR as a draft?

PR #1740, opened 2024-07-18
Removed use of deprecated SYCL_DEVICE_FILTER

oleksandr-pavlyk:
Replaced it with `ONEAPI_DEVICE_SELECTOR` environment variable.

### DIFF
--- a/examples/python/device_selection.py
+++ b/examples/python/device_selection.py
@@ -25,8 +25,8 @@ def create_default_device():
     Create default SyclDevice using `sycl::default_selector`.
 
     Device created can be influenced by environment variable
-    SYCL_DEVICE_FILTER, which determines SYCL devices seen by the
-    SYCL runtime.
+    ONEAPI_DEVICE_SELECTOR, which determines SYCL devices seen by the
+    oneAPI DPC++ runtime.
     """
     d1 = dpctl.SyclDevice()
     d2 = dpctl.select_default_device()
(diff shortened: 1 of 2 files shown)


examples/python/device_selection.py
```diff
@@ -25,8 +25,8 @@ def create_default_device():
     Create default SyclDevice using `sycl::default_selector`.
 
     Device created can be influenced by environment variable
-    SYCL_DEVICE_FILTER, which determines SYCL devices seen by the
-    SYCL runtime.
+    ONEAPI_DEVICE_SELECTOR, which determines SYCL devices seen by the
```
antonwolfy, line 28: `SYCL_DEVICE_FILTER` still mentioned in `docs/_legacy/docfiles/user_guides/manual/dpctl/device_selection.rst` and `docs/_legacy/docfiles/user_guides/manual/dpctl/devices.rst` files.
Is it expected?

examples/python/device_selection.py
```diff
@@ -25,8 +25,8 @@ def create_default_device():
...
+    oneAPI DPC++ runtime.
```
antonwolfy, line 29: Should be `SYCL runtime` replaced with `oneAPI DPC++ runtime` everywhere? Still present in many files.

examples/python/device_selection.py
oleksandr-pavlyk: `docs/_legacy` contains sources of `dpctl` from before gh-1619. This folder will be removed in near future.

examples/python/device_selection.py
oleksandr-pavlyk: Probably yes, it is more precise, as we do not presently support other SYCL implementations, but I would defer that change to a different PR.

antonwolfy (approved):
No more comments from me. Thank you @oleksandr-pavlyk

PR #1753, opened 2024-07-24
Release GIL around blocking operations in libtensor

oleksandr-pavlyk:
Copy from NumPy ndarray to usm_ndarray is blocking, so release GIL.

mask-positions (cumulative value) which returns a total is blocking, so release the GIL.

### DIFF
--- a/dpctl/tensor/libtensor/source/accumulators.cpp
+++ b/dpctl/tensor/libtensor/source/accumulators.cpp
@@ -160,10 +160,14 @@ size_t py_mask_positions(const dpctl::tensor::usm_ndarray &mask,
                       ? mask_positions_contig_i32_dispatch_vector[mask_typeid]
                       : mask_positions_contig_i64_dispatch_vector[mask_typeid];
 
-        size_t total_set = fn(exec_q, mask_size, mask_data, cumsum_data,
-                              host_task_events, depends);
+        size_t total_set{};
+
         {
             py::gil_scoped_release release;
+
+            total_set = fn(exec_q, mask_size, mask_data, cumsum_data,
+                           host_task_events, depends);
+
             sycl::event::wait(host_task_events);
         }
         return total_set;
@@ -198,12 +202,13 @@ size_t py_mask_positions(const dpctl::tensor::usm_ndarray &mask,
     sycl::event copy_shape_ev = std::get<2>(ptr_size_event_tuple);
 
     if (2 * static_cast<size_t>(nd) != std::get<1>(ptr_size_event_tuple)) {
-        copy_shape_ev.wait();
         {
             py::gil_scoped_release release;
+
+            copy_shape_ev.wait();
             sycl::event::wait(host_task_events);
+            sycl::free(shape_strides, exec_q);
         }
-        sycl::free(shape_strides, exec_q);
         throw std::runtime_error("Unexpected error");
     }
 
@@ -213,15 +218,17 @@ size_t py_mask_positions(const dpctl::tensor::usm_ndarray &mask,
     dependent_events.insert(dependent_events.end(), depends.begin(),
                             depends.end());
 
-    size_t total_set =
-        strided_fn(exec_q, mask_size, mask_data, nd, shape_strides, cumsum_data,
-                   host_task_events, dependent_events);
+    size_t total_set;
 
     {
         py::gil_scoped_release release;
+
+        total_set = strided_fn(exec_q, mask_size, mask_data, nd, shape_strides,
+                               cumsum_data, host_task_events, dependent_events);
+
         sycl::event::wait(host_task_events);
+        sycl::free(shape_strides, exec_q);
     }
-    sycl::free(shape_strides, exec_q);
 
     return total_set;
 }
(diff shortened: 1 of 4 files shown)


dpctl/tensor/libtensor/source/accumulators.cpp
```diff
@@ -160,10 +160,14 @@ size_t py_mask_positions(const dpctl::tensor::usm_ndarray &mask,
                       ? mask_positions_contig_i32_dispatch_vector[mask_typeid]
                       : mask_positions_contig_i64_dispatch_vector[mask_typeid];
 
-        size_t total_set = fn(exec_q, mask_size, mask_data, cumsum_data,
-                              host_task_events, depends);
+        size_t total_set{};
```
ndgrigorian, line 163:
```suggestion
        size_t total_set;
```
nit

dpctl/tensor/libtensor/source/copy_numpy_ndarray_into_usm_ndarray.cpp
```diff
@@ -116,21 +116,29 @@ void copy_numpy_ndarray_into_usm_ndarray(
...
 
-            // wait for copy_ev to complete
-            copy_ev.wait_and_throw();
+            {
+                // wait for copy_ev to complete
+                // release GIL to allow other threads (host_tasks)
+                // a chance to acquire GIL
+                py::gil_scoped_release lock{};
+                copy_ev.wait_and_throw();
+            }
```
ndgrigorian, line 141: Do we have tests for these cases? If not, would be nice to add a few.

oleksandr-pavlyk: An example that hung due to unreleased GIL, discovered by @vtavana, was (with latest `dpnp`)

```python
import dpctl.tensor as dpt
import dpnp

a_py = [-1,2,1,3,-4]
a = dpnp.asarray(a_py, dtype=dpnp.complex64)
res = dpnp.fft.fft(a)
print("1")

a_usm = dpt.asarray(a, dtype="c8")
rest = dpnp.fft.fft(a_usm)
print("2")
``` 

It hung during the last call to `fft`, during call to `dpctl` to copy from `numpy.ndarray` to `usm_ndarray`.

I am yet unable to build a stand-alone `dpctl`-pure reproducer.

antonwolfy (dismissed):
The dpnp tests are passed with that changes. The example also work now without any hang.
Thank you @oleksandr-pavlyk

oleksandr-pavlyk:
The last commit fixed a crash in the following snippet:

```python
import dpctl
import dpctl.utils as du
import dpctl.tensor as dpt

class A:
    def __init__(self, arr):
        self._array_obj = arr

    @property
    def __sycl_usm_array_interface__(self):
        return self._array_obj.__sycl_usm_array_interface__

a = dpt.linspace(0, dpt.pi, num=2**27, dtype="f8")

a_dummy = A(a)

a_sin = dpt.sin(a)
print(1)
b = dpt.asarray(a_dummy, dtype="f8")
b /= 1
print(2)
```

Technically such a code is a usage error, since script ends with pending asynchronous tasks. It used to crash due to such tasks requiring GIL which can not be acquired during the interpreter shutdown and caused seg-fault in the thread executing the cleanup  `host_task` which attempted to acquire GIL. 

The fix was to used `weakref.finalize` to perform synchronization of the `dpctl.utils.SequentialOrderManager` at shutdown.

The test is not being added here, but we must document that user is advised to perform the synchronization explicitly before the script ends, and the recommended usage would be

```python
import dpctl
import dpctl.utils as du
import dpctl.tensor as dpt

class A:
    def __init__(self, arr):
        self._array_obj = arr

    @property
    def __sycl_usm_array_interface__(self):
        return self._array_obj.__sycl_usm_array_interface__

a = dpt.linspace(0, dpt.pi, num=2**27, dtype="f8")

a_dummy = A(a)

a_sin = dpt.sin(a)
print(1)
b = dpt.asarray(a_dummy, dtype="f8")
b /= 1
print(2)

# ensure all submitted tasks complete
b.sycl_queue.wait()
```

dpctl/tensor/libtensor/source/copy_numpy_ndarray_into_usm_ndarray.cpp
```diff
@@ -202,6 +210,30 @@ void copy_numpy_ndarray_into_usm_ndarray(
...
+            // wait for copy_ev to complete
+            // release GIL to allow other threads (host_tasks)
+            // a chance to acquire GIL
+            py::gil_scoped_release lock{};
+
+            copy_ev.wait();
+        }
+
+        return;
+    }
```
ndgrigorian, line 235: Should we add some new tests for these specific cases? A couple of sanity checks in case some new memcpy issue occurs.

oleksandr-pavlyk: Yes, good idea. Let me add such a test.

ndgrigorian (approved):
With tests added and CI passing this LGTM!

PR #1756, opened 2024-07-24
Propagate read-only flag from sycl_usm_array_interface in asarray

oleksandr-pavlyk:
Change to `dpt.asarray` to take read-only flag value from ``__sycl_usm_array_interface__`` into account.

```IPython
In [1]: import dpctl, dpctl.utils as du, dpctl.tensor as dpt, dpnp

In [2]: x = dpt.linspace(100, 0, num=101, dtype="f8")

In [3]: x.flags["W"] = False

In [4]: y = dpnp.asarray(x)

In [5]: z = dpt.asarray(y)

In [6]: z.flags
Out[6]:
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  WRITABLE : False
```

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -203,7 +203,7 @@ def _usm_ndarray_from_suai(obj):
         strides=sua_iface.get("strides", None),
     )
     _data_field = sua_iface["data"]
-    if isinstance(_data_field, tuple) and len(_data_field) == 2:
+    if isinstance(_data_field, tuple) and len(_data_field) > 1:
         ro_field = _data_field[1]
     else:
         ro_field = False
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_ctors.py
```diff
@@ -202,6 +202,13 @@ def _usm_ndarray_from_suai(obj):
         buffer=membuf,
         strides=sua_iface.get("strides", None),
     )
+    _data_field = sua_iface["data"]
+    if isinstance(_data_field, tuple) and len(_data_field) == 2:
```
antonwolfy, line 206: Should be the comparison relaxed to be prepared for the future possible changes where the length of data field might be increased and some additional information added to the end?
```suggestion
    if isinstance(_data_field, tuple) and len(_data_field) > 1:
```
Since it is only important to protect against out-of-boundary access, it is sufficient that `len(_data_field) > 1`.

antonwolfy (approved):
All dpnp tests pass with that change. And no more review comments from me.
Thank you @oleksandr-pavlyk

antonwolfy (approved):

PR #1760, opened 2024-07-25
Build with numpy 2.0

oleksandr-pavlyk:
Build conda packages with NumPy 2.0 in the build environment. The `meta.yaml` file for `dpctl` was changed to not use `pin_compatible('numpy', min='x.x', max='x')` per migration guidelines from conda-forge.

Testing should still pick NumPy 1.26.4 from Intel channel.

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Build conda package
         run: |
           # use bootstrap channel to pull NumPy linked with OpenBLAS
-          CHANNELS="-c dppy/label/bootstrap -c ${{ env.INTEL_CHANNEL }} -c conda-forge --override-channels"
+          CHANNELS="-c conda-forge --override-channels"
           VERSIONS="--python ${{ matrix.python }} --numpy 2.0"
           TEST="--no-test"
           conda build \
@@ -109,7 +109,7 @@ jobs:
       - name: Build conda package
         env:
           OVERRIDE_INTEL_IPO: 1   # IPO requires more resources that GH actions VM provides
-        run: conda build --no-test --python ${{ matrix.python }} --numpy 2.0 -c dppy/label/bootstrap -c ${{ env.INTEL_CHANNEL }} -c conda-forge --override-channels conda-recipe
+        run: conda build --no-test --python ${{ matrix.python }} --numpy 2.0 -c conda-forge --override-channels conda-recipe
       - name: Upload artifact
         uses: actions/upload-artifact@v4.3.6
         with:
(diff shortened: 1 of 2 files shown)


ekomarova:
In order for this PR to  be built in the internal CI with 2.0, I need to change the mamba build flag to `--numpy 2.0`. I can try to do this and restart the checks

ekomarova:
For some reason, the environment still resolved in favor of 1.26.4, even despite `--numpy 2.0.0` flag and the absence of any numpy pinning in recipe..

oleksandr-pavlyk:
@ekomarova This PR modified `conda-packages.yml` workflow file to remove use of Intel channel from the `conda build` command and only use `conda-forge`. I just recheck that for both platforms and all version of Python (3.9, 3.10, 3.11, and 3.12) installed numpy 2.0.1 into the build environment.

The test environment continues to use Intel channel and thus 3.9, 3.10, and 3.11 install numpy 1.26.4 into test environment from there, while for Python 3.12 the latest numpy gets installed from conda-forge channel.

ekomarova:
@oleksandr-pavlyk I believe that the reason is that in internal CI we have a strict channels priority `-c gold -c conda-forge` and the presence of only a certain max numpy version (1.26.4) on `gold` channel. And if numpy is missing from the host, then conda in internal CI logic resolved the environment in favor of the first available channel, and in our case it is exactly `gold` with max numpy 1.26, because this version is consistent with the current environment and there is no reason to search another numpy on conda-forge. If there were no strict channel priority or conda-forge was the first (and only one) in the channel chain, then this problem would most likely not exist

ekomarova:
It looks like in order to make this work in internal CI, we need to rebuild numpy from conda-forge with 2025 mkl, otherwise we will get conflicts. We can plan this activity

ekomarova:
Just for information: building numpy 2.0 with conda-forge recipe with 2025 mkl is WIP

ekomarova:
We expect a built package in the next few days

ekomarova:
We managed to fix the windows problem. Next, I plan to test this PR with built numpy 2.0, and then add the status

ekomarova:
The package is hosted on a separate channel https://af01p-igk.devtools.intel.com/ui/repos/tree/General/idp-conda-pkgserver-igk-local/numpy, but I can't test this PR yet due to channel problems. As soon as we fix them, I'll restart the checks

ekomarova:
We need to fix something in CI logic in order for the conda build to pass, but in general this PR is now being built with numpy 2.0.1, and tested with 1.26.4 in internal CI

ekomarova (approved):

ndgrigorian (approved):
LGTM, I think we can merge this.

PR #1762, opened 2024-07-26
Fix crash testing with sycl nightly runtime

oleksandr-pavlyk:
This PR improves SYCLInterface functions `DPCTLDevice_CreateSubDevicesEqually`, `DPCTLDevice_CreateSubDevicesByCounts`, and `DPCTLDevice_CreateSubDevicesByAffinity`. 

These now check whether given device supports respective partition property before attempting it. If the input device does not support it, `nullptr` is returned without error logged. 

Furthermore, `DPCTLDevice_CreateSubDevicesByAffinity` checks whether the input device supports partitioning over request affinity domain, and returns `nullptr` without error logged if it does not.

This changes became necessary since nightly SYCL bundles are now built with assertions on, and runtime hits an assertion specifying `info::partition_affinity_domain::not_applicable` which is not meant to be used as an input.


ndgrigorian (approved):
Nightly build now works so LGTM!

PR #1773, opened 2024-07-30
Drop use of soversion for dpctl sycl interface

oleksandr-pavlyk:
Turns off SOVERSION-ing of libDPCTLSyclInterface

Library SOVersioning is used to allow libraries of different versions
to coexist in the same layout. Since libDPCTLSyclInterface library
is being installed into site-packages/dpctl, and since Python does
not support multiple versions of the same package installed, such a
situation where SOVersioning is required never arises.

SOVersioning causes creation of symbolic links which are not supported
by Python wheels (symbolic links are followed and 3 hard identical
copies of the library are bing created) and only cause wheel size
to be overinflated.

With all this said, it is better to not set VERSION/SOVERSION
target property.


ndgrigorian (approved):

ZzEeKkAa (approved):
Checked locally. Everything works smooth if dpnp is rebuilt on top on these changes and numba-dpex library import updates name from `*.so.0` to `*.so` at `load_dpctl_sycl_interface`

PR #1776, opened 2024-07-31
Add missing include to fix build break

oleksandr-pavlyk:
"libsyclinterface/source/dpctl_sycl_queue_manager.cpp" used `std::stringstream` but was not including `sstream` header. 

This caused build break in internal CI with forthcoming compiler.


ekomarova (approved):
Thanks for fixing this!

ndgrigorian (approved):
Builds now work, LGTM!

PR #1777, opened 2024-07-31
Fix testing step of conda-package workflow

oleksandr-pavlyk:
Add setuptools to list of packages to be installed into test environment in workflow.

This aligns it with test requirements section of meta.yaml. Setuptools is required by tests that check if an extension can be built using dpctl's headers.


oleksandr-pavlyk:
This change fixes error in executing `test_linux` step in https://github.com/IntelPython/dpctl/actions/runs/10182576343

ndgrigorian (approved):

PR #1778, opened 2024-08-01
Implement take_along_axis function per Python Array API

oleksandr-pavlyk:
The function is planned for Python Array API 2024.12 specification.

### DIFF
--- a/docs/doc_sources/api_reference/dpctl/tensor.indexing_functions.rst
+++ b/docs/doc_sources/api_reference/dpctl/tensor.indexing_functions.rst
@@ -15,3 +15,4 @@ by either integral arrays of indices or boolean mask arrays.
     place
     put
     take
+    take_along_axis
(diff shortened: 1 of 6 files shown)


antonwolfy (commented):
I've opened a PR to reuse the new dpctl function in [dpnp-1969](https://github.com/IntelPython/dpnp/pull/1969). The existing and new dpnp tests are passed with that dpctl PR locally.

ndgrigorian:
@oleksandr-pavlyk 
I am going to try to take a look at this today.

One thing that I did realize, though: would it make sense to add `put_along_axis` as an inverse function? Array API obviously won't yet due to mutation issues, but we already have `put`, it seems like a good idea, maybe for a separate PR.

oleksandr-pavlyk:
> One thing that I did realize, though: would it make sense to add `put_along_axis` as an inverse function? 

Yes, makes total sense. Let's do it in a separate PR though.

ndgrigorian (approved):
I've tested the branch out and it works, this LGTM.

Thank you @oleksandr-pavlyk !

PR #1781, opened 2024-08-02
Support `kDLCPU` DLDeviceType for `from_dlpack` and `__dlpack__`

ndgrigorian:
This PR proposes supporting the `kDLCPU` DLDeviceType as a targeted device in `from_dlpack` and `__dlpack__`.

This is accomplished by `__dlpack__` by copying the array data to the host, wrapping it in a NumPy array, and then returning a capsule for this NumPy array.

In `from_dlpack`, a NumPy array is also returned for devices with `__dlpack_device__= (DLDeviceType.kDLCPU, 0)`. This means that `from_dlpack(x_np)` for some NumPy array `x_np` is effectively a no-op with extra validation steps returning a new Python object wrapping the same memory with zero-copy.

New utility functions `numpy_to_dlpack_versioned_capsule`, `_numpy_array_interface_from_dl_tensor`, and the class `_numpy_array_interface_wrapper` are introduced to enable the new functionality.

This PR also refactors `from_dlpack_versioned_capsule` into `from_dlpack_capsule`.

### DIFF
--- a/dpctl/CMakeLists.txt
+++ b/dpctl/CMakeLists.txt
@@ -125,8 +125,6 @@ function(build_dpctl_ext _trgt _src _dest)
             )
         endif()
     endif()
-    # needed to prevent warnings from Cython caused by cimporting from Numpy
-    target_compile_definitions(${_trgt} PRIVATE NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION)
     target_link_libraries(${_trgt} PRIVATE Python::NumPy)
     if (DPCTL_GENERATE_COVERAGE)
         target_compile_definitions(${_trgt} PRIVATE CYTHON_TRACE=1 CYTHON_TRACE_NOGIL=1)
--- a/dpctl/CMakeLists.txt
+++ b/dpctl/CMakeLists.txt
@@ -125,6 +125,8 @@ function(build_dpctl_ext _trgt _src _dest)
             )
         endif()
     endif()
+    # needed to prevent warnings from Cython caused by cimporting from Numpy
+    target_compile_definitions(${_trgt} PRIVATE NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION)
     target_link_libraries(${_trgt} PRIVATE Python::NumPy)
     if (DPCTL_GENERATE_COVERAGE)
         target_compile_definitions(${_trgt} PRIVATE CYTHON_TRACE=1 CYTHON_TRACE_NOGIL=1)
--- a/dpctl/tensor/_dlpack.pyx
+++ b/dpctl/tensor/_dlpack.pyx
@@ -515,27 +515,43 @@ cpdef numpy_to_dlpack_versioned_capsule(ndarray npy_ary, bint copied):
     cdef int i = 0
     cdef int device_id = -1
     cdef Py_ssize_t byte_offset = 0
+    cdef int itemsize = npy_ary.itemsize
 
     dlmv_tensor = <DLManagedTensorVersioned *> stdlib.malloc(
         sizeof(DLManagedTensorVersioned))
     if dlmv_tensor is NULL:
         raise MemoryError(
-            "to_dlpack_versioned_capsule: Could not allocate memory "
+            "numpy_to_dlpack_versioned_capsule: Could not allocate memory "
             "for DLManagedTensorVersioned"
         )
-    shape_strides_ptr = <int64_t *>stdlib.malloc((sizeof(int64_t) * 2) * nd)
+
+    is_c_contiguous = npy_ary.flags["C"]
+    shape = npy_ary.ctypes.shape_as(ctypes.c_int64)
+    strides = npy_ary.ctypes.strides_as(ctypes.c_int64)
+    if not is_c_contiguous:
+        if npy_ary.size != 1:
+            for i in range(nd):
+                if shape[i] != 1 and strides[i] % itemsize != 0:
+                    stdlib.free(dlmv_tensor)
+                    raise BufferError(
+                        "numpy_to_dlpack_versioned_capsule: DLPack cannot encode "
+                        "an array if strides are not a multiple of itemsize"
+                    )
+        shape_strides_ptr = <int64_t *>stdlib.malloc((sizeof(int64_t) * 2) * nd)
+    else:
+        # no need to pass strides in this case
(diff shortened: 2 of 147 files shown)


dpctl/CMakeLists.txt
```diff
@@ -125,6 +125,8 @@ function(build_dpctl_ext _trgt _src _dest)
             )
         endif()
     endif()
+    # needed to prevent warnings from Cython caused by cimporting from Numpy
+    target_compile_definitions(${_trgt} PRIVATE NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION)
```
oleksandr-pavlyk, line 129: I think a better approach would to include `npy_no_deprecated_api` header before `cimport numpy`, see

https://github.com/IntelPython/mkl_random/blob/8bc5e7c6b121df883e77ad53e02e943906922c7d/mkl_random/mklrand.pyx#L39

dpctl/tensor/_dlpack.pyx
```diff
@@ -475,6 +481,108 @@ cpdef to_dlpack_versioned_capsule(usm_ndarray usm_ary, bint copied):
...
+        raise MemoryError(
+            "to_dlpack_versioned_capsule: Could not allocate memory "
+            "for shape/strides"
+        )
+    # this can be a separate function for handling shapes and strides
+    shape = npy_ary.ctypes.shape_as(ctypes.c_int64)
+    strides = npy_ary.ctypes.strides_as(ctypes.c_int64)
+    for i in range(nd):
+        shape_strides_ptr[i] = shape[i]
+        shape_strides_ptr[nd + i] = strides[i] // npy_ary.itemsize
```
oleksandr-pavlyk, line 538: In NumPy the `strides` need not be a multiple of `npy_ary.itemsize`. We must check whether this is the case, and raise and exception, `BufferError` if strides are not multiple of itemsize. Such arrays can not be encoded using dlpack.

dpctl/tensor/_usmarray.pyx
```diff
@@ -1142,13 +1142,27 @@ cdef class usm_ndarray:
...
-                        raise NotImplementedError(
-                            "targeting a device with `__dlpack__` is not "
-                            "currently implemented"
-                        )
+                        if copy == False:
+                            raise BufferError(
+                                "array cannot be placed on the requested device without a copy"
+                            )
+                        if dl_device[0] == (DLDeviceType.kDLCPU):
+                            assert dl_device[1] == 0
```
oleksandr-pavlyk, line 1152:
```suggestion
                        if len(dl_device) == 2 and dl_device == (DLDeviceType.kDLCPU, 0):
```

It is better to not rely on assert in the production code to generate an error

dpctl/tensor/_dlpack.pyx
ndgrigorian: I've pushed a change which resolves this, following along with how NumPy implemented it in their dlpack implementation (ignoring for size-1 axes).

oleksandr-pavlyk:
I think we must raise an exception on attempt to import `np.ndarray` instance to `device=(kDLoneAPI, num)`, or support it properly. At the moment, we silently return value the `np.ndarray` back:

```

In [2]: x = dpt.ones((2, 3))

In [3]: y = dpt.from_dlpack(x, device=(1,0))

In [4]: type(y)
Out[4]: numpy.ndarray

In [5]: z = dpt.from_dlpack(y)

In [6]: type(z)
Out[6]: numpy.ndarray

In [7]: dpt.from_dlpack(z, device=x.__dlpack_device__())
Out[7]:
array([[1., 1., 1.],
       [1., 1., 1.]], dtype=float32)

In [8]: type(_)
Out[8]: numpy.ndarray

In [9]: x.__dlpack_device__()
Out[9]: (<DLDeviceType.kDLOneAPI: 14>, 2)
```

I also think we ought to make `DLDeviceType` enum documented and accessible in a public namespace. I think `dpctl.tensor` is appropriate.

oleksandr-pavlyk:
Actually, we must support importing dlpack capsules with `kDLCPU` to allow copying from an arbitrary importer (that supports exporting to "kDLCPU") via host.

ndgrigorian:
> I also think we ought to make `DLDeviceType` enum documented and accessible in a public namespace. I think `dpctl.tensor` is appropriate.

In working on this, I realized, a page with `DLDeviceType` enum documented would be the perfect place for other constants (`dpt.pi`, `dpt.e`, etc.)

dpctl/tensor/_dlpack.pyx
```diff
@@ -718,6 +718,11 @@ class _numpy_array_interface_wrapper:
         self._memory_owner = memory_owner
 
 
+cdef bint _is_kdlcpu_device(DLDevice *dev):
+    "Check if DLTensor.DLDevice denotes (kDLCPU, 0)"
+    return (dev.device_type == kDLCPU) and (dev.device_id == 0)
```
oleksandr-pavlyk, line 723: From C perspective, this reads odd. In C, this should translate to `dev->device_type` and `dev->device_id`. To be able to use struct member access like this, my reaction would be to write `dev[0].device_type` and `dev[0].device_id`. 

But perhaps Cython handles this gracefully. I would check generated `cxx` file to confirm that correct C++ code is being generated.

ndgrigorian: It looks correct in `_dlpack.cxx`

```
  __Pyx_TraceLine(723,0,__PYX_ERR(0, 723, __pyx_L1_error))
  __pyx_t_2 = (__pyx_v_dev->device_type == kDLCPU);
  if (__pyx_t_2) {
  } else {
    __pyx_t_1 = __pyx_t_2;
    goto __pyx_L3_bool_binop_done;
  }
  __pyx_t_2 = (__pyx_v_dev->device_id == 0);
  __pyx_t_1 = __pyx_t_2;
  __pyx_L3_bool_binop_done:;
  __pyx_r = __pyx_t_1;
  goto __pyx_L0;
```

But it wouldn't take much to go ahead and change it anyway.

oleksandr-pavlyk (approved):
Thank you @ndgrigorian for working on this. It looks good to me, so I am approving. 

I am going to ping @seberg to bring to his attention code for producing/consuming `DLManagedTensorVersioned` capsule in `dpctl/tensor/_dlpack.pyx`.

We merge the PR though and address his feedback, if any, later.

PR #1782, opened 2024-08-05
Add usm memory constructor from existing allocation

oleksandr-pavlyk:
This PR add `dpctl::memory::usm_memory` constructor to create Python memory object from natively made allocation.

The signature is 

```cpp
usm_memory(void *usm_ptr, size_t nbytes, sycl::queue &q, std::shared_ptr<void> owner)
```

where `usm_ptr` is the pointer to USM-allocation, `nbytes` the size of allocation in bytes, `q` the queue associated with the Python memory object, and `owner` is the shared pointer whose destructor frees USM allocation via its custom deleter argument.

Example of using such constructor is added to `examples/pybind11/external_usm_allocation` extension.

### DIFF
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -764,7 +764,7 @@ public:
             reinterpret_cast<DPCTLSyclQueueRef>(q_uptr.get());
 
         auto vacuous_destructor = []() {};
-        py::object mock_owner = py::capsule(vacuous_destructor);
+        py::capsule mock_owner(vacuous_destructor);
 
         // create memory object owned by mock_owner, it is a new reference
         PyObject *_memory =
@@ -773,14 +773,13 @@ public:
 
         using py_uptrT =
             std::unique_ptr<PyObject, decltype(ref_count_decrementer)>;
-        auto memory_uptr = py_uptrT(_memory, ref_count_decrementer);
 
         if (!_memory) {
             throw py::error_already_set();
         }
 
-        std::shared_ptr<void> *opaque_ptr = nullptr;
-        opaque_ptr = new std::shared_ptr<void>(shptr);
+        auto memory_uptr = py_uptrT(_memory, ref_count_decrementer);
+        std::shared_ptr<void> *opaque_ptr = new std::shared_ptr<void>(shptr);
 
         Py_MemoryObject *memobj = reinterpret_cast<Py_MemoryObject *>(_memory);
         // replace mock_owner capsule as the owner
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -27,6 +27,7 @@
 
 #include "dpctl_capi.h"
 #include <complex>
+#include <exception>
 #include <memory>
 #include <pybind11/pybind11.h>
 #include <sycl/sycl.hpp>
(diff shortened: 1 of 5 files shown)


dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -748,6 +748,64 @@ class usm_memory : public py::object
...
+
+        bool is_ok =
+            (PyObject_TypeCheck(
+                 _memory, ::dpctl::detail::dpctl_capi::get().Py_MemoryType_) !=
+             0);
+
+        if (is_ok) {
+            std::shared_ptr<void> *opaque_ptr =
+                new std::shared_ptr<void>(shptr);
+            if (!opaque_ptr) {
```
AlexanderKalistratov, line 782: If I'm not mistaken `new` throws `std::bad_alloc` instead of returning `nullptr` in case it fails to allocate memory.
Unless you specified `std::nothrow`

https://en.cppreference.com/w/cpp/memory/new/nothrow

So you wouldn't reach that point and wouldn't be able to free memory unless you are using `unique_ptr`. Which is not very important, since if we are out of memory application would crash anyway.

examples/pybind11/external_usm_allocation/external_usm_allocation/_usm_alloc_example.cpp
```diff
@@ -120,6 +121,42 @@ py::list tolist(DMatrix &m)
...
+};
+
+dpctl::memory::usm_memory make_zeroed_device_memory(size_t nbytes,
+                                                    sycl::queue &q)
+{
+    char *data = sycl::malloc_device<char>(nbytes, q);
+    q.memset(data, 0, nbytes).wait();
+
+    USMDeleter _deleter(q);
+    auto shptr = std::shared_ptr<void>(data, _deleter);
```
AlexanderKalistratov, line 155: I'm usually using the following construct for device memory:
```C++
template<class T> using DeviceMem = std::unique_ptr<T[], std::function<void (T*)>>;

template<class T> DeviceMem<T> malloc_device(int64_t count, cl::sycl::queue& queue)
{
    return DeviceMem<T>(sycl::malloc_device<T>(count, queue), [queue](T* mem)
    {
        sycl::free(mem, queue);
    });
}
```

So you don't need to create `USMDeleter` and replace it with lambda which is much less verbose.

AlexanderKalistratov (approved):
I have small style comments which can be ignored.

examples/pybind11/external_usm_allocation/external_usm_allocation/_usm_alloc_example.cpp
oleksandr-pavlyk: Thank you for the suggestion @AlexanderKalistratov . I have pushed a change to the example to use lambda with capture instead of `USMDeleter`.

dpctl/apis/include/dpctl4pybind11.hpp
oleksandr-pavlyk: I changed the code to expect `new` to throw an exception. Please have another look

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -748,6 +748,64 @@ class usm_memory : public py::object
...
+     *  lifetime of the USM allocation.
+     */
+    usm_memory(void *usm_ptr,
+               size_t nbytes,
+               const sycl::queue &q,
+               std::shared_ptr<void> shptr)
+    {
+        auto const &api = ::dpctl::detail::dpctl_capi::get();
+        DPCTLSyclUSMRef usm_ref = reinterpret_cast<DPCTLSyclUSMRef>(usm_ptr);
+        sycl::queue *q_ptr = new sycl::queue(q);
```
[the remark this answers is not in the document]
oleksandr-pavlyk: We need to use shared pointer because it is used by Python host code, and by host tasks to ensure life-time of USM allocation extends past completion of kernel execution submissions.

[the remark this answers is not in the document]
AlexanderKalistratov: What I was talking about is to use `unique_ptr` to manage created objects and free it in case of failure automatically:
```C++
auto q_ptr = std::make_unique(new sycl::queue(q));
DPCTLSyclQueueRef QRef = reinterpret_cast<DPCTLSyclQueueRef>(q_ptr.get());
...

if (is_ok) {
    ...
    m_ptr = _memory;
    q_ptr.release();
}
```
So we don't need to call
```C++
delete q_ptr;
```
by hands. And memory would be freed in any scenario except one where we are calling `release`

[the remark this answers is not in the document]
oleksandr-pavlyk: Thank you @AlexanderKalistratov ! I like the suggestion. I am going to push the change shortly

oleksandr-pavlyk:
@AlexanderKalistratov @ndgrigorian I have applied suggested changes, rebased and squashed commits. This is ready for next round of reviewing.

AlexanderKalistratov (approved):
lgtm

ndgrigorian (approved):
Looks good to me, I ran the example on CUDA as well.

Issue #1769, opened 2024-07-29
Shape setter does not work with integer scalar


antonwolfy:
The below example reproduces an issue with passing integer scalar to the shape setter:
```python
import numpy, dpctl, dpctl.tensor as dpt

dpctl.__version__
# Out: '0.18.0dev0+158.g7450558d25'

a = dpt.usm_ndarray((2, 3))
a.shape = 6
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 a.shape = 6

File dpctl/tensor/_usmarray.pyx:579, in dpctl.tensor._usmarray.usm_ndarray.shape.__set__()

TypeError: object of type 'int' has no len()

# but works in numpy
a = numpy.ndarray((2, 3))
a.shape = 6
```

oleksandr-pavlyk:
The `.shape` setter requires value to be a tuple, or a list, or a object with `__len__`.  This is different from NumPy's behavior:

```

In [30]: a = dpt.ones((2,3))

In [31]: class Six:
    ...:     def __init__(self, dim=1):
    ...:         self.v = (1,) * (dim - 1) + (6,)
    ...:     def __len__(self):
    ...:         return len(self.v)
    ...:     def __iter__(self):
    ...:         return iter(self.v)
    ...:

In [32]: a.shape = Six(3)

In [33]: a
Out[33]: usm_ndarray([[[1., 1., 1., 1., 1., 1.]]], dtype=float32)

In [34]: a.shape
Out[34]: (1, 1, 6)

In [35]: import numpy as np

In [36]: b = np.ones((2,3))

In [37]: b.shape = Six(3)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[37], line 1
----> 1 b.shape = Six(3)

TypeError: expected a sequence of integers or a single integer, got '<__main__.Six object at 0x7f7ee86258e0>'

```

antonwolfy:
Should that be properly clarified in the documentation?
Since it might be unclear that `.shape` setter and `reshape` method behaves differently in case when the reshaping to the requested dimensions can be returned as view.

PR #1786, opened 2024-08-05
Fix gh-1769 - setting shape by a scalar

oleksandr-pavlyk:
This PR resolves gh-1769

shape setter now accepts integer values.

### DIFF
--- a/dpctl/tensor/_usmarray.pyx
+++ b/dpctl/tensor/_usmarray.pyx
@@ -563,6 +563,22 @@ cdef class usm_ndarray:
 
     @shape.setter
     def shape(self, new_shape):
+        """
+        Modifies usm_ndarray instance in-place by changing its metadata
+        about the shape and the strides of the array, or raises
+        `AttributeError` exception if in-place change is not possible.
+
+        Args:
+            new_shape: (tuple, int)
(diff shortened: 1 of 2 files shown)


ndgrigorian (commented):
Does not work for -1, where NumPy does

```
In [2]: x = dpt.arange(100, dtype="i4")

In [3]: x.shape = -1
Cell In[3], line 1
----> 1 x.shape = -1

File ~/repos/dpctl/dpctl/tensor/_usmarray.pyx:592, in dpctl.tensor._usmarray.usm_ndarray.shape.__set__()
    590 size = shape_to_elem_count(self.nd_, self.shape_)
    591 if not np.prod(new_shape) == size:
--> 592     raise TypeError(
    593         f"Can not reshape array of size {self.size} into {new_shape}"
    594     )

TypeError: Can not reshape array of size 100 into (-1,)
```

oleksandr-pavlyk:
I added a docstring to document limitations of `usm_ndarray.shape` setter. The intent was never to match NumPy, although it is certainly possible.

ndgrigorian (approved):
LGTM!

antonwolfy (approved):
The dpnp scope of tests passed with the change. Thank you @oleksandr-pavlyk

PR #1789, opened 2024-08-07
Enhance from_dlpack to support imported kDLCPU data to kDLOneAPI

oleksandr-pavlyk:
This PR contributes to gh-1781 to support importing objects resident on other devices to oneAPI devices by copying via host.

```

In [2]: x = dpt.ones((2, 3))

In [3]: x.device
Out[3]: Device(level_zero:gpu:0)

In [4]: y = dpt.from_dlpack(x, device=(1,0))

In [5]: type(y)
Out[5]: numpy.ndarray

In [6]: w1 = dpt.from_dlpack(y, device="opencl:gpu")

In [7]: w1.device
Out[7]: Device(opencl:gpu:0)

In [8]: w2 = dpt.from_dlpack(y, device="gpu")

In [9]: w2.device
Out[9]: Device(level_zero:gpu:0)

In [10]: w3 = dpt.from_dlpack(y, device="cpu")

In [11]: w3.device, type(w3)
Out[11]: (Device(opencl:cpu:0), dpctl.tensor._usmarray.usm_ndarray)
```

### DIFF
--- a/dpctl/tensor/_dlpack.pyx
+++ b/dpctl/tensor/_dlpack.pyx
@@ -168,7 +168,7 @@ cdef void _managed_tensor_versioned_deleter(DLManagedTensorVersioned *dlmv_tenso
         stdlib.free(dlmv_tensor)
 
 
-cdef object _get_default_context(c_dpctl.SyclDevice dev) except *:
+cdef object _get_default_context(c_dpctl.SyclDevice dev):
     try:
         default_context = dev.sycl_platform.default_context
     except RuntimeError:
@@ -178,7 +178,7 @@ cdef object _get_default_context(c_dpctl.SyclDevice dev) except *:
     return default_context
 
 
-cdef int get_parent_device_ordinal_id(c_dpctl.SyclDevice dev) except *:
+cdef int get_parent_device_ordinal_id(c_dpctl.SyclDevice dev) except -1:
     cdef DPCTLSyclDeviceRef pDRef = NULL
     cdef DPCTLSyclDeviceRef tDRef = NULL
     cdef c_dpctl.SyclDevice p_dev
@@ -201,7 +201,7 @@ cdef int get_parent_device_ordinal_id(c_dpctl.SyclDevice dev) except *:
 
 cdef int get_array_dlpack_device_id(
     usm_ndarray usm_ary
-) except *:
+) except -1:
     """Finds ordinal number of the parent of device where array
     was allocated.
     """
@@ -935,6 +935,32 @@ cpdef object from_dlpack_capsule(object py_caps):
             "The DLPack tensor resides on unsupported device."
         )
 
+cdef usm_ndarray _to_usm_ary_from_host_blob(object host_blob, dev : Device):
+    q = dev.sycl_queue
+    np_ary = np.asarray(host_blob)
+    dt = np_ary.dtype
+    if dt.char in "dD" and q.sycl_device.has_aspect_fp64 is False:
+        Xusm_dtype = (
+            "float32" if dt.char == "d" else "complex64"
+        )
+    else:
+        Xusm_dtype = dt
+    usm_mem = dpmem.MemoryUSMDevice(np_ary.nbytes, queue=q)
+    usm_ary = usm_ndarray(np_ary.shape, dtype=Xusm_dtype, buffer=usm_mem)
+    usm_mem.copy_from_host(np.reshape(np_ary.view(dtype="u1"), -1))
+    return usm_ary
+
+
+# only cdef to make it private
+cdef object _create_device(object device, object dl_device):
+    if isinstance(device, Device):
+        return device
+    elif isinstance(device, dpctl.SyclDevice):
+        return Device.create_device(device)
+    else:
+        root_device = dpctl.SyclDevice(str(<int>dl_device[1]))
+        return Device.create_device(root_device)
+
 
 def from_dlpack(x, /, *, device=None, copy=None):
     """ from_dlpack(x, /, *, device=None, copy=None)
@@ -943,7 +969,7 @@ def from_dlpack(x, /, *, device=None, copy=None):
     object ``x`` that implements ``__dlpack__`` protocol.
 
     Args:
-        x (Python object):
+        x (object):
             A Python object representing an array that supports
             ``__dlpack__`` protocol.
         device (Optional[str,
@@ -959,7 +985,8 @@ def from_dlpack(x, /, *, device=None, copy=None):
             returned by :attr:`dpctl.tensor.usm_ndarray.device`, or a
             2-tuple matching the format of the output of the ``__dlpack_device__``
             method, an integer enumerator representing the device type followed by
-            an integer representing the index of the device.
+            an integer representing the index of the device. The only supported
+            :enum:`dpctl.tensor.DLDeviceType` types are "kDLCPU" and "kDLOneAPI".
             Default: ``None``.
         copy (bool, optional)
             Boolean indicating whether or not to copy the input.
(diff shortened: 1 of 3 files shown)


dpctl/tensor/_dlpack.pyx
```diff
@@ -1005,31 +1005,60 @@ def from_dlpack(x, /, *, device=None, copy=None):
...
     except TypeError:
-        dlpack_capsule = dlpack_attr()
-        return from_dlpack_capsule(dlpack_capsule)
+        x_dldev = dlpack_dev_attr()
+        if (dl_device is None) or (dl_device == x_dldev):
+            dlpack_capsule = dlpack_attr()
+            return from_dlpack_capsule(dlpack_capsule)
+        # must copy via host
+        if copy is False:
+            raise ValueError(
```
ndgrigorian, line 1040: Array API recommends this case to be a `BufferError` for `__dlpack__`, I think that we should be consistent here.

dpctl/tensor/_dlpack.pyx
```diff
@@ -1005,31 +1005,60 @@ def from_dlpack(x, /, *, device=None, copy=None):
...
+                "Importing data via DLPack requires copying, but copy=False was provided"
+            )
+        if dl_device[0] != device_OneAPI:
+            raise ValueError(f"Can not import to requested device {dl_device}")
+        if x_dldev == (device_CPU, 0):
+            host_blob = x
+        else:
+            dlpack_capsule = dlpack_attr(max_version=(1, 0), dl_device=(device_CPU, 0), copy=copy)
```
ndgrigorian, line 1048:
```suggestion
            dlpack_capsule = dlpack_attr(max_version=(1, 0), dl_device=(device_CPU, 0), copy=copy)
```
```suggestion
            dlpack_capsule = dlpack_attr(max_version=get_build_dlpack_version(), dl_device=(device_CPU, 0), copy=copy)
```

dpctl/tensor/_dlpack.pyx
```diff
@@ -1005,31 +1005,60 @@ def from_dlpack(x, /, *, device=None, copy=None):
...
+        if (dl_device is None) or (dl_device == x_dldev):
+            dlpack_capsule = dlpack_attr()
+            return from_dlpack_capsule(dlpack_capsule)
+        # must copy via host
+        if copy is False:
+            raise ValueError(
+                "Importing data via DLPack requires copying, but copy=False was provided"
+            )
+        if dl_device[0] != device_OneAPI:
+            raise ValueError(f"Can not import to requested device {dl_device}")
```
ndgrigorian, line 1044:
```suggestion
            raise ValueError(f"Can not import to requested device {dl_device}")
```
```suggestion
            raise BufferError(f"Can not import to requested device {dl_device}")
```

dpctl/tensor/_dlpack.pyx
oleksandr-pavlyk: My thinking was we should request the minimal version that supports `dl_device` and `copy` keywords.

dpctl/tensor/_dlpack.pyx
```diff
@@ -1005,31 +1005,60 @@ def from_dlpack(x, /, *, device=None, copy=None):
...
                 if not isinstance(device, dpctl.SyclDevice):
                     d = Device.create_device(device).sycl_device
-                    dl_device = (device_OneAPI, get_parent_device_ordinal_id(<c_dpctl.SyclDevice>d))
                 else:
-                    dl_device = (device_OneAPI, get_parent_device_ordinal_id(<c_dpctl.SyclDevice>device))
+                    d = device
+                dl_device = (device_OneAPI, get_parent_device_ordinal_id(<c_dpctl.SyclDevice>d))
         dlpack_capsule = dlpack_attr(max_version=get_build_dlpack_version(), dl_device=dl_device, copy=copy)
         return from_dlpack_capsule(dlpack_capsule)
     except TypeError:
```
ndgrigorian, line 1033: The intention of this exception (per array API) is that if `max_version` keyword isn't supported, a TypeError is thrown, and the legacy path is taken. Leaning on this TypeError to move a NumPy array to kDLOneAPI device could somewhat undermine that.

We will need either another try-except block or a check for the `max_version` keyword to remain conformant.

dpctl/tensor/_dlpack.pyx
oleksandr-pavlyk: The docstring of `from_dlpack` says we should raise `ValueError` if "the input array resides on an unsupported device". I suppose that should change too.

ndgrigorian: Good point

dpctl/tensor/_dlpack.pyx
ndgrigorian: Maybe the best way to do this would be to capture the exception in a local variable and check if the exception is specifically `__dlpack__() got an unexpected keyword argument 'max_version'` then take the legacy path, otherwise do this.

oleksandr-pavlyk: Yes, agreed. The `dl_device` has been added at the same time as the support for `max_version` though. 

Perhaps we should expect `__dlpack__` call to raise `BufferError` if the requested `dl_device` could not be supported and trigger copying via host logic.

ndgrigorian: > Yes, agreed. The `dl_device` has been added at the same time as the support for `max_version` though.
> 
> Perhaps we should expect `__dlpack__` call to raise `BufferError` if the requested `dl_device` could not be supported and trigger copying via host logic.

This would be a good solution, though we would still have to be careful with the source of the exception, as `copy` keyword also throws a `BufferError` when it can't be honored.

ndgrigorian:
@oleksandr-pavlyk 
A small (very, very niche) problem I've found:
`from_dlpack_capsule` itself can raise `BufferError`. Under the right conditions it's possible to circumvent some otherwise unacceptable operations by `from_dlpack` because we try-except `BufferError`:
```
In [1]: import dpctl.tensor as dpt, dpctl, numpy as np, dpctl.tensor._dlpack as _dlp

In [2]: x_np = np.ones((10, 10), dtype="i4")

In [3]: x_np.strides = (x_np.strides[0] - 1, x_np.strides[1]) # strides are no longer multiple of itemsize

In [4]: dpt.from_dlpack(x_np)
File ~/repos/dpctl/dpctl/tensor/_dlpack.pyx:1047, in dpctl.tensor._dlpack.from_dlpack()
   1046         dl_device = (device_OneAPI, get_parent_device_ordinal_id(<c_dpctl.SyclDevice>d))
-> 1047 dlpack_capsule = dlpack_attr(max_version=get_build_dlpack_version(), dl_device=dl_device, copy=copy)
   1048 return from_dlpack_capsule(dlpack_capsule)

TypeError: __dlpack__() got an unexpected keyword argument 'max_version'

During handling of the above exception, another exception occurred:

BufferError                               Traceback (most recent call last)
Cell In[4], line 1
----> 1 dpt.from_dlpack(x_np)

File ~/repos/dpctl/dpctl/tensor/_dlpack.pyx:1053, in dpctl.tensor._dlpack.from_dlpack()
   1051 x_dldev = dlpack_dev_attr()
   1052 if (dl_device is None) or (dl_device == x_dldev):
-> 1053     dlpack_capsule = dlpack_attr()
   1054     return from_dlpack_capsule(dlpack_capsule)
   1055 # must copy via host

BufferError: DLPack only supports strides which are a multiple of itemsize.

In [5]: dpt.from_dlpack(x_np, device=(14, 1))
Out[5]:
usm_ndarray([[       1,        1,        1,        1,        1,        1,
                     1,        1,        1,        1],
             [     256,      256,      256,      256,      256,      256,
                   256,      256,      256,      256],
             [   65536,    65536,    65536,    65536,    65536,    65536,
                 65536,    65536,    65536,    65536],
             [16777216, 16777216, 16777216, 16777216, 16777216, 16777216,
              16777216, 16777216, 16777216, 16777216],
             [       1,        1,        1,        1,        1,        1,
                     1,        1,        1,        1],
             [     256,      256,      256,      256,      256,      256,
                   256,      256,      256,      256],
             [   65536,    65536,    65536,    65536,    65536,    65536,
                 65536,    65536,    65536,    65536],
             [16777216, 16777216, 16777216, 16777216, 16777216, 16777216,
              16777216, 16777216, 16777216, 16777216],
             [       1,        1,        1,        1,        1,        1,
                     1,        1,        1,        1],
             [     256,      256,      256,      256,      256,      256,
                   256,      256,      256,      256]], dtype=int32)
```

oleksandr-pavlyk:
Thanks for the example @ndgrigorian. I have pushed the fix, by replacing `host_blob = x` with ``host_blob = result_of_roundtripping_through_dlpack(x)``. Added a test based in your example.

oleksandr-pavlyk:
@ndgrigorian Once this is ready to go, I intend to squash all changes into few commits. Please rereview

dpctl/tensor/_dlpack.pyx
```diff
@@ -1010,31 +1036,90 @@ def from_dlpack(x, /, *, device=None, copy=None):
...
+            x_dldev = dlpack_dev_attr()
+            if (dl_device is None) or (dl_device == x_dldev):
+                dlpack_capsule = dlpack_attr()
+                return from_dlpack_capsule(dlpack_capsule)
+            # must copy via host
+            if copy is False:
+                raise BufferError(
+                    "Importing data via DLPack requires copying, but copy=False was provided"
+                )
+            if x_dldev == (device_CPU, 0) and dl_device[0] == device_OneAPI:
```
ndgrigorian, line 1098: Shouldn't this be in reverse? I thought the plan was to allow imported `kDLOneAPI` arrays to be moved to CPU, but not the converse.

ndgrigorian: In any case, if we don't make it bidirectional, it would be good to note somewhere why, so that users aren't confused.

oleksandr-pavlyk: This part of code executed if `input_array.__dlpack__` does not support `max_version`, `dl_device`, and `copy` keywords, but `from_dlpack` was given a non-default `device` keyword. We can only support this is the requested `device` coincides with `input_array.__dlpack_version__()` (checked earlier), or if the requested `device` is our own, and the input_array is either on host, or is on another oneAPI device (this `dpnp` that does not yet support `max_version`). 

I am going to push a change to allow for the last scenario as well.

PR #1790, opened 2024-08-07
Add codeowner file in .github

oleksandr-pavlyk:
Add `.github/CODEOWNERS` file.

### DIFF
--- a/.github/workflows/CODEOWNERS
+++ b/.github/workflows/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @oleksandr-pavlyk @ndgrigorian


.github/workflows/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* oleksandr-pavlyk
+* ndgrigorian
```
ndgrigorian, line 2:
```suggestion
*       @oleksandr-pavlyk @ndgrigorian
```
According to the [Github documentation on codw owners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) I think it should look something like this, currently the file says both are invalid

oleksandr-pavlyk:
I will squash and force push. Thanks for looking up the correct syntax @ndgrigorian

PR #1791, opened 2024-08-07
Try/catch wrap a call to deallocate in silent usm_host_allocator class to be used by std::vector for array metadata transfers 

oleksandr-pavlyk:
Introduced class `template <typename T> dpctl::tensor::offset_utils::usm_host_allocator` deriving from `sycl::usm_allocator<T, sycl::alloc::kind::host>` that wraps the call to ``base::deallocate`` in try/catch to prevent crashes due to seemingly benign exceptions thrown by call to `sycl::free` by CPU device runtime which are under investigation.

In case such an exception is caught, a message is printed to `std::cerr`, but the exception is otherwise ignored.

Run pytest with `-s` for testing with nightly sycl bundle to be able to see such message printed.

Also remove use of `--no-sycl-interface-test` option, since DPCTLSyclInterface library is no longer so-versioned.

### DIFF
--- a/.github/workflows/os-llvm-sycl-build.yml
+++ b/.github/workflows/os-llvm-sycl-build.yml
@@ -13,11 +13,11 @@ jobs:
 
     env:
       DOWNLOAD_URL_PREFIX: https://github.com/intel/llvm/releases/download
-      DRIVER_PATH: 2023-WW27
-      OCLCPUEXP_FN: oclcpuexp-2023.16.6.0.28_rel.tar.gz
-      TBB_URL: https://github.com/oneapi-src/oneTBB/releases/download/v2021.9.0/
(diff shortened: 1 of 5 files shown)


oleksandr-pavlyk:
The log of test run with `-s` option does not show message about exception caught during call to `sycl::free` during execution of the crashing test. 

Testing with SYCL OS build internally, the crash is occurring in `test_matmul_nilpotent1[i1]` in unified runtime library. This one must be different,  specific to the architecture of the runner's CPU. It may, perhaps, be masking the problem with 
 `test_matmul_nilpotent1[i1]`. 

I will try to reproduce the issue though, as it keeps consistently reoccurring at the same test.

ndgrigorian (approved):
Nightly build passes now, so I think it would be good to get this in, looks good to me!

PR #1792, opened 2024-08-07
Fix for gh 1785

oleksandr-pavlyk:
Resolves gh-1785

Treat `float` and `complex` types as non-booleans. They used to be treated so because `bool(key)` works for these types.

Changed exception raised for unrecognized key from `TypeError` to `IndexError`,


ndgrigorian (approved):
Looks good to me!

oleksandr-pavlyk:
@vtavana Ping

vtavana (approved):
Thank you, @oleksandr-pavlyk! 
Works fine in `dpnp` as well.

PR #1797, opened 2024-08-12
Replace sycl::free with sycl_free_noexcept

oleksandr-pavlyk:
Replace ``sycl::free`` with ``sycl_free_noexcept``.

Add implementation of ``dpctl::tensor::offset_utils::sycl_free_noexcept`` templated function that calls ``sycl::free`` from within try/catch. The exception is logged to ``std::cerr``, but otherwise ignored.

Replaced calls to `sycl::free` in `host_task` with use of ``sycl_free_noexcept``.

This change should improve stability of test suite run executed on CPU device due to a known issue.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
@@ -32,6 +32,7 @@
 
 #include "dpctl_tensor_types.hpp"
 #include "utils/offset_utils.hpp"
+#include "utils/sycl_alloc_utils.hpp"
 #include "utils/sycl_utils.hpp"
 #include "utils/type_dispatch_building.hpp"
 #include "utils/type_utils.hpp"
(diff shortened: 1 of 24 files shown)


ndgrigorian (commented):
In general this is a good change, but `offset_utils` seems like an unintuitive place to put it. I'm not sure where else it would belong though...

oleksandr-pavlyk:
@ndgrigorian Is `utils/sycl_utils.hpp` more appropriate?

ndgrigorian:
> @ndgrigorian Is `utils/sycl_utils.hpp` more appropriate?

I think that makes more sense.

oleksandr-pavlyk:
@ndgrigorian I created `sycl_alloc_utils.hpp` to house these utility functions, and deployed it everywhere.

ndgrigorian (approved):

PR #1798, opened 2024-08-12
Implements `dpctl.tensor.put_along_axis`

ndgrigorian:
This PR leverages the index-building function `_range` used in `take_along_axis` to implement `dpctl.tensor.put_along_axis`.

### DIFF
--- a/docs/doc_sources/api_reference/dpctl/tensor.indexing_functions.rst
+++ b/docs/doc_sources/api_reference/dpctl/tensor.indexing_functions.rst
@@ -14,5 +14,6 @@ by either integral arrays of indices or boolean mask arrays.
     extract
     place
     put
+    put_along_axis
     take
     take_along_axis
--- a/dpctl/tensor/__init__.py
+++ b/dpctl/tensor/__init__.py
@@ -65,6 +65,7 @@ from dpctl.tensor._indexing_functions import (
     nonzero,
     place,
     put,
+    put_along_axis,
     take,
     take_along_axis,
 )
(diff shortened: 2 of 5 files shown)


oleksandr-pavlyk:
The `append` at line 236 should probably be removed too.

oleksandr-pavlyk:
I am afraid the line "See :func:`dpctl.tensor.put` for an example." may be confusing. The `indices` in `put_along_axis` is a regular array of the same rank as the input, but in the presence of duplicate indices, we may have different number of unique indices, and the array of unique indices may be ragged (more of a list of lists of different lengths). 

I would only leave the note.

ndgrigorian:
> I am afraid the line "See :func:`dpctl.tensor.put` for an example." may be confusing. The `indices` in `put_along_axis` is a regular array of the same rank as the input, but in the presence of duplicate indices, we may have different number of unique indices, and the array of unique indices may be ragged (more of a list of lists of different lengths).
> 
> I would only leave the note.

Good point! I'll remove it. It might be nice to show an example for this case separately.

oleksandr-pavlyk:
The implementation looks good! Tests are needed to avoid the drop in coverage

oleksandr-pavlyk (approved):
I have added tests for `put_along_axis`. Changes now increase coverage. 

Thank you for implementing `put_along_axis` @ndgrigorian ! Looks good to me

oleksandr-pavlyk (approved):
@ndgrigorian I think GH expects  you to review changes I pushed and approve before it would unblock merging without admin override.

ndgrigorian:
@oleksandr-pavlyk 
I just realized that it's not in the docs, I'll push a commit to add `put_along_axis` there, then you will have to approve again and we can merge it. Thanks for adding tests!

oleksandr-pavlyk (approved):

PR #1802, opened 2024-08-19
Explicitly include headers used in the c++ file pertaining to reduction

oleksandr-pavlyk:
This is a good, recommended practice.

Moved global definition of namespaces `td_ns` and `su_ns` in 'reductions.hpp' into namespace.

Fixed resulting build breakage by including needed headers and defining those namespaces locally in affected files.

Changed definition of `logaddexp` in `math_utils.hpp` to use selector instead of if/else branches to improve performance. Switch revealed silent inefficiency from using `std::log1`p for all types (for `sycl::half` it would silently upcast to float and evaluate ``std::log1p<float>(sycl::half)``).

Used `if constexpr` to use sycl::log1p for half and single FP types. For double, use std::log1p until the issue is fixed.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/reductions.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/reductions.hpp
@@ -39,15 +39,15 @@
 #include "utils/type_dispatch_building.hpp"
 #include "utils/type_utils.hpp"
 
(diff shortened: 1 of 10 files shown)


ndgrigorian (approved):
Nightly compiler test failures are unrelated, this LGTM.

Good catch on the namespaces!

PR #1804, opened 2024-08-19
Fix issues for numpy 2.1.0

oleksandr-pavlyk:
Fixed `test_full_gh_1230` to work with NumPy 2.1.0.

Fixed `dpt.from_dlpack` by handling `ValueError` possibly raised by `ndarray.__dlpack__` when its `dl_device` 
keyword argument is unsupported.

---

### DIFF
--- a/dpctl/tests/test_usm_ndarray_ctor.py
+++ b/dpctl/tests/test_usm_ndarray_ctor.py
@@ -1798,7 +1798,10 @@ def test_full_gh_1230(dt):
     dt_maxint = dpt.iinfo(dtype).max
 
     if (dtype.itemsize < 8) and (np.lib.NumpyVersion(np.__version__) < "2.0.0"):
-        X = dpt.full(1, fill_value=(dt_maxint + 1), dtype=dt)
+        try:
+            X = dpt.full(1, fill_value=(dt_maxint + 1), dtype=dt)
+        except OverflowError:
+            pytest.skip("Expected OverflowError raised")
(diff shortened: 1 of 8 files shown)


dpctl/tests/test_usm_ndarray_ctor.py
```diff
@@ -1791,14 +1791,26 @@ def test_full_strides():
...
     X_np = dpt.asnumpy(X)
     assert X.dtype == dpt.dtype(dtype)
-    assert np.array_equal(X_np, np.full_like(X_np, dt_maxint + 1))
+    expected = np.full_like(
+        X_np, _reduce_to_idtype_range(dt_maxint + 1, X.dtype)
+    )
+    assert np.array_equal(X_np, expected)
 
     with pytest.raises(OverflowError):
         dpt.full(1, dpt.iinfo(dpt.uint64).max + 1, sycl_queue=q)
```
ndgrigorian, line 1816: This fix will work, but I would argue this is a temporary fix, we should follow suit with NumPy and disallow Python scalars outside of the dtype range in `full` entirely, which would render this test obsolete.

The issue was originally raised because NumPy allowed this behavior, but we did not.

We can also defer that work to a separate PR.

ndgrigorian:
@oleksandr-pavlyk 
Unfortunately, it seems that changing _to_scalar has exposed some other, subtle overflows in other tests.

ndgrigorian (approved):
Tests are back to passing, LGTM!

PR #1806, opened 2024-08-21
tensor.zeros to use async memset

oleksandr-pavlyk:
Addressed outstanding FIXME note to use async call to populate array with zeros.

Added optimization for `_full_usm_ndarray` to use `handler::memset` instead of `handler::fill` for 1-byte wide types and for other types when fill value is bitwise zero.

### DIFF
--- a/dpctl/tensor/libtensor/source/full_ctor.cpp
+++ b/dpctl/tensor/libtensor/source/full_ctor.cpp
@@ -80,10 +80,65 @@ sycl::event full_contig_impl(sycl::queue &exec_q,
 {
     dstTy fill_v = py::cast<dstTy>(py_value);
 
-    using dpctl::tensor::kernels::constructors::full_contig_impl;
+    sycl::event fill_ev;
 
-    sycl::event fill_ev =
-        full_contig_impl<dstTy>(exec_q, nelems, fill_v, dst_p, depends);
+    if constexpr (sizeof(dstTy) == sizeof(char)) {
+        const auto memset_val = sycl::bit_cast<unsigned char>(fill_v);
+        fill_ev = exec_q.submit([&](sycl::handler &cgh) {
+            cgh.depends_on(depends);
+
+            cgh.memset(reinterpret_cast<void *>(dst_p), memset_val,
+                       nelems * sizeof(dstTy));
+        });
+    }
+    else {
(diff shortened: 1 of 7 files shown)


ndgrigorian:
@oleksandr-pavlyk 
Seems a few array API tests for full regressed with this change, we should see if it's a result of the changes or independent.

dpctl/tensor/libtensor/source/full_ctor.cpp
```diff
@@ -80,10 +80,65 @@ sycl::event full_contig_impl(sycl::queue &exec_q,
...
+            cgh.depends_on(depends);
+
+            cgh.memset(reinterpret_cast<void *>(dst_p), memset_val,
+                       nelems * sizeof(dstTy));
+        });
+    }
+    else {
+        bool is_zero = false;
+        if constexpr (sizeof(dstTy) == 1) {
+            is_zero = (std::uint8_t{0} == sycl::bit_cast<std::uint8_t>(fill_v));
```
ndgrigorian, line 97: This is cool, I hadn't realized some C++20 features were available in the `sycl` namespace.

dpctl/tensor/libtensor/source/zeros_ctor.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+static zeros_contig_fn_ptr_t zeros_contig_dispatch_vector[td_ns::num_types];
+
+std::pair<sycl::event, sycl::event>
+usm_ndarray_zeros(const dpctl::tensor::usm_ndarray &dst,
+                  sycl::queue &exec_q,
+                  const std::vector<sycl::event> &depends)
+{
+    // start, end should be coercible into data type of dst
+
```
ndgrigorian, line 106:
```suggestion
```

ndgrigorian (approved):
Approved, the change looks good to me

dpctl/tensor/libtensor/source/full_ctor.cpp
oleksandr-pavlyk: SYCL spec documents features adopted from future revisions of C++ standard: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:progmodel.futurecppversion

PR #1807, opened 2024-08-22
Do not push_back compute event to host_tasks list

oleksandr-pavlyk:
Do not push_back compute event to host_tasks list.

This is even though sequence of tasks represented by dot_ev does contain host_task to manage temporary USM allocations.

---


ndgrigorian:
- `boolean_indexing.cpp`
- `repeat.cpp`
- `dot.cpp` (changed here)

These are the files where we add compute events to our vectors of host tasks. It may be worth removing them as well.

ndgrigorian (approved):
LGTM!

PR #1808, opened 2024-08-23
Implement dedicated strided `full` kernel and add `order="K"` support to constructors

ndgrigorian:
This PR implements a strided kernel for `full` to enable `order="K"` support in `empty_like`, `full_like`, `ones_like`, and `zeros_like`.

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -1480,7 +1480,20 @@ def full_like(
             _manager.add_event_pair(hev, copy_ev)
             return res
         else:
+            fill_value_type = type(fill_value)
+            dtype = _get_dtype(dtype, sycl_queue, ref_type=fill_value_type)
             res = _empty_like_orderK(x, dtype, usm_type, sycl_queue)
+            if fill_value_type in [float, complex] and np.issubdtype(
+                dtype, np.integer
+            ):
+                fill_value = int(fill_value.real)
+            elif fill_value_type is complex and np.issubdtype(
+                dtype, np.floating
+            ):
+                fill_value = fill_value.real
+            elif fill_value_type is int and np.issubdtype(dtype, np.integer):
+                fill_value = _to_scalar(fill_value, dtype)
+
             _manager = dpctl.utils.SequentialOrderManager[sycl_queue]
             # populating new allocation, no dependent events
             hev, full_ev = ti._full_usm_ndarray(fill_value, res, sycl_queue)
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -904,7 +904,7 @@ def zeros(
             data type of the array. Can be typestring,
             a :class:`numpy.dtype` object, :mod:`numpy` char string,
             or a NumPy scalar type. Default: ``None``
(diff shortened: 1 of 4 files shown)


dpctl/tensor/_ctors.py
```diff
@@ -1392,15 +1452,49 @@ def full_like(
...
+                fill_value,
+                dtype=dtype,
+                order=order,
+                usm_type=usm_type,
+                sycl_queue=sycl_queue,
+            )
+            X = dpt.broadcast_to(X, sh)
+            res = _empty_like_orderK(x, dtype, usm_type, sycl_queue)
+            _manager = dpctl.utils.SequentialOrderManager[sycl_queue]
+            # populating new allocation, no dependent events
```
oleksandr-pavlyk, line 1475: Actually, when `fill_value` has type `dpt.usm_ndarray` we still need to specify dependent events to order this kernel after kernels that potentially populate `fill_value` array.

dpctl/tensor/_ctors.py
```diff
@@ -1244,20 +1273,37 @@ def zeros_like(
...
-        usm_type=usm_type,
-        sycl_queue=sycl_queue,
-    )
+    x_flags = x.flags
+    f_contig = x_flags["F"]
+    c_contig = x_flags["C"]
+    if order == "A":
+        order = "F" if f_contig and not c_contig else "C"
+    if order == "K" and (f_contig or c_contig):
+        order = "C" if c_contig else "F"
```
oleksandr-pavlyk, line 1283: This is a common block of code. It should factored out into `_normalize_order(order, x)` with definition:

```python
def _normalize_order(order, x):
    x_flags = x.flags
    f_contig = x_flags["F"]
    c_contig = x_flags["C"]
    if order == "A":
        order = "F" if f_contig and not c_contig else "C"
    if order == "K" and (f_contig or c_contig):
        order = "C" if c_contig else "F"
    return order
```

and used throughout

ndgrigorian: Good point!

dpctl/tensor/libtensor/source/full_ctor.cpp
```diff
@@ -194,8 +249,42 @@ usm_ndarray_full(const py::object &py_value,
             full_contig_event);
     }
     else {
-        throw std::runtime_error(
-            "Only population of contiguous usm_ndarray objects is supported.");
+        int nd = dst.get_ndim();
+        auto const &dst_shape = dst.get_shape_vector();
+        auto const &dst_strides = dst.get_strides_vector();
+
+        auto fn = full_strided_dispatch_vector[dst_typeid];
```
oleksandr-pavlyk, line 256: At some point it makes sense to implement stride simplification logic here too, perhaps in a future pr.

oleksandr-pavlyk (approved):
Thank you @ndgrigorian ! Looks good to me

dpctl/tensor/libtensor/source/full_ctor.cpp
ndgrigorian: Agreed, I think it would have to use `compact_iteration_space` rather than `simplify_iteration_space_1`.

PR #1810, opened 2024-08-23
tweak to functor implementing gemm 

oleksandr-pavlyk:
This adds  an optimization to save rows/cols of sub-matrices stored in SLM into registers prior to summing over delta_m, delta_n and updating private result with products.

This does not make any difference in performance of the compiled code though due to compiler optimization being effective as loops are unrolled, but makes compiler's job easier and the intent clearer.

 - I verified that compiler output for the kernel with this change is identical to the output before it (checked on Intel Max GPU). 
 - I checked that code compiles for CUDA backend and test suite run is green

----


ndgrigorian (approved):
LGTM!

PR #1814, opened 2024-08-27
Add support for `pow` with inexact base and integer exponent, refactor in-place binary operation type support

ndgrigorian:
This PR adds dedicated kernels to the `pow` dtype matrix for some cases of a floating-point base and integer exponent.

This greatly improves the performance by removing the need to copy and up-cast both arrays.

Also refactors in-place binary operation type support: rather than relying on the out-of-place type support matrix and table, binary in-place operations now have their own dedicated tables. This means that in-place operations can more easily support type combinations which the out-of-place functions don't.


ndgrigorian:
Superseded by #1815 , which contains the same commits refactoring in-place type support tables, but excludes `sycl::pown` overloads due to issues with precision on AMD CPUs revealed in the CI.

PR #1820, opened 2024-09-04
Add `dpctl.tensor._tensor_elementwise_impl._divide_by_scalar` utility function and use it in statistical functions

ndgrigorian:
This PR proposes adding a utility function `_divide_by_scalar` to the `_tensor_elementwise_impl` sub-module.

This function improves the performance of division where the denominator is a Python scalar by avoiding a call to `asarray` and allocating device memory.

This function is also deployed to the statistical functions.

Slips in a change to remove a superfluous `wait()` from `_clip.py`

### DIFF
--- a/dpctl/tensor/_clip.py
+++ b/dpctl/tensor/_clip.py
@@ -206,7 +206,6 @@ def _clip_none(x, val, out, order, _binary_fn):
             )
             _manager.add_event_pair(ht_copy_out_ev, copy_ev)
             out = orig_out
-        ht_binary_ev.wait()
         return out
     else:
         if order == "K":
--- a/dpctl/tensor/_statistical_functions.py
+++ b/dpctl/tensor/_statistical_functions.py
@@ -141,15 +141,8 @@ def _var_impl(x, axis, correction, keepdims):
     res_shape = res.shape
     # when nelems - correction <= 0, yield nans
     div = max(nelems - correction, 0)
-    if div:
-        dep_evs = _manager.submitted_events
-        ht_e7, d_e2 = tei._divide_by_scalar(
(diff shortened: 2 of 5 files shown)


dpctl/tensor/_clip.py
```diff
@@ -206,7 +206,6 @@ def _clip_none(x, val, out, order, _binary_fn):
             )
             _manager.add_event_pair(ht_copy_out_ev, copy_ev)
             out = orig_out
-        ht_binary_ev.wait()
```
oleksandr-pavlyk, line 209: Good catch!

dpctl/tensor/_statistical_functions.py
```diff
@@ -144,17 +141,18 @@ def _var_impl(x, axis, correction, keepdims):
...
-        div_ary = dpt.broadcast_to(div_ary, res.shape)
+    if div:
+        dep_evs = _manager.submitted_events
+        ht_e7, d_e2 = tei._divide_by_scalar(
+            src=res, scalar=div, dst=res, sycl_queue=q, depends=dep_evs
+        )
+        _manager.add_event_pair(ht_e7, d_e2)
+        return res, [d_e2]
+
+    div = dpt.nan
```
oleksandr-pavlyk, line 152:
```suggestion
    if not div:
        div = dpt.nan
```

I made the change originally because `nelems > corrections` branch used divide by scalar, and the else branch used divide in-place. 

Later, both branches were changed to use divide by scalar, but I should have also changed the branching logic.

oleksandr-pavlyk (changes requested):

oleksandr-pavlyk (approved):

PR #1821, opened 2024-09-05
Removed post-link and pre-unlink scripts

oleksandr-pavlyk:
They provided a crutch, which is now handled by
intel-gpu-ocl-icd-system conda package, which a user may install as needed.

Since MKL currently requires OpenCL GPU device to be available, dpnp has run-time dependency on
intel-gpu-ocl-icd-system to make OpenCL GPU device discoverable by OpenCL loader from Python environment.


oleksandr-pavlyk:
The scripts being removed here were added in 7dc575a761ea7444b40d18f35665182e4889b6e5

ndgrigorian (approved):

PR #1828, opened 2024-09-10
Updates black version used in Python code style workflow

ndgrigorian:
Version in [pre-commit config](https://github.com/IntelPython/dpctl/blob/master/.pre-commit-config.yaml) differed from the version in the workflow, which could lead to conflicts between the workflow and pre-commit.

### DIFF
--- a/.github/workflows/python_style_checks.yml
+++ b/.github/workflows/python_style_checks.yml
@@ -43,7 +43,7 @@ jobs:
         with:
           src: "."
           options: "--check"
-          version: "22.12.0"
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk (dismissed):
LGTM! Thanks @ndgrigorian 

I wonder if instead of adding to CHANGELOG directly, we should automate keeping track of changes.

oleksandr-pavlyk (approved):
Thanks! The section "Maintenance" is overdue :)

PR #1829, opened 2024-09-10
Specialize copy_from_numpy_into_usm_ndarray for contig case

oleksandr-pavlyk:
Specialize `_copy_from_numpy_into_usm_ndarray` for contiguous case.

Tensor implementation module temporarily exports both `_copy_numpy_into_usm_ndarray_legacy`, and
`_copy_numpy_into_usm_ndarray` functions to performance comparison.

The specialization reuses the same kernel, instantiated with `NoOpIndexer`, reducing the number of kernel arguments.
It also eliminates small allocation of USM temporaries needed to pass shape/strides information to the general kernel, and the cost of copying from host to these temporary allocations.


dpctl/tensor/libtensor/source/copy_numpy_ndarray_into_usm_ndarray.cpp
```diff
@@ -23,6 +23,7 @@
 //===----------------------------------------------------------------------===//
 
 #include <algorithm>
+#include <exception>
```
ndgrigorian, line 26: This header seems to be currently unnecessary.

ndgrigorian:
@oleksandr-pavlyk 
Please add the changes to the changelog as well, otherwise this LGTM

dpctl/tensor/libtensor/source/copy_numpy_ndarray_into_usm_ndarray.cpp
oleksandr-pavlyk: This is needed, to be able to `throw std::runtime_error`, but I must have removed validation that strides of NumPy array are divisible by itemsize while removed `*_legacy` functions. I am going to reinstate it in a minute. Thank you for catching it!

oleksandr-pavlyk: It would, most likely, compile with that line removed but removal goes against include-what-you-use guideline.

ndgrigorian: But isn't `std::runtime_error` defined in `stdexcept` rather than `exception`?

oleksandr-pavlyk: Yes, https://en.cppreference.com/w/cpp/error/runtime_error. So I am going to push a change to replace `<except>` with `<stdexcept>`

ndgrigorian (approved):
LGTM, this is a nice improvement, thank you @oleksandr-pavlyk

PR #1838, opened 2024-09-16
Throw exception in py_sort if inputs are not C-contiguous

oleksandr-pavlyk:
This changes behavior of `py_sort` C++ function to throw `py::value_error` exception should either input array be not C-contiguous.


ndgrigorian:
While you are in there, [this comment](https://github.com/IntelPython/dpctl/blob/0cdd79e7cfb7b4bf9d6f4053ee9ca83fea1d0dcc/dpctl/tensor/libtensor/source/sorting/sort.cpp#L126) can be removed too, since this is not reduction nor a special case.

oleksandr-pavlyk:
@ndgrigorian Done, and also applied the same change to `argsort.cpp`.

ndgrigorian (approved):
Good change, LGTM!

PR #1847, opened 2024-09-23
Simplify iteration space as static library

oleksandr-pavlyk:
The "simplify_iteration_space.cpp" was being compiled as source for every pybind11 native extension.

This change compiles it as a static library once, and links the library into pybind11 native extension modules, shaving some build time. 

"simplify_iteration_space.cpp" also no longer includes "dpctl4pybind11.hpp", and hence does not need SYCL target enabled, compiling it as ordinary host code.

---


ndgrigorian (approved):
Tested changes locally and it works fine. Extensions which leverage `simplify_iteration_space` also build correctly.

LGTM!

oleksandr-pavlyk:
Compiled for CUDA target. The test suite run was a PASS.

PR #1852, opened 2024-09-30
Technical debt

oleksandr-pavlyk:
This PR 

- adds missing `#include <stdexcept>` to files that use `std::runtime_error` exception
- adds missing `#include <new>` to files that use `std::bad_alloc` exception
- functions that submit kernels which are compiled by means of calling `sycl::get_kernel_bundle<sycl::bundle_state::executable>` now use `cgh.use_kernel_bundle(kb)` to avoid recompiling that kernel.

---

### DIFF
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -27,9 +27,9 @@
 
 #include "dpctl_capi.h"
 #include <complex>
-#include <exception>
 #include <memory>
 #include <pybind11/pybind11.h>
+#include <stdexcept>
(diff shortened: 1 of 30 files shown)


ndgrigorian:
@oleksandr-pavlyk 
Seems `dpctl4pybind11.hpp` includes exception instead of stdexcept, `kernels/elementwise_functions/common.hpp` doesn't include stdexcept either. Those were the only cases I found left (outside of the examples).

oleksandr-pavlyk:
@ndgrigorian Thank you for catching the omissions. I have pushed two commits with fixes.

ndgrigorian (approved):
LGTM!

PR #1859, opened 2024-10-09
Argsort performance improvement

oleksandr-pavlyk:
This change modifies implementation of `tensor.argsort` functions making is about 2x faster.

Instead of implementing argsort as sort over structures (index, value),
with subsequent projection to index, it is now implemented as sort
over linear indices themselves, with dereferencing comparator, and
subsequent mapping from linear index to row-wise index.

On Iris Xe, `tensor.argsort` call took 215 ms to find sorting permutation for
vector of 5670000 elements of type int32, and it now takes 106 ms.

The new implementation no longer makes temporary allocations for
storing indices. Previously, it would allocate 
`2*(sizeof(ValueT) + sizeof(IndexT))`, now it just uses the output 
allocation.

Tests exist.

---


oleksandr-pavlyk:
Drop in coverage must be related to some changes in coverall analysis code. PR did not make changes to files coverall reports have reduced coverage.

ndgrigorian (approved):
This LGTM, the performance improvement is very nice. Thank you @oleksandr-pavlyk !

PR #1866, opened 2024-10-15
Lower the threshold to use sequential sort

oleksandr-pavlyk:
The threshold lowered from 64 elements to sort, to 16 elements. This recovers monotonicity of run-times curve.

When number of array elements along the axis we sort is less than the threshold, a quadratic complexity insertion sort is used. Too high a value of the threshold (64) made for the artifact, where sorting 63 elements took several times longer than sorting 65 elements.

Setting it to 16 recovers monotonicity (sorting 16 elements is not slower than sorting 17 elements).


ndgrigorian (approved):
LGTM!

PR #1867, opened 2024-10-17
Feature/radix sort

oleksandr-pavlyk:
This PR implements radix sort, and exposes it via new `kind` keyword of `dpt.sort` and `dpt.argsort` functions. 
Supported values of `kind` keyword are `"stable"`, `"radixsort"`, and `"mergesort"`, with the default value being `None` (same as `"stable"`). The `"stable"` kind uses radix sort for boolean and short integral types,

---

### DIFF
--- a/dpctl/tensor/_sorting.py
+++ b/dpctl/tensor/_sorting.py
@@ -22,6 +22,11 @@ from ._numpy_helper import normalize_axis_index
 from ._tensor_sorting_impl import (
     _argsort_ascending,
     _argsort_descending,
+    _radix_argsort_ascending,
+    _radix_argsort_descending,
+    _radix_sort_ascending,
+    _radix_sort_descending,
+    _radix_sort_dtype_supported,
     _sort_ascending,
     _sort_descending,
 )
@@ -29,7 +34,15 @@ from ._tensor_sorting_impl import (
 __all__ = ["sort", "argsort"]
 
 
-def sort(x, /, *, axis=-1, descending=False, stable=True):
+def _get_mergesort_impl_fn(descending):
+    return _sort_descending if descending else _sort_ascending
+
+
+def _get_radixsort_impl_fn(descending):
+    return _radix_sort_descending if descending else _radix_sort_ascending
+
+
+def sort(x, /, *, axis=-1, descending=False, stable=True, kind=None):
     """sort(x, axis=-1, descending=False, stable=True)
 
     Returns a sorted copy of an input array `x`.
@@ -49,7 +62,10 @@ def sort(x, /, *, axis=-1, descending=False, stable=True):
             relative order of `x` values which compare as equal. If `False`,
             the returned array may or may not maintain the relative order of
             `x` values which compare as equal. Default: `True`.
-
+        kind (Optional[Literal["stable", "mergesort", "radixsort"]]):
+            Sorting algorithm. The default is `"stable"`, which uses parallel
+            merge-sort or parallel radix-sort algorithms depending on the
+            array data type.
     Returns:
         usm_ndarray:
             a sorted array. The returned array has the same data type and
@@ -74,10 +90,33 @@ def sort(x, /, *, axis=-1, descending=False, stable=True):
             axis,
         ]
         arr = dpt.permute_dims(x, perm)
+    if kind is None:
+        kind = "stable"
+    if not isinstance(kind, str) or kind not in [
+        "stable",
+        "radixsort",
+        "mergesort",
+    ]:
+        raise ValueError(
+            "Unsupported kind value. Expected 'stable', 'mergesort', "
+            f"or 'radixsort', but got '{kind}'"
+        )
+    if kind == "mergesort":
+        impl_fn = _get_mergesort_impl_fn(descending)
+    elif kind == "radixsort":
+        if _radix_sort_dtype_supported(x.dtype.num):
+            impl_fn = _get_radixsort_impl_fn(descending)
+        else:
+            raise ValueError(f"Radix sort is not supported for {x.dtype}")
+    else:
+        dt = x.dtype
+        if dt in [dpt.bool, dpt.uint8, dpt.int8, dpt.int16, dpt.uint16]:
+            impl_fn = _get_radixsort_impl_fn(descending)
+        else:
+            impl_fn = _get_mergesort_impl_fn(descending)
     exec_q = x.sycl_queue
     _manager = du.SequentialOrderManager[exec_q]
     dep_evs = _manager.submitted_events
-    impl_fn = _sort_descending if descending else _sort_ascending
     if arr.flags.c_contiguous:
         res = dpt.empty_like(arr, order="C")
         ht_ev, impl_ev = impl_fn(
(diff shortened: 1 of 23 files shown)


oleksandr-pavlyk:
Build with `DPCTL_TARGET_CUDA=ON` succeeds, test suite run on RTX 3050 passed.

ndgrigorian:
https://github.com/IntelPython/dpctl/blob/3c05c1bdaa74c92171bbcf0cc0750370120fee47/dpctl/tensor/libtensor/include/kernels/sorting/merge_sort.hpp#L809

This line and the initial `sorted_block_size` value seem to be unused.

    static constexpr size_t determine_automatically = 0;
    size_t sorted_block_size =
        (sort_nelems >= 512) ? 512 : determine_automatically;

True or not, in `merge_sort_detail::sort_over_work_group_contig_impl` the `sorted_block_size` is set to the same value with no check against the current value of `sorted_block_size`. So these lines can seemingly be removed entirely.

dpctl/tensor/_sorting.py
```diff
@@ -74,10 +92,33 @@ def sort(x, /, *, axis=-1, descending=False, stable=True):
...
+        kind = "stable"
+    if not isinstance(kind, str) or kind not in [
+        "stable",
+        "radixsort",
+        "mergesort",
+    ]:
+        raise ValueError(
+            "Unsupported kind value. Expected 'stable', 'mergesort', "
+            f"or 'radixsort', but got '{kind}'"
+        )
```
ndgrigorian, line 105: Logic needs to be changed so that if `stable=True`, `"stable"` is always selected regardless of input `kind`.

This is because array API specifies the `stable` keyword.

Same for `argsort`

oleksandr-pavlyk: Presently, we so not implement unstable sorting algorithms. They are all stable. `kind="stable"` is a heuristic algorithm which switches between radix and merge parallel sorting algorithms based on the data type.

ndgrigorian: I see, it's a bit confusing given the `stable` keyword but it seems this is done by NumPy as well, so I don't think any change is necessary.

ndgrigorian:
Nit: renaming of `sort.cpp`/`sort.hpp` and `argsort.cpp`/`argsort.hpp` to `merge_sort` and `merge_argsort` may be good for clarity.

dpctl/tensor/libtensor/source/sorting/py_argsort_common.hpp
```diff
@@ -0,0 +1,152 @@
...
+    bool is_src_c_contig = src.is_c_contiguous();
+    bool is_dst_c_contig = dst.is_c_contiguous();
+
+    if (is_src_c_contig && is_dst_c_contig) {
+        static constexpr py::ssize_t zero_offset = py::ssize_t(0);
+
+        auto fn = sort_contig_fns[src_typeid][dst_typeid];
+
+        if (fn == nullptr) {
+            throw py::value_error("Not implemented for given index type");
```
ndgrigorian, line 133: This should say "given types" now that this can be radix sort based, which doesn't implement for complex.

dpctl/tensor/libtensor/source/sorting/py_sort_common.hpp
```diff
@@ -0,0 +1,149 @@
...
+                              "the same value data type");
+    }
+
+    bool is_src_c_contig = src.is_c_contiguous();
+    bool is_dst_c_contig = dst.is_c_contiguous();
+
+    if (is_src_c_contig && is_dst_c_contig) {
+        constexpr py::ssize_t zero_offset = py::ssize_t(0);
+
+        auto fn = sort_contig_fns[src_typeid];
```
ndgrigorian, line 131: Similarly, this now needs a check for `fn == nullptr`

dpctl/tensor/libtensor/include/kernels/sorting/radix_sort.hpp
```diff
@@ -0,0 +1,1917 @@
...
+    };
+    class use_global_mem_tag
+    {
+    };
+
+public:
+    template <typename ValueT, typename OutputT, typename ProjT>
+    sycl::event operator()(sycl::queue &exec_q,
+                           size_t n_iters,
+                           size_t n_values,
```
ndgrigorian, line 1018: Nit but some comments here that `n_values` are the number of elements to sort would improve readability. Currently, it isn't obvious without reading carefully or looking at the calls from the impl functions.

oleksandr-pavlyk: Renamed.

dpctl/tensor/libtensor/source/sorting/py_sort_common.hpp
oleksandr-pavlyk: Check added.

dpctl/tensor/libtensor/source/sorting/py_argsort_common.hpp
oleksandr-pavlyk: Reworded!

oleksandr-pavlyk:
@ndgrigorian I have pushed changes to address all the issues highlighted in review.

ndgrigorian (approved):
This LGTM now, I've run some benchmarks for different array sizes and shapes and run into no problems.

Thank you @oleksandr-pavlyk !

dpctl/tensor/libtensor/include/kernels/sorting/radix_sort.hpp
```diff
@@ -0,0 +1,1918 @@
...
+                    [ctx, count_ptr]() { sycl_free_noexcept(count_ptr, ctx); });
+            });
+
+            return sort_ev;
+        }
+
+        ValueT *tmp_arr =
+            sycl::malloc_device<ValueT>(n_iters * n_to_sort, exec_q);
+        if (nullptr == tmp_arr) {
+            sycl::free(count_ptr, exec_q);
```
ndgrigorian, line 1630: Actually, I've just realized on re-read: should these `sycl::free` calls be replaced with `sycl_free_noexcept` @oleksandr-pavlyk ?

dpctl/tensor/libtensor/include/kernels/sorting/radix_sort.hpp
```diff
@@ -0,0 +1,1921 @@
...
+
+template <typename ValueT, typename IndexT>
+class populate_indexed_data_for_radix_sort_krn;
+
+template <typename ValueT, typename IndexT>
+class index_write_out_for_radix_sort_krn;
+
+template <typename argTy, typename IndexTy>
+sycl::event
+radix_argsort_axis1_contig_impl(sycl::queue &exec_q,
```
ndgrigorian, line 1758: This impl function seems to be unused anywhere. Should it be kept around or removed as dead code?

oleksandr-pavlyk: Good catch, @ndgrigorian ! I am going to remove it

ndgrigorian (approved):
I think this is ready to go in, LGTM!

oleksandr-pavlyk:
@antonwolfy Thank you for the suggestion. I am going to act on it in a separate PR

dpctl/tensor/_sorting.py
```diff
@@ -74,10 +92,33 @@ def sort(x, /, *, axis=-1, descending=False, stable=True):
...
+    if kind == "mergesort":
+        impl_fn = _get_mergesort_impl_fn(descending)
```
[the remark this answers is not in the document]
oleksandr-pavlyk: @antonwolfy Any Python object can be converted to `bool`, hence casting of `descending` to C++ type `bool` always succeeds, and pybind11 will not encounter issues converting object `descending` to a boolean value.

PR #1872, opened 2024-10-22
Add alternative implementation of device timer to SyclTimer class

oleksandr-pavlyk:
This PR adds Python API to submit empty body single task to a queue. `dpctl.SyclTimer` is modified to acquire `device_timer` keyword argument with supported values being `"queue_barrier"` (legacy behavior, a default), and `"order_manager"`.

With `"order_manager"`, timer submits the empty body single tasks (fence tasks) to the queue, using order manager to order them so as to fence timed submissions. For example, execution of the following snippet:

```python
   prior_ops
   with timer(profiling_queue):
        compute_using_the_queue
   subsequent_ops
```

results in a task graph `[prior_tasks] -> [fence_start_task] -> [ compute_tasks] -> [fence_end_task] -> [subsequent_tasks]`.

Timer uses profiling data from events associated with fence tasks to estimate execution time of compute tasks as measured by the device's timer.

The `device_timer="order_manager"` is useful to time `dpctl.tensor` operations which leverage order manager.

### DIFF
--- a/dpctl/_sycl_timer.py
+++ b/dpctl/_sycl_timer.py
@@ -84,8 +84,7 @@ class OrderManagerDeviceTimer(BaseDeviceTimer):
 
 class SyclTimer:
     """
-    Context to measure device time and host wall-time of execution
-    of commands submitted to :class:`dpctl.SyclQueue`.
+    Context to time execution of tasks submitted to :class:`dpctl.SyclQueue`.
 
     :Example:
         .. code-block:: python
@@ -99,13 +98,18 @@ class SyclTimer:
             milliseconds_sc = 1e3
             timer = dpctl.SyclTimer(time_scale = milliseconds_sc)
 
+            untimed_code_block_1
             # use the timer
             with timer(queue=q):
-                code_block1
+                timed_code_block1
+
+            untimed_code_block_2
 
             # use the timer
             with timer(queue=q):
-                code_block2
+                timed_code_block2
+
+            untimed_code_block_3
 
             # retrieve elapsed times in milliseconds
             wall_dt, device_dt = timer.dt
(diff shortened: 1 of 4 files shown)


dpctl/_sycl_timer.py
```diff
@@ -73,25 +111,36 @@ class SyclTimer:
...
         property. In order to be able to collect the profiling information,
         the ``dt`` property ensures that both submitted barriers complete their
         execution and thus effectively synchronizes the queue.
 
+        `device_timer` keyword argument controls the type of tasks submitted.
+        With `device_timer="queue_barrier"`, queue barrier tasks are used. With
+        `device_timer="order_manager"`, a single empty body task is inserted
+        instead relying on order manager (used by `dpctl.tensor` operations) to
+        order these tasks so that they fence operations performed within
+        timer's context.
```
ndgrigorian, line 127: It may be nice to add some of the details from this PR description, i.e., 
"Tasks will follow an order of `[prior_tasks] -> [fence_start_task] -> [compute_tasks] -> [fence_end_task] -> [subsequent_tasks]` for some `prior_tasks` started before timing, `compute_tasks` which are being timed, `subsequent_tasks` which are performed after timing, and `fence_start_task` and `fence_end_task` used to find the delta time."

I think this is pretty insightful on its own, though it may belong in documentation or examples rather than this docstring.

ndgrigorian: It may also be good to specify the circumstances under which users will want to use `order_manager` vs. `queue_barrier`, and mention the improved accuracy of `order_manager` implementation.

oleksandr-pavlyk: I have pushed changes to the `SyclTimer.__doc__` string.

ndgrigorian (approved):
LGTM!

PR #1877, opened 2024-10-23
Bugfix/gh 1874 result type

oleksandr-pavlyk:
Closes gh-1874

If inputs consist solely of weak dtypes, pre-prepopulate `dtypes` lists with scalar behind the first `weak_dtypes` element. 

Add tests that `dpt.result_types(sc1, sc2)` gives the same result regardless whether the scalar is a Python scalar (weak dtype), or corresponding NumPy scalar (strong dtype). 


ndgrigorian (dismissed):
This LGTM, and array API is considering specifyingthat only scalar types are acceptable in `result_type`, so this is a subtle change toward supporting the next version of the spec.

vtavana (approved):
Thank you, @oleksandr-pavlyk for fixing it.

ndgrigorian (approved):
Windows tests now passing, LGTM!

PR #1878, opened 2024-10-23
Improve `dpctl.tensor.full` error for invalid `fill_value`

ndgrigorian:
This PR proposes adding a check to `full` when the value to construct the array with is not a scalar value, i.e., some type of object or a string.

Previously these cases would fail, but pybind11 would throw RuntimeError. Now, a TypeError is thrown from Python, and the result array is never allocated, preventing wasted memory allocation.

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -1038,6 +1038,19 @@ def _cast_fill_val(fill_val, dt):
         return fill_val
 
 
+def _validate_fill_value(fill_val):
+    """
+    Validates that `fill_val` is a numeric or boolean scalar.
+    """
+    # TODO: verify if `np.True_` and `np.False_` should be instances of
+    # Number in NumPy, like other NumPy scalars and like Python bools
+    # check for `np.bool_` separately as NumPy<2 has no `np.bool`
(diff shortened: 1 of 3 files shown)


oleksandr-pavlyk (dismissed):
LGTM!

ndgrigorian:
I've added this change to the changelog as well.

oleksandr-pavlyk (dismissed):

dpctl/tensor/_ctors.py
```diff
@@ -1110,6 +1111,16 @@ def full(
...
+    # TODO: verify if `np.True_` and `np.False_` should be instances of
+    # Number in NumPy, like other NumPy scalars and like Python bools
+    # check for `np.bool_` separately as NumPy<2 has no `np.bool`
+    elif not isinstance(fill_value, Number) and not isinstance(
+        fill_value, np.bool_
+    ):
+        raise TypeError(
+            "`full` array cannot be constructed with value of type "
+            f"{type(fill_value)}"
+        )
```
oleksandr-pavlyk, line 1123: It would be nice to change this to 

```
else:
     _validate_fill_value(fill_value)
```

and result `_validate_fill_value` function in both `full` and `full_like` implementation.

PR #1879, opened 2024-10-25
Subgroup load store cleanup

oleksandr-pavlyk:
This PR affects implementation of all element-wise functions for contiguous inputs. It replaces `(vec_sz, n_vecs)` template parameters used from the same value `(4, 2)` to `(VecSize_v<T>, 1)` which is type dependent, chosen to ensure that for each type work-items in the same sub-group load the entire cache line for each vector argument. 

The number of iterations used by the kernel is always set to 1 to make memory access pattern in the work-group more predictable.

The work-groups size has been increased from 128 to 256 (8 threads with simd32 used) so that work-group occupies all threads of a single vector core. 

The performance of element-wise functions, i.e. `add`, `multiply` has increased up to 30% for very large inputs (7300 million elements). No performance degradation was observed. 

For example, testing addition of `330241024` element arrays, the kernel execution for type double went down from 13.5 ms to 10.5ms:

this PR:
```
"dpctl::tensor::kernels::add::add_contig_kernel<double, double, double, 1u, 1u>[SIMD32 {1290004; 1; 1} {256; 1; 1}]",          201,           2117591360,     5.300246,             10535280,              9817920,             14923680 
```

main branch:
```
"dpctl::tensor::kernels::add::add_contig_kernel<double, double, double, 4u, 2u>[SIMD32 {322501; 1; 1} {128; 1; 1}]",          201,           2712198720,     5.414135,             13493525,             10416640,             15942560
```

---


oleksandr-pavlyk:
Closing this PR

PR #1883, opened 2024-10-29
Change descending from template parameter to an argument

oleksandr-pavlyk:
This PR removes template parameter `is_ascending` from radix sort functions and introduces it as function call argument instead. 

This halves the binary size due to reduction in device code portion. Furthermore, radix sorting functions are singled out into a dedicated Python module `_tensor_sorting_radix_impl` further reducing the build time and memory footprint and provide increased parallelization opportunities to the linker. 

The overall build time drops from 45 minutes to 33 minutes on my Core i7 laptop.


oleksandr-pavlyk:
On GPU Max 1100, sorting an array of `int16` data-type of size 10241024 elements  evaluates at about the same 2.10 ms with this change as it does before it.

ndgrigorian (approved):
LGTM, Windows build failure is unrelated and examples are failing in other PRs.

PR #1889, opened 2024-11-12
Elementwise functions tuning

oleksandr-pavlyk:
This PR revisits elementwise functions functors for contiguous inputs.

1. Since work-group size is chose so that it is always a multiple of any permissible sub-group size, there is no point in using more expensive `sg.get_local_range()`, so it is replaced with cheaper `sg.get_max_local_range()`. This change also slightly reduces the binary size due to leaner kernel (from `36428264` bytes down to `36345112` bytes).

2. Implementations of each elementwise function for contiguous input can now set hyperparameters `vec_sz` and `n_vecs` differently for different input types. This ability is applied to `add_contig_impl` for some modest performance improvement for `int32_t`, `uint32_t`, `int64_t`, `uint64_t`, `float` and `double`.

3. Fixed missing check in implementation of `minimum` and `maximum` for `sycl::half` type for vector inputs, which caused test failures on AMD CPUs in CI during earlier iterations of this work (gh-1879).

4. Added missing `include <type_traits>` in type dispatching headers, and simplified code.

---


dpctl/tensor/libtensor/include/kernels/elementwise_functions/vec_size_util.hpp
```diff
@@ -0,0 +1,74 @@
+//=== tanh.hpp -   Unary function TANH                     ------
```
ndgrigorian, line 1:
```suggestion
//=== vec_size_util.hpp -   Utilities for element-wise function hyperparameters                     ------
```

oleksandr-pavlyk: I push a fix

oleksandr-pavlyk:
I ran performance micro benchmark on elementwise binary functions `add` and `multiply` on Meteor Lake iGPU and on RTX 3050 for Laptops GPU. There was not performance regression noted for changes in this branch versus the development branch.

ndgrigorian (dismissed):
LGTM, this brings a welcome performance improvement!

ndgrigorian (approved):

oleksandr-pavlyk:
The coverall got the results of coverage run, per its own records:

image

which corresponds to the latest commit 88c3e1a5455190f6a685b00f7de6a3023be9c806 on this branch.

PR #1891, opened 2024-11-13
Contribution to fix gh 1887

oleksandr-pavlyk:
This PR contributes to gh-1890, by streamlining code for performing copy to contiguous array.

1. Save common subexpressions to variables
2. Sub-group size type changed to uint16 (from uint32)
3. sg.get_local_range() replaced with sg.get_max_local_range()

   This is safe to do since work-group size is chosen to be
   a multiple of sub-group size for all possile choices
   of sub-group size (1, 8, 16, 32, 64)

4. Simplified computation of base value in generic branch for
    complex types, or when sg_load is disabled, to avoid a
    division (and left a comment)

---

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/copy_as_contiguous.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/copy_as_contiguous.hpp
@@ -132,8 +132,8 @@ public:
 
 template <typename T,
           typename IndexerT,
-          std::uint32_t n_vecs,
           std::uint32_t vec_sz,
+          std::uint32_t n_vecs,
           bool enable_sg_load,
           typename KernelName>
 sycl::event submit_c_contiguous_copy(sycl::queue &exec_q,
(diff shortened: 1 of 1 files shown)


dpctl/tensor/libtensor/include/kernels/copy_as_contiguous.hpp
```diff
@@ -150,9 +160,11 @@ sycl::event submit_c_contiguous_copy(sycl::queue &exec_q,
     const std::size_t lws =
         ((preferred_lws + max_sg_size - 1) / max_sg_size) * max_sg_size;
 
-    constexpr std::uint32_t nelems_per_wi = n_vecs * vec_sz;
-    size_t n_groups =
-        (nelems + nelems_per_wi * lws - 1) / (nelems_per_wi * lws);
+    constexpr std::uint8_t nelems_per_wi = n_vecs * vec_sz;
```
ndgrigorian, line 163: If we're going to use `uint8_t` here, maybe a note should be made at the top of the functor or in the impl function that we assume `n_vecs * vec_sz` fits in the type.

oleksandr-pavlyk: I added static assert to verify that.

ndgrigorian (approved):
LGTM!

PR #1894, opened 2024-11-15
Fix undefined behavior in integer advanced indexing and indexing functions

ndgrigorian:
This PR proposes a solution to undefined behavior that could occur in some edge cases with integer advanced indexing, where indices OOB for `ssize_t` (aka `std::ptrdiff_t`) would be cast directly to `ssize_t` and overflow or underflow.

As `ssize_t`/`std::ptrdiff_t` is defined to be a signed type with the same size as `size_t`, this means that on 32-bit systems, overflow/underflow could occur for even smaller values.

This PR also re-organizes `integer_advanced_indexing.hpp` by reducing namespace clutter, and moves the rewritten `ClipIndex` and `WrapIndex` structs into a separate header file. This enables them to be re-used more easily in extensions.

### DIFF
--- a/dpctl/tensor/libtensor/include/utils/indexing_utils.hpp
+++ b/dpctl/tensor/libtensor/include/utils/indexing_utils.hpp
@@ -1,0 +1,140 @@
+//===----- indexing_utils.hpp - Utilities for indexing modes  -----*-C++-*/===//
+//
+//                      Data Parallel Control (dpctl)
+//
+// Copyright 2020-2024 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


dpctl/tensor/libtensor/include/utils/indexing_utils.hpp
```diff
@@ -0,0 +1,140 @@
...
+ * uses py::ssize_t for shapes and strides internally and Python uses
+ * py_ssize_t for sizes of e.g. lists.
+ */
+
+template <typename IndT> struct WrapIndex
+{
+    ssize_t operator()(ssize_t max_item, IndT ind) const
+    {
+        ssize_t projected;
+        max_item = sycl::max<ssize_t>(max_item, 1);
```
oleksandr-pavlyk, line 52:
```suggestion
        max_item = std::max<ssize_t>(max_item, 1);
```

PR #1895, opened 2024-11-15
Contribution to 1894

oleksandr-pavlyk:
This PR fixes build break in gh-1894.

In `indexing_utils.hpp` I moved common constexpr variables out of branches .

Replaced `static constexpr` with `constexpr`. Since these are
defined in procedure scope, `static` is not required.

Introduced typed temporary variables, so that type deduction
for `sycl::min`, `sycl::max`, `sycl::clamp` can work and removed
explicit use of their template parameter.

Added explicit static_cast on value of `projected` variable
computed as IndT type.

In `integer_advanced_indexing.hpp`, I made `ProjectT proj` constexpr.

---


ndgrigorian (approved):
Builds passing, LGTM!

PR #1896, opened 2024-11-15
[MAINT] Save repeated expression in accumulation functor operator to a variable

oleksandr-pavlyk:
Checking in small change in accumulator header file. 

It saved repeated sub-expressions into a temporary to clean up the code and hopefully reduce the size of the binary by a few bytes.


ndgrigorian (approved):

oleksandr-pavlyk:
Note that the change in this PR introduces gh-1901, and gh-1902 fixes the issue introduced here.

PR #1898, opened 2024-11-18
Address CI warnings to not use nodefaults channel

oleksandr-pavlyk:
Remove use of nodefaults channel, use conda-remove-defaults: true instead

This change is prompted by CI warning:

```
'nodefaults' channel detected: will remove 'defaults' if added implicitly.
In the future, 'nodefaults' as a way of removing 'defaults' won't be
supported. Please set 'conda-remove-defaults' = 'true' to remove this warning.
```

image

---

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -91,7 +91,8 @@ jobs:
           miniforge-variant: Miniforge3
           miniforge-version: latest
           activate-environment: build
-          channels: conda-forge,nodefaults
+          channels: conda-forge
+          conda-remove-defaults: true
           python-version: ${{ matrix.python }}
 
       - name: Install conda build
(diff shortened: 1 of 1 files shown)


ndgrigorian (dismissed):
Warnings are gone, LGTM

oleksandr-pavlyk:
Deprecation warnings are gone. Future warnings are still present, but the log does not show `defaults` channel being active even though it may be temporarily added to the list of channels by conda.

.github/workflows/conda-package.yml
```diff
@@ -800,7 +802,8 @@ jobs:
         with:
           run-post: false
           channel-priority: "disabled"
-          channels: conda-forge,nodefaults
+          channels: conda-forge
```
antonwolfy, line 805: `upload_windows` job uses `conda-incubator/setup-miniconda` step, but there is `channels` property missing. I think it needs to be aligned with other usages of `conda-incubator/setup-miniconda` step and so both `channels: conda-forge` and `conda-remove-defaults: true` properties to be defined there.

antonwolfy: Otherwise the base environment is creating with packages from `pkgs/main` channel which is restricted.

antonwolfy (approved):
Thank you @oleksandr-pavlyk, LGTM!

ndgrigorian:
Unfortunately, the deprecated warnings appear to be back, for the Windows step

ndgrigorian (approved):
I misread in my last comment, it was the future warnings. LGTM!

Issue #1913, opened 2024-12-02
dpctl raises exception with empty boolean array passed to `__setitem__()`


antonwolfy:
It seems dpctl doesn't handle a corner case properly when an empty boolean array passed through advanced indexing in `__setitem__()`:
```python
import numpy, dpctl, dpctl.tensor as dpt

dpctl.__version__
# Out: '0.19.0dev0+280.g4a94ef9c25'

numpy.__version__
# Out: '2.1.3'

a = dpt.ones((2, 3, 4))
a[dpt.asarray([], dtype=numpy.bool_)] = 1
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[24], line 1
----> 1 a[dpt.asarray([], dtype=numpy.bool_)] = 1

File dpctl/tensor/_usmarray.pyx:1445, in dpctl.tensor._usmarray.usm_ndarray.__setitem__()

File /localdisk/work/antonvol/soft/miniforge3/envs/numpy_20/lib/python3.12/site-packages/dpctl/tensor/_copy_utils.py:928, in _place_impl(ary, ary_mask, vals, axis)
    926 rhs = dpt.broadcast_to(rhs, expected_vals_shape)
    927 dep_ev = _manager.submitted_events
--> 928 hev, pl_ev = ti._place(
    929     dst=ary,
    930     cumsum=cumsum,
    931     axis_start=pp,
    932     axis_end=pp + mask_nd,
    933     rhs=rhs,
    934     sycl_queue=exec_q,
    935     depends=dep_ev,
    936 )
    937 _manager.add_event_pair(hev, pl_ev)
    938 return

ValueError: Inconsistent array dimensions

# it works in NumPy
na[numpy.asarray([], dtype=numpy.bool_)] = 1

# and it works with integer array in dpctl:
a[dpt.asarray([], dtype=numpy.intp)] = 1
```

PR #1915, opened 2024-12-02
Fix 1913 empty boolean mask

oleksandr-pavlyk:
Closes gh-1913

Add special-casing for empty mask in the implementation of `tensor.place`, which is used by `usm_ndarray.__setitem__` as well.

Added tests.

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -924,6 +924,8 @@ def _place_impl(ary, ary_mask, vals, axis=0):
     else:
         rhs = dpt.astype(vals, ary.dtype)
     rhs = dpt.broadcast_to(rhs, expected_vals_shape)
+    if mask_nelems == 0:
+        return
     dep_ev = _manager.submitted_events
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_copy_utils.py
```diff
@@ -924,6 +924,8 @@ def _place_impl(ary, ary_mask, vals, axis=0):
     else:
         rhs = dpt.astype(vals, ary.dtype)
     rhs = dpt.broadcast_to(rhs, expected_vals_shape)
+    if mask_nelems == 0:
+        return
```
ndgrigorian, line 928: Is it possible to write this check earlier (i.e., line 910) as a very small optimization?

oleksandr-pavlyk: I was considering that, but that the check that value array can be broadcast to the desired shape would be skipped. I think error checking should not be different between generic case and this special case.

ndgrigorian (approved):
LGTM!

PR #1918, opened 2024-12-03
Technical debt changes in radix_sort.hpp

oleksandr-pavlyk:
1. Add missing `#include <array>` needed since `std::array` is used several times in the code.
2. Made `peer_helper` class specializations members const, as well as marking peer_contribution method const.

   This allows to make `peer_prefix_hlp` class instance const as well.

3. Added a comment to why subgroup_ballot algorithm for peer helper is only applicable for sub-groups with sizes 32 and narrower.

---

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/sorting/radix_sort.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/sorting/radix_sort.hpp
@@ -27,6 +27,7 @@
 
 #pragma once
 
+#include <array>
 #include <cstdint>
 #include <limits>
 #include <stdexcept>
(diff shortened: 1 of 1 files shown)


ndgrigorian:
The `atomic_fetch_or` peer_prefix_algo seems to be entirely unused in the code.

Is it dead code, or should it be kept around?

oleksandr-pavlyk:
> The `atomic_fetch_or` peer_prefix_algo seems to be entirely unused in the code.
> 
> Is it dead code, or should it be kept around?

At present, it is indeed unused. It is useful to keep around for debugging purposes. I did check, locally, that the code compiles and works correctly if `atomic_fetch_or` is used instead of `subgroup_ballot` algorithm.

The other useful feature of `atomic_fetch_or` algorithm is that work-group size in reorder phase kernels need not be constrained to the size of the sub-group, although `radix_sort_reorder_submit` always sets it to be.

ndgrigorian (approved):
In that case LGTM!

Issue #1745, opened 2024-07-19
dpctl pip install broken on windows for python3.11


icfaust:
System info:
```
OS Name:                   Microsoft Windows 11 Enterprise
OS Version:                10.0.22631 N/A Build 22631
System Type:               x64-based PC
Processor(s):              1 Processor(s) Installed.
                           [01]: Intel64 Family 6 Model 140 Stepping 1 GenuineIntel ~2995 Mhz
```
Using python3.11

Reproducer
```
python -m venv env2
env2\Scripts\activate.bat
python -m pip install dpctl
python -c "import dpctl"
```
![image](https://github.com/user-attachments/assets/3c6d8827-8a66-454a-9e90-e9550866e1cb)

Not sure if I am doing something wrong, but I am seeing this failure locally and in github action VMs using pip

icfaust:
temporary fix:
```
import os, site
path_to_env = site.getsitepackages()[0]
path_to_libs = os.path.join(path_to_env, "Library", "bin")
os.add_dll_directory(path_to_libs)
import dpctl
```

icfaust:
here is a rough github workflow which can show the error. also when fixed it doesnt show any devices (second error?)
```
name: oneDAL DPCPP nightly sklearnex
on:
  workflow_dispatch:

env:
  PYTHON_VERSION: 3.9
  DPCTL_VERSION: 0.17.0
  DPNP_VERSION: 0.15.0

jobs:
dpctl:
    name: dpctl_test
    runs-on: windows-2022
    steps:
      - name: Checkout
        uses: actions/checkout@v4
      - name: Install Python
        uses: actions/setup-python@v5
        with:
          python-version: ${{ env.PYTHON_VERSION }}
      - name: Install dpctl
        shell: cmd
        run: |
          set PATH=C:\msys64\usr\bin;%PATH%
          python -m venv env2
          call .\env2\Scripts\activate.bat
          set DPCTL_SUPPORTED=3.9 3.11
          FOR %%i IN (%DPCTL_SUPPORTED%) DO if ${{ env.PYTHON_VERSION }}==%%i pip install dpctl==${{ env.DPCTL_VERSION }} dpnp==${{ env.DPNP_VERSION }}
[...]

PR #1922, opened 2024-12-04
Support dpctl in vritual environment out of the box

oleksandr-pavlyk:
Since location of "Library\bin" in the virtual environment is not on the default search path, importing of `dpctl` fails due to unmet dependencies for native extensions of `dpctl` submodules.

This change introduces `"_init_helper.py"` file which implements the following logic using built-in `os` Python module:

1. If `os.add_dll_directory` attribute exists, and `VIRTUAL_ENV` environment variable is set, and ``os.path.join(os.environ["VIRTUAL_ENV"], "Library", "bin")`` folder exists, call ``os.add_dll_directory`` with that directory.

With this change the gh-1745 is fixed, and ``python -m dpctl -f`` works out of the box.

Only GPU devices are visible, and to enable CPU device two additional steps must be performed:

  1. Edit ``%VIRUAL_ENV%\Library\bin\cl.cfg`` and set `CL_CONFIG_TBB_DLL_PATH` variable at the bottom of the configuration file to the **expanded value** of ``%VIRUAL_ENV%\Library\bin\tbb12.dll`` but use **forward** slashes, instead of native **backward** slashes.
  2. Append ``%VIRUAL_ENV%\Library\bin`` to the ``PATH`` using ``set "PATH=%PATH%:%VIRTUAL_ENV%\Library\bin"``

After these changes `python -m dpctl -f` should see CPU device.

@icfaust

---

### DIFF
--- a/dpctl/_init_helper.py
+++ b/dpctl/_init_helper.py
@@ -1,0 +1,27 @@
+#                      Data Parallel Control (dpctl)
+#
+# Copyright 2020-2024 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import os.path
+
+if hasattr(os, "add_dll_directory"):
+    # For virtual environments on Windows, add folder
+    # with DPC++ libraries to the DLL search path gh-1745
+    if "VIRTUAL_ENV" in os.environ:
+        venv_dir = os.environ["VIRTUAL_ENV"]
+        expected_dir = os.path.join(venv_dir, "Library", "bin")
+        if os.exists(expected_dir):
+            os.add_dll_directory(expected_dir)
--- a/dpctl/__init__.py
+++ b/dpctl/__init__.py
@@ -26,6 +26,7 @@ __author__ = "Intel Corp."
 import os
 import os.path
 
+from . import _init_helper
 from ._device_selection import select_device_with_aspects
 from ._sycl_context import SyclContext, SyclContextCreationError
 from ._sycl_device import (
@@ -137,3 +138,4 @@ def get_include():
 
 __version__ = get_versions()["version"]
 del get_versions
+del _init_helper


dpctl/_init_helper.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+import os.path
+
+if hasattr(os, "add_dll_directory"):
+    # For virtual environments on Windows, add folder
+    # with DPC++ libraries to the DLL search path gh-1745
+    if "VIRTUAL_ENV" in os.environ:
+        venv_dir = os.environ["VIRTUAL_ENV"]
+        expected_dir = os.path.join(venv_dir, "Library", "bin")
+        if os.exists(expected_dir):
```
ndgrigorian, line 26:
```suggestion
        if os.path.exists(expected_dir):
```

icfaust:
@ethanglaser PR fix for the issue we talked about.

icfaust:
You should be able to steer and modify the ```cl.cfg``` if you use ```os.sep``` to generalize your pathing instead of hardcoding ```/```

ndgrigorian (approved):
Installed wheel on Windows machine and it worked out of the box, so LGTM!

oleksandr-pavlyk:
@icfaust By the way, it turns out that augmenting `PATH` variable in item 2 rendered changes in item 1 superfluous. The CPU device is discoverable even without modifying `"cl.cfg"` file.

dpctl/_init_helper.py
```diff
@@ -22,6 +22,8 @@
...
     if "VIRTUAL_ENV" in os.environ:
         venv_dir = os.environ["VIRTUAL_ENV"]
-        expected_dir = os.path.join(venv_dir, "Library", "bin")
-        if os.path.exists(expected_dir):
-            os.add_dll_directory(expected_dir)
+        # Expect to find file per PEP-0405
+        expected_file = os.path.join(venv_dir, "pyvenv.cfg")
+        dll_dir = os.path.join(venv_dir, "Library", "bin")
+        if os.path.isdir(dll_dir) and os.path.isfile(expected_file):
+            os.add_dll_directory(dll_dir)
```
ndgrigorian, line 29: If I'm understanding PEP-0405 correctly (and according to some [StackOveflow discussion](https://stackoverflow.com/questions/1871549/how-to-determine-if-python-is-running-inside-a-virtualenv)) another approach could be (per Python docs) use of [base_exec_prefix](https://docs.python.org/3/library/sys.html#sys.base_exec_prefix) and [exec_prefix](https://docs.python.org/3/library/sys.html#sys.exec_prefix), for example:
```
import sys
...
if sys.base_exec_prefix != sys.exec_prefix:
        dll_dir = os.path.join(sys.base_exec_prefix, "Library", "bin")
```

ndgrigorian:
@oleksandr-pavlyk 
In addition, with `sys` being imported, we can also do `if sys.platform == "win32":` instead of checking for `add_dll_directory` attribute, which seems a little clearer.

dpctl/_init_helper.py
oleksandr-pavlyk: Great idea. That way we do not depend on the environment variable, closing possible security hole.

oleksandr-pavlyk:
Thank you for the suggestion @ndgrigorian .  

I have pushed the change to implement it, and the fix no longer relies on environment variables, alleviating security concerns.

ndgrigorian (approved):
LGTM, tested in virtualenv on Windows and it worked.

Thanks @oleksandr-pavlyk !

PR #1932, opened 2024-12-12
Speed up custom reductions

oleksandr-pavlyk:
The function used to perform custom reduction in a single
work-item (leader of the work-group sequentially).

It now does so cooperatively few iterations, and
processes remaining non-reduced elements sequentially
in the leading work-item.

The custom_reduce_over_group got sped up about a factor of 3x.

The following now shows timing of the reduction kernel

```
unitrace -d -v -i 20 python -c "import dpctl.tensor as dpt; dpt.min(dpt.ones(10**7, dtype=dpt.float32)).sycl_queue.wait()"
```

or par (less that 10%) slower than the int32 kernel, which uses
built-in sycl::reduce_over_group:

```
unitrace -d -v -i 20 python -c "import dpctl.tensor as dpt; dpt.min(dpt.ones(10**7, dtype=dpt.int32)).sycl_queue.wait()"
```

Also fixed uses of group_load/group_store and enabled them for 2025.0 compiler. SYCLOS warnings should be gone now.

### DIFF
--- a/dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
+++ b/dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
@@ -132,27 +132,81 @@ size_t choose_workgroup_size(const size_t nelems,
     return wg;
 }
 
+namespace
+{
+
+template <typename LocAccT, typename OpT>
+void _fold(LocAccT &local_mem_acc,
+           const std::uint32_t lid,
+           const std::uint32_t cutoff,
+           const std::uint32_t step,
+           const OpT &op)
+{
+    if (lid < cutoff) {
+        local_mem_acc[lid] = op(local_mem_acc[lid], local_mem_acc[step + lid]);
+    }
+}
+
+template <typename LocAccT, typename OpT>
+void _fold(LocAccT &local_mem_acc,
+           const std::uint32_t lid,
+           const std::uint32_t step,
+           const OpT &op)
+{
+    if (lid < step) {
+        local_mem_acc[lid] = op(local_mem_acc[lid], local_mem_acc[step + lid]);
+    }
+}
+
+} // namespace
+
 template <typename T, typename GroupT, typename LocAccT, typename OpT>
 T custom_reduce_over_group(const GroupT &wg,
                            LocAccT local_mem_acc,
                            const T &local_val,
                            const OpT &op)
 {
-    size_t wgs = wg.get_local_linear_range();
-    local_mem_acc[wg.get_local_linear_id()] = local_val;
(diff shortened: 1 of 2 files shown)


ndgrigorian (approved):
This brings a great performance improvement for reductions not using `sycl::reduce_over_group`.

This LGTM, thank you @oleksandr-pavlyk !

dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
```diff
@@ -132,27 +132,81 @@ size_t choose_workgroup_size(const size_t nelems,
...
 template <typename T, typename GroupT, typename LocAccT, typename OpT>
 T custom_reduce_over_group(const GroupT &wg,
                            LocAccT local_mem_acc,
                            const T &local_val,
                            const OpT &op)
 {
-    size_t wgs = wg.get_local_linear_range();
-    local_mem_acc[wg.get_local_linear_id()] = local_val;
+    constexpr std::uint32_t low_sz = 8u;
+    constexpr std::uint32_t high_sz = 1024u;
```
oleksandr-pavlyk, line 170: These constants were determined experimentally, when testing on Iris Xe, Core Ultra 7 Arc 140V, and PVC.

dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
```diff
@@ -132,27 +132,81 @@ size_t choose_workgroup_size(const size_t nelems,
...
+                n_witems = n_witems_;
+            }
+        }
+    }
+    else {
+        // wgs is a power of 2
+#pragma unroll
+        for (std::uint32_t sz = high_sz; sz >= low_sz; sz >>= 1) {
+            if (n_witems >= sz) {
+                n_witems = (n_witems + 1) >> 1;
```
oleksandr-pavlyk, line 195: Since `n_witems` is known to be a power of 2 at every iteration, this can be simplified to `n_witems >>= 1`.

dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
oleksandr-pavlyk: This comment is left for posterity. No action is needed.

ndgrigorian: A comment could be added which mentions this specifically

dpctl/tensor/libtensor/include/utils/sycl_utils.hpp
oleksandr-pavlyk: I will make this change in a follow-up PR.

PR #1946, opened 2024-12-20
Add iota kernel

oleksandr-pavlyk:
This PR builds on top of feature/topk branch. 

It adds `iota_impl` in new `sort_utils.hpp` file, and uses it in `merge_sort.hpp`, `radix_sort.hpp` and `topk.hpp`.

It also fixes possible USM allocation leak in exception handling.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
@@ -34,9 +34,10 @@
 #include <vector>
 
 #include "kernels/dpctl_tensor_types.hpp"
-#include "merge_sort.hpp"
-#include "radix_sort.hpp"
-#include "search_sorted_detail.hpp"
+#include "kernels/sorting/merge_sort.hpp"
+#include "kernels/sorting/radix_sort.hpp"
+#include "kernels/sorting/search_sorted_detail.hpp"
+#include "kernels/sorting/sort_utils.hpp"
 #include "utils/sycl_alloc_utils.hpp"
 #include <sycl/ext/oneapi/sub_group_mask.hpp>
 
@@ -95,20 +96,26 @@ topk_full_merge_sort_impl(sycl::queue &exec_q,
         throw std::runtime_error("Unable to allocate device_memory");
     }
 
+    using IotaKernelName = topk_populate_index_data_krn<argTy, IndexTy, CompT>;
+
+#if 1
+    using dpctl::tensor::kernels::sort_utils_detail::iota_impl;
+
+    sycl::event populate_indexed_data_ev = iota_impl<IotaKernelName, IndexTy>(
+        exec_q, index_data, iter_nelems * axis_nelems, depends);
+#else
     sycl::event populate_indexed_data_ev =
         exec_q.submit([&](sycl::handler &cgh) {
             cgh.depends_on(depends);
 
             auto const &range = sycl::range<1>(iter_nelems * axis_nelems);
 
-            using KernelName =
-                topk_populate_index_data_krn<argTy, IndexTy, CompT>;
-
-            cgh.parallel_for<KernelName>(range, [=](sycl::id<1> id) {
+            cgh.parallel_for<IotaKernelName>(range, [=](sycl::id<1> id) {
                 std::size_t i = id[0];
                 index_data[i] = static_cast<IndexTy>(i);
             });
         });
+#endif
 
     std::size_t sorted_block_size;
     // Sort segments of the array
@@ -480,18 +487,25 @@ sycl::event topk_radix_impl(sycl::queue &exec_q,
         radix_sort_details::IndexedProj<IndexTy, argTy, IdentityProjT>;
     const IndexedProjT proj_op{arg_tp, IdentityProjT{}};
 
+    using IotaKernelName = topk_iota_krn<argTy, IndexTy>;
+
+#if 1
+    using dpctl::tensor::kernels::sort_utils_detail::iota_impl;
+
+    sycl::event iota_ev = iota_impl<IotaKernelName, IndexTy>(
+        exec_q, workspace, total_nelems, depends);
+#else
     sycl::event iota_ev = exec_q.submit([&](sycl::handler &cgh) {
         cgh.depends_on(depends);
 
-        using KernelName = topk_iota_krn<argTy, IndexTy>;
-
-        cgh.parallel_for<KernelName>(
+        cgh.parallel_for<IotaKernelName>(
             sycl::range<1>(total_nelems), [=](sycl::id<1> id) {
                 size_t i = id[0];
                 IndexTy sort_id = static_cast<IndexTy>(i);
                 workspace[i] = sort_id;
             });
     });
+#endif
 
     sycl::event radix_sort_ev =
         radix_sort_details::parallel_radix_sort_impl<IndexTy, IndexedProjT>(
--- a/dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
@@ -30,8 +30,9 @@
 #include <iterator>
 #include <limits>
 #include <stdexcept>
-#include <sycl/sycl.hpp>
 #include <vector>
+
+#include <sycl/sycl.hpp>
 
 #include "kernels/dpctl_tensor_types.hpp"
 #include "kernels/sorting/merge_sort.hpp"
@@ -90,11 +91,11 @@ topk_full_merge_sort_impl(sycl::queue &exec_q,
                           const CompT &comp,
                           const std::vector<sycl::event> &depends)
 {
-    IndexTy *index_data =
-        sycl::malloc_device<IndexTy>(iter_nelems * axis_nelems, exec_q);
-    if (index_data == nullptr) {
-        throw std::runtime_error("Unable to allocate device_memory");
-    }
+    auto index_data_owner =
+        dpctl::tensor::alloc_utils::smart_malloc_device<IndexTy>(
+            iter_nelems * axis_nelems, exec_q);
+    // extract USM pointer
+    IndexTy *index_data = index_data_owner.get();
 
     using IotaKernelName = topk_populate_index_data_krn<argTy, IndexTy, CompT>;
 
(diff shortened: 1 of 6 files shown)


dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
```diff
@@ -468,9 +475,16 @@ sycl::event topk_radix_impl(sycl::queue &exec_q,
...
+
     IndexTy *tmp_tp = sycl::malloc_device<IndexTy>(total_nelems, exec_q);
 
-    if (nullptr == workspace || nullptr == tmp_tp) {
+    if (nullptr == tmp_tp) {
+        using dpctl::tensor::alloc_utils::sycl_free_noexcept;
+        sycl_free_noexcept(workspace, exec_q);
         throw std::runtime_error(
             "Not enough device memory for radix sort topk");
     }
```
ndgrigorian, line 490: In addition to this, the `parallel_radix_sort_impl` call also can raise `runtime_error` if it fails to allocate more additional memory.

It should be wrapped in a `try-except` block to account for that to avoid another memory leak here.

At some point it would be nice to use `unique_ptr` here and avoid this entirely, but that would be a more in-depth refactoring.

ndgrigorian: The same change as above should be made in `radix_argsort_axis1_contig_impl` as well.

oleksandr-pavlyk: This has been implemented. Lifetime of USM allocations in "sorting/" is handled by a `unique_ptr`, which is released after the `host_task` has been submitted.

dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
```diff
@@ -505,17 +531,18 @@ sycl::event topk_radix_impl(sycl::queue &exec_q,
...
-            std::size_t src_idx = iter_gid * axis_nelems + axis_gid;
-            std::size_t dst_idx = iter_gid * k + axis_gid;
+            const std::size_t src_idx = iter_gid * axis_nelems + axis_gid;
+            const std::size_t dst_idx = gid;
 
-            IndexTy res_ind = tmp_tp[src_idx];
-            vals_tp[dst_idx] = arg_tp[res_ind];
-            inds_tp[dst_idx] = res_ind % axis_nelems;
+            const IndexTy res_ind = tmp_tp[src_idx];
+            const v = arg_tp[res_ind];
```
ndgrigorian, line 543:
```suggestion
            const argTy v = arg_tp[res_ind];
```

dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
```diff
@@ -525,12 +552,10 @@ sycl::event topk_radix_impl(sycl::queue &exec_q,
...
 
         using dpctl::tensor::alloc_utils::sycl_free_noexcept;
-        cgh.host_task([ctx, workspace, tmp_tp] {
-            sycl_free_noexcept(workspace, ctx);
-            sycl_free_noexcept(tmp_tp, ctx);
-        });
+        cgh.host_task([ctx, workspace] { sycl_free_noexcept(workspace, ctx); });
     });
 
+    workspace_owner.release();
```
ndgrigorian, line 558: I think it would be best to release the pointer before calling free.

Release also returns the original pointer.

This can be refactored to:
```c++
    auto work_space_ptr = workspace_owner.release();
    sycl::event cleanup_ev = exec_q.submit([&](sycl::handler &cgh) {
        cgh.depends_on(write_topk_ev);
        const sycl::context &ctx = exec_q.get_context();

        using dpctl::tensor::alloc_utils::sycl_free_noexcept;
        cgh.host_task([ctx, work_space_ptr] { sycl_free_noexcept(work_space_ptr, ctx); });
    });
```

If we plan to generalize this, it might also be interesting to construct some kind of variadic template for releasing some number of `unique_ptr`s wrapping USM allocations. Then we could cut this redundant and frequently used `host_task` code out everywhere.

dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
```diff
@@ -427,6 +443,7 @@ sycl::event topk_merge_impl(
                     [ctx, index_data] { sycl_free_noexcept(index_data, ctx); });
             });
 
+        index_data_owner.release();
```
ndgrigorian, line 446: The same thing should be done here (move release to before and use a temporary for the pointer).

dpctl/tensor/libtensor/include/utils/sycl_alloc_utils.hpp
```diff
@@ -78,6 +79,28 @@ template <typename T> void sycl_free_noexcept(T *ptr, sycl::queue &q) noexcept
...
+        sycl_free_noexcept(ptr, ctx_);
+    }
+};
+
+template <typename T>
+std::unique_ptr<T, USMDeleter> make_owner(T *ptr, const sycl::queue &q)
+{
+    auto usm_deleter = USMDeleter(q);
+    return std::unique_ptr<T, USMDeleter>(ptr, usm_deleter);
+}
```
ndgrigorian, line 102: I implemented something similar for the `dpnp.choose` PR.

A couple of remaining issues have been mentioned there, but it might be worth taking a look.
https://github.com/IntelPython/dpnp/pull/2201

The biggest difference is that, rather than allocating and then using the `unique_ptr` as functionally a deleter, it wraps the `malloc_device` call as well. And as was pointed out in the PR, it could wrap the `std::runtime_error` if it fails to allocate, too.

We could do something similar here and factor away the `malloc_device` calls as well. Then any function calls that require a raw pointer can be passed `usm_unique_ptr_obj.get()` to make clear that a raw pointer managed by `unique_ptr` is being used.

If we did something like this, we can also expose it for dpnp and it can be removed from the PR.

oleksandr-pavlyk: I am implemented `smart_malloc_device` to make allocation returning `unique_ptr`, and `async_smart_free` which takes the smart pointer, submits the `host_task` and releases the smart pointer.

dpctl/tensor/libtensor/include/utils/sycl_alloc_utils.hpp
```diff
@@ -73,11 +76,99 @@ void sycl_free_noexcept(T *ptr, const sycl::context &ctx) noexcept
...
+smart_malloc_shared(std::size_t count,
+                    const sycl::queue &q,
+                    const sycl::property_list &propList = {})
+{
+    return smart_malloc<T>(count, q, sycl::usm::alloc::shared, propList);
+}
+
+template <typename T>
+std::unique_ptr<T, USMDeleter>
+smart_malloc_jost(std::size_t count,
```
ndgrigorian, line 136:
```suggestion
smart_malloc_host(std::size_t count,
```

ndgrigorian:
Ping @AlexanderKalistratov 
If the utilities added here (`smart_malloc_*` functions) are merged in, they can be reused by dpnp and replace the similar code in choose PR

dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
```diff
@@ -465,71 +498,54 @@ sycl::event topk_radix_impl(sycl::queue &exec_q,
...
+    IndexTy *tmp_tp = workspace + padded_total_nelems;
 
     using IdentityProjT = radix_sort_details::IdentityProj;
     using IndexedProjT =
         radix_sort_details::IndexedProj<IndexTy, argTy, IdentityProjT>;
     const IndexedProjT proj_op{arg_tp, IdentityProjT{}};
 
+    using IotaKernelName = topk_iota_krn<argTy, IndexTy>;
+
+#if 1
```
ndgrigorian, line 516: Should these preprocessor directives be removed?

Besides these I think this looks ready to go in

dpctl/tensor/libtensor/include/utils/sycl_alloc_utils.hpp
```diff
@@ -73,11 +76,99 @@ void sycl_free_noexcept(T *ptr, const sycl::context &ctx) noexcept
...
+    sycl::event ht_e = exec_q.submit([&](sycl::handler &cgh) {
+        cgh.depends_on(depends);
+
+        cgh.host_task([ptrs, dels]() {
+            for (size_t i = 0; i < ptrs.size(); ++i) {
+                dels[i](ptrs[i]);
+            }
+        });
+    });
+    (args.release(), ...);
```
ndgrigorian, line 167:
```suggestion
    (ptrs.push_back(reinterpret_cast<void *>(args.release())), ...);

    std::vector<USMDeleter> dels;
    dels.reserve(n);
    (dels.push_back(args.get_deleter()), ...);

    sycl::event ht_e = exec_q.submit([&](sycl::handler &cgh) {
        cgh.depends_on(depends);

        cgh.host_task([ptrs, dels]() {
            for (size_t i = 0; i < ptrs.size(); ++i) {
                dels[i](ptrs[i]);
            }
        });
    });
```
since `release` returns the pointer anyway

dpctl/tensor/libtensor/include/utils/sycl_alloc_utils.hpp
```diff
@@ -73,11 +76,99 @@ void sycl_free_noexcept(T *ptr, const sycl::context &ctx) noexcept
...
+{
+    return smart_malloc<T>(count, q, sycl::usm::alloc::host, propList);
+}
+
+template <typename... Args>
+sycl::event async_smart_free(sycl::queue &exec_q,
+                             const std::vector<sycl::event> &depends,
+                             Args &&...args)
+{
+    constexpr std::size_t n = sizeof...(Args);
```
ndgrigorian, line 148: Also, big edge case but this compiles fine with no final argument (i.e., `n == 0`).

We could raise an exception in this case, or we could wait on `depends` and not launch the `host_task` (return empty event?), or do neither and let it through.

ndgrigorian: I favor the assertion, given that it would almost definitely be a mistake.

dpctl/tensor/libtensor/include/utils/sycl_alloc_utils.hpp
ndgrigorian: It may make more sense (if this change is made) to build the `dels` vector with `get_deleter` before calling the `release` and acquiring pointers.

dpctl/tensor/libtensor/include/kernels/sorting/topk.hpp
oleksandr-pavlyk: done. Thanks for catching these instances.

dpctl/tensor/libtensor/include/utils/sycl_alloc_utils.hpp
oleksandr-pavlyk: I think we should release only after submission of `host_task` is known to have worked, which is after we obtained the event. Otherwise we risk leaking USM allocation is submission raises an exception.

ndgrigorian: > I think we should release only after submission of `host_task` is known to have worked, which is after we obtained the event. Otherwise we risk leaking USM allocation is submission raises an exception.

Good point!

ndgrigorian (approved):
This LGTM, we can merge this into the topk branch and drop the test file PR, then remove the commit that adds `test_top_k_largest_1d_radix_i1`

PR #1950, opened 2024-12-26
[MAINT] Use `std::size_t` from `cstddef` and use `dpctl::tensor::ssize_t` where `ssize_t` is used

ndgrigorian:
This PR changes the reliance of `dpctl` headers on `size_t` being translated by the compiler by instead including `cstddef` and replacing `size_t` with `std::size_t` throughout.

Additionally, `using dpctl::tensor::ssize_t` is used wherever `ssize_t` comes up.

### DIFF
--- a/dpctl/apis/include/dpctl4pybind11.hpp
+++ b/dpctl/apis/include/dpctl4pybind11.hpp
@@ -27,8 +27,10 @@
 
 #include "dpctl_capi.h"
 #include <complex>
+#include <cstddef> // for std::size_t for C++ linkage
 #include <memory>
 #include <pybind11/pybind11.h>
+#include <stddef.h> // for size_t for C linkage
 #include <stdexcept>
 #include <sycl/sycl.hpp>
 #include <utility>
@@ -759,7 +761,7 @@ public:
      *  lifetime of the USM allocation.
      */
     usm_memory(void *usm_ptr,
-               size_t nbytes,
+               std::size_t nbytes,
                const sycl::queue &q,
                std::shared_ptr<void> shptr)
     {
@@ -819,7 +821,7 @@ public:
         return reinterpret_cast<char *>(MRef);
     }
 
-    size_t get_nbytes() const
+    std::size_t get_nbytes() const
     {
         auto const &api = ::dpctl::detail::dpctl_capi::get();
         Py_MemoryObject *mem_obj = reinterpret_cast<Py_MemoryObject *>(m_ptr);
--- a/libsyclinterface/tests/test_service.cpp
+++ b/libsyclinterface/tests/test_service.cpp
@@ -39,7 +39,7 @@ TEST(TestServicesFns, ChkDPCPPVersion)
     ASSERT_TRUE(ver.length() > 0);
 
     std::string ver_from_cmplr(ASSTR(__VERSION__));
-    std::size_t found = ver_from_cmplr.find(ver);
+    auto found = ver_from_cmplr.find(ver);
 
     // version returned by DPCTLService_GetDPCPPVersion
     // should occur as a substring in the version obtained
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add support of CV-qualifiers in `is_complex<T>` helper [gh-1900](https://github.com/IntelPython/dpctl/pull/1900)
 * Tuning work for elementwise functions with modest performance gains (under 10%) [gh-1889](https://github.com/IntelPython/dpctl/pull/1889)
 * Support for Python 3.13 for `dpctl` [gh-1941](https://github.com/IntelPython/dpctl/pull/1941)
+* Change libtensor to use `std::size_t` and `dpctl::tensor::ssize_t` throughout and fix missing includes for `std::size_t` and `size_t` [gh-1950](https://github.com/IntelPython/dpctl/pull/1950)
 
 ## [0.18.3] - Dec. 07, 2024
 
--- a/dpctl/_host_task_util.hpp
+++ b/dpctl/_host_task_util.hpp
@@ -31,6 +31,7 @@
 
 #pragma once
 #include <exception>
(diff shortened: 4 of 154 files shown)


oleksandr-pavlyk:
@ndgrigorian Looks like these two instances were missed:

```
(dev_dpctl) opavlyk@opavlyk-mobl:~/repos/dpctl$ grep -r size_t dpctl/tensor/libtensor/ | grep -v "std::size_t" | grep -v "ssize_t" | grep -v "size_to"
dpctl/tensor/libtensor/include/kernels/elementwise_functions/trunc.hpp:                              size_t nelems,
dpctl/tensor/libtensor/include/kernels/elementwise_functions/trunc.hpp:                   size_t nelems,
```

ndgrigorian:
> @ndgrigorian Looks like these two instances were missed:
> 
> ```
> (dev_dpctl) opavlyk@opavlyk-mobl:~/repos/dpctl$ grep -r size_t dpctl/tensor/libtensor/ | grep -v "std::size_t" | grep -v "ssize_t" | grep -v "size_to"
> dpctl/tensor/libtensor/include/kernels/elementwise_functions/trunc.hpp:                              size_t nelems,
> dpctl/tensor/libtensor/include/kernels/elementwise_functions/trunc.hpp:                   size_t nelems,
> ```

I've pushed a change fixing these, good catch!

oleksandr-pavlyk:
@ndgrigorian These changes look good to me. I think it is ready for review and ready to be merged in. 

One thing to check is whether changes here might conflict with changes in feature/topk branch

ndgrigorian:
> @ndgrigorian These changes look good to me. I think it is ready for review and ready to be merged in.
> 
> One thing to check is whether changes here might conflict with changes in feature/topk branch

I'll go ahead and check if they do. I will also have to rebase onto master and check for more instances with the accumulator improvement PR merged.

oleksandr-pavlyk:
The generate coverage workflow failed with 8 ctest failures related to `sycl::kernel_bundle` and `sycl::kernel`

oleksandr-pavlyk:
The reason for failures is the `std::size_t` should not he used in declarations and definitions of `EXTERN "C"` functions due to name mangling

oleksandr-pavlyk:
The C compatible type is defined in `"stddef.h"`

ndgrigorian:
> The reason for failures is the `std::size_t` should not he used in declarations and definitions of `EXTERN "C"` functions due to name mangling

Uses in `libsyclinterface` have been removed

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -93,9 +94,9 @@ class dpctl_capi
     void *(*Memory_GetOpaquePointer_)(Py_MemoryObject *);
     DPCTLSyclContextRef (*Memory_GetContextRef_)(Py_MemoryObject *);
     DPCTLSyclQueueRef (*Memory_GetQueueRef_)(Py_MemoryObject *);
-    size_t (*Memory_GetNumBytes_)(Py_MemoryObject *);
+    std::size_t (*Memory_GetNumBytes_)(Py_MemoryObject *);
```
oleksandr-pavlyk, line 97: `Memory_GtNumBytes`, as well as `Memory_Make` and other `dpctl` C-API symbol use C linkage (no name scrambling), hence they should use `size_t` instead of `std::size_t`.

libsyclinterface/tests/test_service.cpp
```diff
@@ -39,7 +40,7 @@ TEST(TestServicesFns, ChkDPCPPVersion)
     ASSERT_TRUE(ver.length() > 0);
 
     std::string ver_from_cmplr(ASSTR(__VERSION__));
-    std::size_t found = ver_from_cmplr.find(ver);
+    size_t found = ver_from_cmplr.find(ver);
```
oleksandr-pavlyk, line 43: I would just use `auto found = ver_from_cmplr.find(ver)`, since `std::string::find` is documented to have return type `std::string::size_type`.

dpctl/apis/include/dpctl4pybind11.hpp
ndgrigorian: I'll remove it completely from `dpctl4pybind11.hpp` and just include `stddef.h`, since it should turn out to be the same either way.

ndgrigorian:
> @ndgrigorian These changes look good to me. I think it is ready for review and ready to be merged in.
> 
> One thing to check is whether changes here might conflict with changes in feature/topk branch

Unfortunately there are conflicts in `merge_sort.hpp` and `radix_sort.hpp`

It may be best to merge this after topk

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -1258,18 +1259,18 @@ struct ManagedMemory
 
 } // end of namespace detail
 
-template <std::size_t num>
+template <size_t num>
```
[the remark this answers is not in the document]
ndgrigorian: We would need to add `cstddef` to `dpctl4pybind11` as well.

My only concern is that there may be confusion if both `size_t` and `std::size_t` are being used in the same file—but maybe a comment addressing it where `cstddef` is included would be sufficient.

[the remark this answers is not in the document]
ndgrigorian: Done, and added comments

ndgrigorian:
@oleksandr-pavlyk 
I've also reverted use of `std::size_t` in `_host_task_utils.hpp` since `async_dec_ref`

In the process, I've also noticed that we seemingly use `py::ssize_t` as the type for the number of elements in an array for `get_size` method and in `size` Python equivalent, while throughout libtensor Python bindings, we use `size_t` to represent the number of elements.

This inconsistency can be resolved in a later PR, but we should probably use `size_t` everywhere for this.

ndgrigorian:
@oleksandr-pavlyk 
topk is merged and conflicts with this branch and master are resolved

oleksandr-pavlyk:
The GH rules require an approval from somebody other than the author of the last commit.

Please add a change to changelog to work around that issue

PR #1951, opened 2024-12-29
Work-around for issue with CPU driver, add tests

oleksandr-pavlyk:
Due to a reported issue with OpenCL:CPU device implementation, the kernel for short inputs and short signed integral types may produce incorrect results when compiled for short SIMD width (as chosen automatically for AMD EPYC 7763 64 Processors CPU).

A work-around is to introduce a redundant barrier call (only for CPU devices when sub-group is short).

Also edits made while triaging the issue are applied: counters for short input arrays are stored in 16-bit unsigned integers, rather than in 32-bit ones, `uint16_t` replaced with `std::uint16_t`, etc.

Tests for `tensor.top_k` were expanded to ensure 100% coverage.

### DIFF
--- a/.github/workflows/os-llvm-sycl-build.yml
+++ b/.github/workflows/os-llvm-sycl-build.yml
@@ -9,7 +9,7 @@ permissions: read-all
 jobs:
   install-compiler:
     name: Build with nightly build of DPC++ toolchain
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     timeout-minutes: 90
 
     env:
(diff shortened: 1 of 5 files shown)


oleksandr-pavlyk:
@ndgrigorian For some reason `pytest` fails to recognize the added command-line option, and I can reproduce that locally (install the build artifact created by CI and try the command CI is trying from an empty folder, but I do not understand why it fails).

ndgrigorian:
> @ndgrigorian For some reason `pytest` fails to recognize the added command-line option, and I can reproduce that locally (install the build artifact created by CI and try the command CI is trying from an empty folder, but I do not understand why it fails).

Yes, and it works fine in the coverage CI as well.

oleksandr-pavlyk:
@ndgrigorian The CI is green now

oleksandr-pavlyk:
All line of `top_k` function implementation are now covered.

I think this PR is ready to be merged into the targeted base branch.

Let me only rebase to squash some commits.

ndgrigorian (approved):
Tests now passing, LGTM!

PR #1957, opened 2025-01-08
Technical debt changes

oleksandr-pavlyk:
This PR checks in some technical debt changes.

- Do not user anonymous namespace in C++ header files.  

*Rationale*: Entities defined in anonymous namespace have internal linkage, that is, they are generated in every translation units that includes the header file bloating the size of compiled binary, as well increasing compilation time

- Move inline kernel submissions for tasks that do not depend on all template parameters of the function to a separate function.

*Rationale*: Inline use requires kernel names to be distinct for function different instantiations, producing multiple copies of identical kernels, bloating the binary. 

This change alone shed 13MB off the size of `_tensor_accumulation_impl` native extension (from 49MB to 36MB).

---

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
@@ -27,7 +27,6 @@
 #include <cstddef>
 #include <cstdint>
 #include <limits>
-#include <new>
 #include <sycl/sycl.hpp>
 #include <utility>
 #include <vector>
@@ -170,15 +169,6 @@ template <typename inputT,
           typename ScanOpT,
           bool include_initial>
 class inclusive_scan_iter_local_scan_striped_krn;
-
-template <typename inputT,
-          typename outputT,
-          nwiT n_wi,
-          typename TransformerT,
-          typename OtherTransformerT,
-          typename ScanOpT,
-          bool include_initial>
-class inclusive_scan_iter_chunk_update_krn;
 
 template <typename inputT,
           typename outputT,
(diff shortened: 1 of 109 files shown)


oleksandr-pavlyk:
@AlexanderKalistratov I pushed the changes you suggested to make in `async_smart_free` in 80f288ccafe8de960e07effe19cdef64df5784b9

ndgrigorian:
The changes in this PR have also reduced the build time of dpctl, by about 10 minutes for me

this branch
```
34m33.407s
```
master
```
real    45m25.987s
```

dpctl/tensor/libtensor/include/kernels/accumulators.hpp
```diff
@@ -1255,20 +1222,19 @@ std::size_t cumsum_val_contig_impl(sycl::queue &q,
...
         cgh.depends_on(comp_ev);
         cgh.copy<cumsumT>(last_elem, last_elem_host_usm, 1);
     });
     copy_e.wait();
     std::size_t return_val = static_cast<std::size_t>(*last_elem_host_usm);
 
-    using dpctl::tensor::alloc_utils::sycl_free_noexcept;
-    sycl_free_noexcept(last_elem_host_usm, q);
+    // free USM host allocation
+    host_usm_owner.reset(nullptr);
```
AlexanderKalistratov, line 1237: You actually could call `reset` without parameters:
```c++
host_usm_owner.reset();
```

Also I'm not quite sure why you need to call it here at all.

dpctl/tensor/libtensor/include/kernels/accumulators.hpp
```diff
@@ -1370,20 +1336,19 @@ cumsum_val_strided_impl(sycl::queue &q,
...
         cgh.depends_on(comp_ev);
         cgh.copy<cumsumT>(last_elem, last_elem_host_usm, 1);
     });
     copy_e.wait();
     std::size_t return_val = static_cast<std::size_t>(*last_elem_host_usm);
 
-    using dpctl::tensor::alloc_utils::sycl_free_noexcept;
-    sycl_free_noexcept(last_elem_host_usm, q);
+    // free USM-host temporary
+    host_usm_owner.reset(nullptr);
```
AlexanderKalistratov, line 1351: Same question here

dpctl/tensor/libtensor/include/kernels/accumulators.hpp
oleksandr-pavlyk: This was added to make it explicit that USM-host allocation was being freed. 

If removed, it would be done implicitly. Doing in it explicitly improves readability and maintainability, in my opinion.
I added a comment to make that intent clear.

dpctl/tensor/libtensor/include/kernels/accumulators.hpp
oleksandr-pavlyk: Comment added.

ndgrigorian (approved):
Changes LGTM overall

Thank you @oleksandr-pavlyk

PR #1964, opened 2025-01-13
Usm ndrray array protocol must raise

oleksandr-pavlyk:
Implement `usm_ndarray.__array__` that raises `TypeError` to disallow implicit conversion of `usm_ndarray` to `numpy.ndarray` with `dtype=object`.

### DIFF
--- a/dpctl/tensor/_usmarray.pyx
+++ b/dpctl/tensor/_usmarray.pyx
@@ -1552,6 +1552,14 @@ cdef class usm_ndarray:
     def __repr__(self):
         return usm_ndarray_repr(self)
 
+    def __array__(self, dtype=None, copy=None):
+        "NumPy array protocol"
+        raise TypeError(
+            "Implicit conversion to a NumPy array is not allowed. "
+	    "Use `dpctl.tensor.asnumpy` to copy data from this "
+	    "`dpctl.tensor.usm_ndarray` instance to NumPy array"
+	)
+
 
 cdef usm_ndarray _real_view(usm_ndarray ary):
     """
--- a/dpctl/tensor/_usmarray.pyx
+++ b/dpctl/tensor/_usmarray.pyx
@@ -1552,8 +1552,15 @@ cdef class usm_ndarray:
     def __repr__(self):
         return usm_ndarray_repr(self)
 
-    def __array__(self, dtype=None, copy=None):
-        "NumPy array protocol"
+    def __array__(self, dtype=None, /, *, copy=None):
+        """NumPy's array protocol method to disallow implicit conversion.
+
+	Without this definition, `numpy.asarray(usm_ar)` converts
+	usm_ndarray instance into NumPy array with data type `object`
+	and every element being 0d usm_ndarray.
+
+        https://github.com/IntelPython/dpctl/pull/1384#issuecomment-1707212972
+	"""
         raise TypeError(
             "Implicit conversion to a NumPy array is not allowed. "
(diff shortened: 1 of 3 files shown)


ndgrigorian:
Should we define copy and dtype parameters?

NumPy claims it will raise warnings otherwise, but of course, this is just here to raise an error—not sure if it will be a problem or not.

oleksandr-pavlyk:
@ndgrigorian Could you please cite where sych claim is made? I looked at https://numpy.org/doc/stable/reference/generated/numpy.asarray.html and could not find it

oleksandr-pavlyk:
> Should we define copy and dtype parameters?
> 

If you mean in the test, than we could, but it should not make a difference.

ndgrigorian:
> @ndgrigorian Could you please cite where sych claim is made? I looked at https://numpy.org/doc/stable/reference/generated/numpy.asarray.html and could not find it

https://numpy.org/doc/2.1/user/basics.interoperability.html#dunder-array-interface
It is mentioned here

> The signature of the method should be __array__(self, dtype=None, copy=None). If a passed dtype isn’t None and different than the object’s data type, a casting should happen to a specified type. If copy is None, a copy should be made only if dtype argument enforces it. For copy=True, a copy should always be made, where copy=False should raise an exception if a copy is needed.
> 
> If a class implements the old signature __array__(self), for np.array(a) a warning will be raised saying that dtype and copy arguments are missing.

oleksandr-pavlyk:
I have changed `__array__` from being a property to being a method with the signature NumPy expects. Thank you for catching this error @ndgrigorian

dpctl/tensor/_usmarray.pyx
```diff
@@ -1552,6 +1552,14 @@ cdef class usm_ndarray:
     def __repr__(self):
         return usm_ndarray_repr(self)
 
+    def __array__(self, dtype=None, copy=None):
```
antonwolfy, line 1555:
```suggestion
    def __array__(self, dtype=None, /, *, copy=None)
```
based on numpy implementation

oleksandr-pavlyk: Please cite the source. 

The only consequence of this change is that `usm_arr.__array__(numpy.int32, False)` would raise `TypeError` based on unexpected number of positional arguments, rather than `TypeError` informing of implicit conversion to numpy.

```

In [1]: def foo(n, dtype=None, copy=None):
   ...:     pass
   ...:

In [2]: def foo2(n, dtype=None, /, *, copy=None):
   ...:     pass
   ...:

In [3]: foo(1, 'i4')

In [4]: foo2(1, 'i4')

In [5]: foo(1, 'i4', False)

In [6]: foo2(1, 'i4', False)
Cell In[6], line 1
----> 1 foo2(1, 'i4', False)

TypeError: foo2() takes from 1 to 2 positional arguments but 3 were given
```

Also, this examples define `__array__(self, dtype=None, copy=None)` signature:

https://github.com/numpy/numpy/blob/bbf4836a82d255c2a2161d277abb4b6f5455ca6d/doc/source/user/basics.dispatch.rst

antonwolfy: I was based on signature define for `class ndarray` in NumPy repo:

https://github.com/numpy/numpy/blob/bbf4836a82d255c2a2161d277abb4b6f5455ca6d/numpy/__init__.pyi#L2048

ndgrigorian (approved):
LGTM!

PR #1975, opened 2025-01-20
Remove const qualifiers for class/struct members per C++ guidelines

oleksandr-pavlyk:
Further remove `const`  qualifier from struct/class members per C++ guidelines.

The test should be that each such class instance be device copyable. Perhaps static asserts to this effect can be added to the code.


ndgrigorian (approved):
LGTM! Honestly it's fewer than expected

PR #1976, opened 2025-01-21
Reduce elementwise extension size

oleksandr-pavlyk:
This PR factors out inline submit to populate padded vector in specializations for binary operations on matrix and a vector.

Doing so allows to generate fewer of such kernels, resulting in the binary size decrease.

Before:
```
(dev_dpctl) opavlyk@mtl-world:~/repos/dpctl$ ls -l dpctl/tensor/_tensor_elementwise_impl.cpython-312-x86_64-linux-gnu.so
-rw-r--r-- 1 opavlyk opavlyk 38659896 Jan 19 20:58 dpctl/tensor/_tensor_elementwise_impl.cpython-312-x86_64-linux-gnu.so
```

After:
```
(dev_dpctl) opavlyk@mtl-world:~/repos/dpctl$ ls -l dpctl/tensor/_tensor_elementwise_impl.cpython-312-x86_64-linux-gnu.so
-rw-r--r-- 1 opavlyk opavlyk 37176600 Jan 21 06:36 dpctl/tensor/_tensor_elementwise_impl.cpython-312-x86_64-linux-gnu.so
```

Also this PR adds `static_assert` in `offset_utils.hpp` to verify that indexers are device copyable.

It also sneaks in changes of defining local typename for the functor being submitted in `cgh.parallel_for` to simplify the invocation.

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/clip.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/clip.hpp
@@ -218,7 +218,6 @@ sycl::event clip_contig_impl(sycl::queue &q,
             using KernelName = clip_contig_kernel<T, vec_sz, n_vecs>;
             using Impl =
                 ClipContigFunctor<T, vec_sz, n_vecs, enable_sg_loadstore>;
-            static_assert(sycl::is_device_copyable_v<Impl>);
 
             cgh.parallel_for<KernelName>(
                 sycl::nd_range<1>(gws_range, lws_range),
@@ -231,7 +230,6 @@ sycl::event clip_contig_impl(sycl::queue &q,
                 disabled_sg_loadstore_wrapper_krn<InnerKernelName>;
             using Impl =
                 ClipContigFunctor<T, vec_sz, n_vecs, disable_sg_loadstore>;
-            static_assert(sycl::is_device_copyable_v<Impl>);
 
             cgh.parallel_for<KernelName>(
                 sycl::nd_range<1>(gws_range, lws_range),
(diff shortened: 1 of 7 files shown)


dpctl/tensor/libtensor/include/kernels/clip.hpp
```diff
@@ -216,22 +216,26 @@ sycl::event clip_contig_impl(sycl::queue &q,
         {
             constexpr bool enable_sg_loadstore = true;
             using KernelName = clip_contig_kernel<T, vec_sz, n_vecs>;
+            using Impl =
+                ClipContigFunctor<T, vec_sz, n_vecs, enable_sg_loadstore>;
+            static_assert(sycl::is_device_copyable_v<Impl>);
```
ndgrigorian, line 221: It may make sense to add these assertions everywhere that a kernel functor is used, but that can be deferred to a subsequent PR.

ndgrigorian (dismissed):
LGTM!

dpctl/tensor/libtensor/include/kernels/clip.hpp
oleksandr-pavlyk: Actually no, I should remove this one too. For example, `sycl::local_accessor<T, dim>` is not device copyable, so all functors that work with local accessors won't be. 

I started adding these assertions and realized they were a mistake. Will remove this one too. Indexers should be device copyable though. Basic types are by out of the box.

ndgrigorian (approved):

PR #1979, opened 2025-01-23
Fix array API inspection behavior with `device` keyword

### LINKED ISSUE #1977: `__array_namespace_info__` returns `complex128` dtype when no fp64
It seems dpctl only checks if `float64` is supported based on device capabilities when building a dictionary of all Array API data types:
```python

dpctl.__version__
# Out: '0.19.0dev0+480.g360ce45dc6'

info = dpnp.__array_namespace_info__()
info.dtypes()
# Out:
# {'bool': dtype('bool'),
#  'float32': dtype('float32'),
#  'complex64': dtype('complex64'),
#  'complex128': dtype('complex128'),
#  'int8': dtype('int8'),
#  'int16': dtype('int16'),
#  'int32': dtype('int32'),
#  'int64': dtype('int64'),
#  'uint8': dtype('uint8'),
#  'uint16': dtype('uint16'),
#  'uint32': dtype('uint32'),
#  'uint64': dtype('uint64')}

# no fp64 support by default device:
dpctl.select_default_device().has_aspect_fp64
# Out: False
```
As a result `float64` is filtered out but `complex128` is present in the output dictionary, which looks faulty.

### LINKED ISSUE #1978: `__array_namespace_info__.dtypes` raises `AttributeError` for device passed as a string
Based on dpctl [documentation](https://intelpython.github.io/dpctl/latest/api_reference/dpctl/generated/generated/dpctl.tensor._array_api.Info.dtypes.html) `Info.dtypes` support a string value of `device` keyword argument:

> device (Optional[:class:`dpctl.SyclDevice`, :class:`dpctl.SyclQueue`, :class:`dpctl.tensor.Device`, str])

but the example below does not work and raises `AttributeError` exception:
```python

dpctl.__version__
# Out: '0.19.0dev0+480.g360ce45dc6'

info = dpnp.__array_namespace_info__()
info.dtypes(device="gpu")
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 info.dtypes(device="gpu")

File ~/miniforge3/envs/dpnp_dev/lib/python3.11/site-packages/dpctl/tensor/_array_api.py:203, in Info.dtypes(self, device, kind)
    201 elif isinstance(device, dpt.Device):
    202     device = device.sycl_device
--> 203 _fp64 = device.has_aspect_fp64
    204 if kind is None:
    205     return {
    206         key: val
    207         for key, val in self._all_dtypes.items()
    208         if (key != "float64" or _fp64)
    209     }

AttributeError: 'str' object has no attribute 'has_aspect_fp64'
```

ndgrigorian:
This PR proposes changing the handling of the `device` keyword in the array API inspection utilities to:

- raise a more helpful exception for unexpected inputs
- work correctly with filter strings and SyclQueues

Also fixes the `dtypes` utility returning `complex128` in its output when the provided device does not support double precision types.

Resolves #1978 , resolves #1977 

### DIFF
--- a/dpctl/tensor/_array_api.py
+++ b/dpctl/tensor/_array_api.py
@@ -49,6 +49,20 @@ def _isdtype_impl(dtype, kind):
         raise TypeError(f"Unsupported data type kind: {kind}")
 
 
+def _get_device_impl(d):
+    if d is None:
+        return dpctl.select_default_device()
+    elif isinstance(d, dpctl.SyclDevice):
+        return d
+    elif isinstance(d, (dpt.Device, dpctl.SyclQueue)):
+        return d.sycl_device
+    else:
+        try:
+            return dpctl.SyclDevice(d)
+        except TypeError:
+            raise TypeError(f"Unsupported type for device argument: {type(d)}")
+
+
 __array_api_version__ = "2023.12"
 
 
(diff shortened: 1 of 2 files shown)


ndgrigorian:
@antonwolfy 
This is now ready for your review

antonwolfy (approved):
I checked with dpnp changes added in [dpnp-2275](https://github.com/IntelPython/dpnp/pull/2275) and it works as expected now.
Thank you @ndgrigorian for a quick fix. LGTM!

dpctl/tensor/_array_api.py
```diff
@@ -196,22 +207,20 @@ def dtypes(self, *, device=None, kind=None):
...
                 for key, val in self._all_dtypes.items()
-                if (key != "float64" or _fp64)
+                if _fp64 or (key != "float64" and key != "complex128")
             }
         else:
             return {
                 key: val
                 for key, val in self._all_dtypes.items()
-                if (key != "float64" or _fp64) and _isdtype_impl(val, kind)
+                if (_fp64 or (key != "float64" and key != "complex128"))
```
oleksandr-pavlyk, line 222: Perhaps it is useful to also check for fp16 aspect and filter "float16" out if it can be present.

oleksandr-pavlyk (approved):

dpctl/tensor/_array_api.py
ndgrigorian: It will already not be present, as it is not mandated by array API, and array API strictly requires only data types present in array API be present in dtypes list ([see here](https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.dtypes.html#array_api.info.dtypes))

PR #1980, opened 2025-01-27
[MAINT] Unpin conda-build

ndgrigorian:
This PR proposes unpinning conda-build in the conda-package workflow, as well as updating the workflow to account for conda-build (after 25.1.0) moving to producing .conda files instead of tarballs.

The change to .conda files is motivated by improvements to package compression, which bring dpctl from ~24 MB to ~13.5 MB, about a 40% decrease in size.

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -48,7 +48,7 @@ jobs:
         run: echo $CONDA/bin >> $GITHUB_PATH
       - name: Install conda-build
         # FIXME: unpin when conda-build fixes issues with lief
-        run: conda install conda-build"<=24.11.2" -c conda-forge --override-channels
+        run: conda install conda-build -c conda-forge --override-channels
       - name: Store conda paths as envs
         shell: bash -l {0}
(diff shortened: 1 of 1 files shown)


antonwolfy (approved):
LGTM

oleksandr-pavlyk:
PR is lacking description of intent of the change. Is the reason for the change to produce `.conda` artifacts, or is the change to `.conda` consequence of updating version of conda-build, which was on a to-do list for some time?

ndgrigorian:
> PR is lacking description of intent of the change. Is the reason for the change to produce `.conda` artifacts, or is the change to `.conda` consequence of updating version of conda-build, which was on a to-do list for some time?

Yes, great point, I'll fix that by adding a PR description. :)

It's a consequence of the workflow using conda-build 25.1.1, which produces .conda instead of tarballs by default. The reason is apparently due to [improved compression](https://github.com/conda/conda-build/issues/5183).

ndgrigorian:
For example of the compression, before:
Before

after:
After

PR #1982, opened 2025-01-31
Fix incorrect tests in `test_usm_ndarray_dlpack`

ndgrigorian:
The test `test_dlpack_device` attempted to query a device using DLPack device ID from the list of root devices created by `all_root_devices` fixture in `test_usm_ndarray_dlpack`.

This logic is incorrect, as `all_root_devices` caches only the first two root devices per platform. As the DLPack device ID is based on the ordinal ID of the device as it is canonically presented in `get_devices`, this could fail if a given platform had more than 2 associated devices (in this case, for a machine with 2 PVC. Each has 2 stacks, and therefore, 4 `level_zero` and `opencl:gpu` devices).

### DIFF
--- a/dpctl/tests/test_usm_ndarray_dlpack.py
+++ b/dpctl/tests/test_usm_ndarray_dlpack.py
@@ -84,7 +84,7 @@ def test_dlpack_device(usm_type, all_root_devices):
         assert type(dev) is tuple
         assert len(dev) == 2
         assert dev[0] == device_oneAPI
-        assert sycl_dev == all_root_devices[dev[1]]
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk (approved):
Yes, you're right. Expect this change to slow down run-time of this test file on the machine with many cards, but correctness first.

Was it failing in Tiber?

ndgrigorian:
> Yes, you're right. Expect this change to slow down run-time of this test file on the machine with many cards, but correctness first.
> 
> Was it failing in Tiber?

It was during a test run on a machine with two 1550 Max cards.

PR #1985, opened 2025-02-04
Copy from/to multidimensional buffers

sommerlukas:
Using typed memoryviews restricts queue `memcpy` to buffers with a single dimension only.

With this change, multi-dimensional buffers with contiguous layout can now be `memcpy`'d.

### DIFF
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -66,13 +66,13 @@ from .enum_types import backend_type
 
 from cpython cimport pycapsule
 from cpython.buffer cimport (
-    PyObject_CheckBuffer,
     Py_buffer,
-    PyObject_GetBuffer,
-    PyBUF_SIMPLE,
     PyBUF_ANY_CONTIGUOUS,
+    PyBUF_SIMPLE,
     PyBUF_WRITABLE,
-    PyBuffer_Release
+    PyBuffer_Release,
+    PyObject_CheckBuffer,
+    PyObject_GetBuffer,
 )
 from cpython.ref cimport Py_DECREF, Py_INCREF, PyObject
 from libc.stdlib cimport free, malloc
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -372,6 +372,8 @@ cdef DPCTLSyclEventRef _memcpy_impl(
     elif _is_buffer(dst):
         ret_code = PyObject_GetBuffer(dst, &dst_buf_view, PyBUF_SIMPLE | PyBUF_ANY_CONTIGUOUS | PyBUF_WRITABLE)
         if ret_code != 0: # pragma: no cover
+            if src_is_buf:
+                PyBuffer_Release(&src_buf_view)
             raise RuntimeError("Could not access buffer")
         c_dst_ptr = dst_buf_view.buf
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk (approved):
Thank you for the fix @sommerlukas !

The change looks good to me, lets wait for the CI though

dpctl/_sycl_queue.pyx
```diff
@@ -338,14 +346,20 @@ cdef DPCTLSyclEventRef _memcpy_impl(
...
+    cdef bint dst_is_buf = False
+    cdef int ret_code = 0
 
     if isinstance(src, _Memory):
         c_src_ptr = <void*>(<_Memory>src).get_data_ptr()
     elif _is_buffer(src):
-        src_host_buf = src
-        c_src_ptr = <void *>&src_host_buf[0]
+        ret_code = PyObject_GetBuffer(src, &src_buf_view, PyBUF_SIMPLE | PyBUF_ANY_CONTIGUOUS)
+        if ret_code != 0:
```
ndgrigorian, line 359:
```suggestion
        if ret_code != 0: # pragma: no cover
```

I don't think there's any good way to test these errors, so we can probably just ignore them for the coverage

Same for line 374

ndgrigorian (dismissed):
LGTM! Thank you @sommerlukas

dpctl/_sycl_queue.pyx
```diff
@@ -356,8 +370,11 @@ cdef DPCTLSyclEventRef _memcpy_impl(
     if isinstance(dst, _Memory):
         c_dst_ptr = <void*>(<_Memory>dst).get_data_ptr()
     elif _is_buffer(dst):
-        dst_host_buf = dst
-        c_dst_ptr = <void *>&dst_host_buf[0]
+        ret_code = PyObject_GetBuffer(dst, &dst_buf_view, PyBUF_SIMPLE | PyBUF_ANY_CONTIGUOUS | PyBUF_WRITABLE)
+        if ret_code != 0: # pragma: no cover
+            raise RuntimeError("Could not access buffer")
```
oleksandr-pavlyk, line 375: Should the source buffer be released before the exception is raised?

ndgrigorian:
Noting that the new test passes on both HIP and CUDA devices when the nd-memory-copy branch is built

dpctl/_sycl_queue.pyx
oleksandr-pavlyk: I was asking about the case thar gret buffer was successful on source buffer but failed on destination buffer

sommerlukas: Sorry, I misunderstood earlier.

It seems unlikely that both `src` and `dst` are Python buffers (we'd essentially copy from Python buffer to Python buffer in that case), but I've still added a fix to release the source buffer in that case.

oleksandr-pavlyk: Thanks for pushing the fix @sommerlukas . The use case of copying from host to host is also supported by `queue::memcpy`.

sommerlukas: Ok 👍 

This should be safe now with the latest change.

oleksandr-pavlyk (approved):
LGTM! Thank you @sommerlukas

PR #1028, opened 2022-12-22
Transitioned dpctl to require DPC++ 2023, removed host related functions, fixed linker crash, enabled SyclKernel.max_sub_group_size property

oleksandr-pavlyk:
The crash manifested itself as "Relocation trucated to fit" error, and was caused by the large size of device code produced in debug build. The suggested solution is to use link option `-fsycl-link-huge-device-code`.

See https://github.com/intel/llvm/blob/sycl/sycl/doc/UsersManual.md#link-options

```
FAILED: dpctl/tensor/_tensor_impl.cpython-39-x86_64-linux-gnu.so
[...]

PR #1992, opened 2025-02-11
Add `get_devices` method to `dpctl.SyclPlatform` class

### LINKED ISSUE #509: dpctl.SyclPlatform should provide get_devices method
This would correspond to ``sycl::platform::get_devices(deviceType)``, see table 16 in 4.6.2.1 of SYCL 2020 standard

https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#table.members.platform

ndgrigorian:
This PR proposes implementing the `get_devices` method to `dpctl.SyclPlatform` class as per SYCL 2020 spec, which specifies including a `device_type` filter.

Cleans up some remaining references to host devices, missed in https://github.com/IntelPython/dpctl/pull/1028

Closes #509 

### DIFF
--- a/dpctl/_sycl_device_factory.pyx
+++ b/dpctl/_sycl_device_factory.pyx
@@ -154,12 +154,12 @@ cpdef list get_devices(backend=backend_type.all, device_type=device_type_t.all):
     ``backend`` in addition to only ``device_type``.
 
     Args:
-        backend (optional):
+        backend (str, :class:`dpctl.backend_type`, optional):
             A :class:`dpctl.backend_type` enum value or a string that
             specifies a SYCL backend. Currently, accepted values are: "cuda",
             "hip", "opencl", "level_zero", or "all".
             Default: ``dpctl.backend_type.all``.
-        device_type (optional):
+        device_type (str, :class:`dpctl.device_type`, optional):
             A :class:`dpctl.device_type` enum value or a string that
             specifies a SYCL device type. Currently, accepted values are:
             "gpu", "cpu", "accelerator", or "all".
@@ -210,12 +210,12 @@ cpdef int get_num_devices(
     :class:`dpctl.device_type` and :class:`dpctl.backend_type`.
 
     Args:
-        backend (optional):
+        backend (str, :class:`dpctl.backend_type`, optional):
             A :class:`dpctl.backend_type` enum value or a string that
             specifies a SYCL backend. Currently, accepted values are: "cuda",
             "hip", "opencl", "level_zero", or "all".
             Default: ``dpctl.backend_type.all``.
-        device_type (optional):
+        device_type (str, :class:`dpctl.device_type`, optional):
             A :class:`dpctl.device_type` enum value or a string that
             specifies a SYCL device type. Currently, accepted values are:
             "gpu", "cpu", "accelerator", or "all".
--- a/dpctl/_sycl_device_factory.pyx
+++ b/dpctl/_sycl_device_factory.pyx
@@ -162,7 +162,7 @@ cpdef list get_devices(backend=backend_type.all, device_type=device_type_t.all):
         device_type (optional):
             A :class:`dpctl.device_type` enum value or a string that
             specifies a SYCL device type. Currently, accepted values are:
-            "gpu", "cpu", "accelerator", "host", or "all".
+            "gpu", "cpu", "accelerator", or "all".
             Default: ``dpctl.device_type.all``.
     Returns:
         list:
@@ -218,7 +218,7 @@ cpdef int get_num_devices(
         device_type (optional):
             A :class:`dpctl.device_type` enum value or a string that
             specifies a SYCL device type. Currently, accepted values are:
-            "gpu", "cpu", "accelerator", "host", or "all".
+            "gpu", "cpu", "accelerator", or "all".
             Default: ``dpctl.device_type.all``.
     Returns:
         int:
--- a/dpctl/_sycl_platform.pyx
+++ b/dpctl/_sycl_platform.pyx
@@ -26,6 +26,10 @@ from libcpp cimport bool
 from ._backend cimport (  # noqa: E211
     DPCTLCString_Delete,
     DPCTLDeviceSelector_Delete,
+    DPCTLDeviceVector_Delete,
+    DPCTLDeviceVector_GetAt,
+    DPCTLDeviceVector_Size,
+    DPCTLDeviceVectorRef,
     DPCTLFilterSelector_Create,
     DPCTLPlatform_AreEq,
     DPCTLPlatform_Copy,
@@ -34,6 +38,7 @@ from ._backend cimport (  # noqa: E211
     DPCTLPlatform_Delete,
     DPCTLPlatform_GetBackend,
     DPCTLPlatform_GetDefaultContext,
+    DPCTLPlatform_GetDevices,
     DPCTLPlatform_GetName,
     DPCTLPlatform_GetPlatforms,
     DPCTLPlatform_GetVendor,
@@ -46,17 +51,21 @@ from ._backend cimport (  # noqa: E211
     DPCTLPlatformVector_Size,
     DPCTLPlatformVectorRef,
     DPCTLSyclContextRef,
+    DPCTLSyclDeviceRef,
     DPCTLSyclDeviceSelectorRef,
     DPCTLSyclPlatformRef,
     _backend_type,
+    _device_type,
 )
 
 import warnings
 
 from ._sycl_context import SyclContextCreationError
 from .enum_types import backend_type
+from .enum_types import device_type as device_type_t
 
 from ._sycl_context cimport SyclContext
+from ._sycl_device cimport SyclDevice
 
 __all__ = [
     "get_platforms",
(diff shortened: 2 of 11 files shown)


ndgrigorian:
@oleksandr-pavlyk 
For some context, this PR is setting the stage to support [composite devices](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_composite_device.asciidoc)

It occurred to me that `get_devices` itself is not supported by SyclPlatform, leading to me noticing this long-standing issue.

libsyclinterface/source/dpctl_sycl_platform_interface.cpp
```diff
@@ -269,3 +270,46 @@ size_t DPCTLPlatform_Hash(__dpctl_keep const DPCTLSyclPlatformRef PRef)
...
+    try {
+        DevicesVectorPtr = new vecTy();
+    } catch (std::exception const &e) {
+        delete DevicesVectorPtr;
+        error_handler(e, __FILE__, __func__, __LINE__);
+        return nullptr;
+    }
+
+    // handle unknown device
+    if (!DTy) {
```
oleksandr-pavlyk, line 297: If this is a proxy test for `DTy == DPCTLSyclDeviceType::DPCTL_UNKNOWN_DEVICE`, I'd suggest to be explicit instrad of relying of this enum to map to zero integral value

dpctl/tests/test_sycl_device_factory.py
```diff
@@ -51,8 +51,6 @@ def string_to_device_type(dty_str):
         return dty.accelerator
     elif dty_str == "cpu":
         return dty.cpu
```
antonwolfy, line 53: The `"host"` device type is still present in `docs/_legacy/docfiles/user_guides/manual/dpctl/device_selection.rst`. Should it be cleaned up there?

Also, I note that `device_selection.rst` page is not available on `https://intelpython.github.io/dpctl/latest/index.html`. Is that expected?

dpctl/_sycl_platform.pyx
```diff
@@ -366,6 +375,78 @@ cdef class SyclPlatform(_SyclPlatform):
...
+        """
+        cdef _device_type DTy = _device_type._ALL_DEVICES
+        cdef DPCTLDeviceVectorRef DVRef = NULL
+        cdef size_t num_devs
+        cdef size_t i
+        cdef DPCTLSyclDeviceRef DRef
+
+        if isinstance(device_type, str):
+            dty_str = device_type.strip().lower()
+            if dty_str == "accelerator":
```
antonwolfy, line 412: It seems `"host"` device type is present in the [sycl-2020](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#appendix.device.descriptors), but cleaned up from dpctl and not considered anymore. Is that expected? Should it be clarified in the PR description?

dpctl/_sycl_platform.pyx
```diff
@@ -366,6 +375,78 @@ cdef class SyclPlatform(_SyclPlatform):
         """
         return DPCTLPlatform_Hash(self._platform_ref)
 
+    def get_devices(self, device_type=device_type_t.all):
+        """
+        Returns the list of :class:`dpctl.SyclDevice` objects associated with
+        :class:`dpctl.SyclPlatform` instance selected based on
+        the given :class:`dpctl.device_type`.
```
antonwolfy, line 382: A link to `device_type` is not available in the documentation.

dpctl/_sycl_platform.pyx
```diff
@@ -366,6 +375,78 @@ cdef class SyclPlatform(_SyclPlatform):
...
+                Default: ``dpctl.device_type.all``.
+
+        Returns:
+            list:
+                A :obj:`list` of :class:`dpctl.SyclDevice` objects
+                that belong to this platform.
+
+        Raises:
+            TypeError:
+                If `device_type` is not a str or :class:`dpctl.device_type`
```
antonwolfy, line 398:
```suggestion
                If `device_type` is not a string or :class:`dpctl.device_type`
```

dpctl/_sycl_platform.pyx
```diff
@@ -366,6 +375,78 @@ cdef class SyclPlatform(_SyclPlatform):
...
+                "``enum_types.device_type``."
+            )
+        DVRef = DPCTLPlatform_GetDevices(self.get_platform_ref(), DTy)
+        if (DVRef is NULL):
+            raise ValueError("Internal error: NULL device vector encountered")
+        num_devs = DPCTLDeviceVector_Size(DVRef)
+        devices = []
+        for i in range(num_devs):
+            DRef = DPCTLDeviceVector_GetAt(DVRef, i)
+            devices.append(SyclDevice._create(DRef))
```
antonwolfy, line 445: It seems exactly the same code present in many places, somewhere it is declared as internal function:
```cython
cdef list _get_devices(DPCTLDeviceVectorRef DVRef):
    cdef list devices = []
    cdef size_t nelems = 0
    if DVRef:
        nelems = DPCTLDeviceVector_Size(DVRef)
        for i in range(0, nelems):
            DRef = DPCTLDeviceVector_GetAt(DVRef, i)
            D = SyclDevice._create(DRef)
            devices.append(D)

    return devices
```
Should it be exposed as internal helper function and reused in all that places?

dpctl/tests/test_sycl_platform.py
```diff
@@ -212,3 +213,49 @@ def test_get_platforms():
...
         pytest.fail("Encountered an exception inside get_platforms().")
+
+
+def test_platform_get_devices():
+    platforms = dpctl.get_platforms()
+    if platforms:
+        for p in platforms:
+            assert len(p.get_devices())
+    else:
+        pytest.skip("No platforms available")
```
antonwolfy, line 224: It seems there is no need for skip here, we can just exit silently.

libsyclinterface/source/dpctl_sycl_platform_interface.cpp
```diff
@@ -269,3 +270,46 @@ size_t DPCTLPlatform_Hash(__dpctl_keep const DPCTLSyclPlatformRef PRef)
...
+                      __FILE__, __func__, __LINE__);
+        return nullptr;
+    }
+
+    using vecTy = std::vector<DPCTLSyclDeviceRef>;
+    vecTy *DevicesVectorPtr = nullptr;
+    try {
+        DevicesVectorPtr = new vecTy();
+    } catch (std::exception const &e) {
+        delete DevicesVectorPtr;
```
antonwolfy, line 291: Is it safe to delete here? I might be `nullptr` here, isn't?

dpctl/tests/test_sycl_device_factory.py
ndgrigorian: Both cases are expected and fine. The `_legacy/*` files are the old documentation, from prior to the overhaul in https://github.com/IntelPython/dpctl/pull/1619

dpctl/_sycl_platform.pyx
ndgrigorian: It should be clarified. `"host"` device type is no longer present in SYCL 2020's newer revisions, and was deprecated some time ago.

dpctl/tests/test_sycl_platform.py
ndgrigorian: We have generally preferred to skip in such cases, simply to give the user the message. I think I prefer it this way as well.

libsyclinterface/source/dpctl_sycl_platform_interface.cpp
ndgrigorian: Delete on `nullptr` is a safe operation in C++, so I believe it should be safe here.

[Per C++ documentation](https://en.cppreference.com/w/cpp/language/delete)
> If ptr is a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's unspecified), but the default deallocation functions are guaranteed to do nothing when passed a null pointer.

dpctl/_sycl_platform.pyx
ndgrigorian: Yes, we've reused this a few times, so I agree with that. Perhaps a `_common.pyx` or `_helper.pyx` is a sensible inclusion.

This may be better saved for a follow-up PR though.

dpctl/_sycl_platform.pyx
ndgrigorian: I've updated the PR description to mention this, as well as reference https://github.com/IntelPython/dpctl/pull/1028, where host devices were first removed.

dpctl/_sycl_platform.pyx
ndgrigorian: It appears the enum types defined in `enum_types.py` are not documented anywhere. They will need a page similar to the constants page for `dpctl.tensor`

Good catch!

dpctl/_sycl_platform.pyx
ndgrigorian: It appears the enum types defined in `enum_types.py` are not documented anywhere. They will need a page similar to the constants page for `dpctl.tensor`

Good catch!

ndgrigorian: I will add that in a separate PR, for the sake of not over-loading this one

libsyclinterface/source/dpctl_sycl_platform_interface.cpp
```diff
@@ -269,3 +270,46 @@ size_t DPCTLPlatform_Hash(__dpctl_keep const DPCTLSyclPlatformRef PRef)
...
+        error_handler(e, __FILE__, __func__, __LINE__);
+        return nullptr;
+    }
+
+    // handle unknown device
```
antonwolfy, line 296: I believe it might be helpful to place a comment somewhere (and/or cover in tests) that `automatic` and `custom` device type are handling as `unknown`.

antonwolfy (approved):
@ndgrigorian, I have a minor comment, but in overall LGTM

libsyclinterface/source/dpctl_sycl_platform_interface.cpp
ndgrigorian: That's a good idea

ndgrigorian: I've added a comment mentioning this

dpctl/_sycl_platform.pyx
```diff
@@ -366,6 +375,79 @@ cdef class SyclPlatform(_SyclPlatform):
...
         return DPCTLPlatform_Hash(self._platform_ref)
 
+    def get_devices(self, device_type=device_type_t.all):
+        """
+        Returns the list of :class:`dpctl.SyclDevice` objects associated with
+        :class:`dpctl.SyclPlatform` instance selected based on
+        the given :class:`dpctl.device_type`.
+
+        Args:
+            device_type (str, :class:`dpctl.device_type`):
```
antonwolfy, line 385: Shouldn't the optional be kept here?
```suggestion
            device_type (str, :class:`dpctl.device_type`, optional):
```

dpctl/_sycl_device_factory.pyx
```diff
@@ -162,7 +162,7 @@ cpdef list get_devices(backend=backend_type.all, device_type=device_type_t.all):
         device_type (optional):
```
antonwolfy, line 162: It seems the docstring of `device_type` needs to updated also, like [here](https://github.com/IntelPython/dpctl/pull/1992/files#diff-83d5f19556e2bf2011ee51924c99979156bacd405cfb4351bfbb39bb661a42bcR385)
```suggestion
        device_type (str, :class:`dpctl.device_type`, optional):
```

dpctl/_sycl_device_factory.pyx
```diff
@@ -218,7 +218,7 @@ cpdef int get_num_devices(
...
```
antonwolfy, line 218:
```suggestion
        device_type (str, :class:`dpctl.device_type`, optional):
```

antonwolfy (approved):
No more comments from me. Thank you @ndgrigorian

PR #1993, opened 2025-02-14
Support for composite devices

ndgrigorian:
This PR proposes supporting the [oneAPI composite devices extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_composite_device.asciidoc) in dpctl, which exposes multi-tile GPU devices as root devices for each tile, but also provides access to the composite devices from level zero prior to `ZE_FLAT_DEVICE_HIERARCHY=FLAT` becoming the default. i.e., with `ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE`, the composite devices are root devices, but with `ZE_FLAT_DEVICE_HIERARCHY=COMBINED`, a user gains access to these composite devices through dedicated APIs.


ndgrigorian:
@antonwolfy 
ping

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -849,3 +849,56 @@ DPCTLDevice_GetSubGroupSizes(__dpctl_keep const DPCTLSyclDeviceRef DRef,
...
+{
+    using vecTy = std::vector<DPCTLSyclDeviceRef>;
+    vecTy *ComponentDevicesVectorPtr = nullptr;
+    if (DRef) {
+        auto D = unwrap<device>(DRef);
+        try {
+            auto componentDevices =
+                D->get_info<sycl::ext::oneapi::experimental::info::device::
+                                component_devices>();
+            ComponentDevicesVectorPtr = new vecTy();
```
antonwolfy, line 864: Would it make sense to reserve the number of elements in the vector?
```suggestion
            ComponentDevicesVectorPtr = new vecTy(componentDevices.size());
```

dpctl/__init__.py
```diff
@@ -35,6 +35,7 @@
     SyclSubDeviceCreationError,
 )
 from ._sycl_device_factory import (
+    get_composite_devices,
```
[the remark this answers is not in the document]
ndgrigorian: No, since that's the old documentation anyway

dpctl/_sycl_device.pyx
```diff
@@ -1728,6 +1756,41 @@ cdef class SyclDevice(_SyclDevice):
...
+
+        The returned list will be empty if this SYCL device is not a composite
+        device, i.e., if `is_composite` is ``False``.
+
+        Returns:
+            List[:class:`dpctl.SyclDevice`]:
+                List of component devices.
+
+        Raises:
+            dpctl.SyclSubdeviceCreationError:
```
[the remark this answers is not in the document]
ndgrigorian: > Is there any test that cover raising `SyclSubDeviceCreationError` for `composite_device()` call?

`SyclSubDeviceCreationError` is a typo in and of itself, it isn't raised here

dpctl/_sycl_platform.pyx
```diff
@@ -448,6 +449,41 @@ cdef class SyclPlatform(_SyclPlatform):
...
+
+        Returns:
+            list:
+                A :obj:`list` of composite :class:`dpctl.SyclDevice` objects
+                that belong to this platform.
+
+        Raises:
+            TypeError:
+                If `device_type` is not a str or :class:`dpctl.device_type`
+                enum.
```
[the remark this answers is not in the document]
ndgrigorian: good catch, it's a mistake from copying over the text from `get_devices`

dpctl/_sycl_device_factory.pyx
```diff
@@ -202,6 +204,32 @@ cpdef list get_devices(backend=backend_type.all, device_type=device_type_t.all):
...
+            A list of available composite :class:`dpctl.SyclDevice` instances.
+    """
+    cdef DPCTLDeviceVectorRef DVRef = NULL
+    cdef list composite_devices
+
+    DVRef = DPCTLDeviceMgr_GetCompositeDevices()
+    composite_devices = _get_devices(DVRef)
+    DPCTLDeviceVector_Delete(DVRef)
+
+    return composite_devices
```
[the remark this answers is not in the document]
ndgrigorian: There is no deletion called in `_get_devices` here, which is why it's deleted here and in `get_devices`

libsyclinterface/source/dpctl_sycl_device_interface.cpp
ndgrigorian: Used a call to reserve afterward, to avoid default-initialization of `DPCTLSyclDeviceRef`

dpctl/_sycl_device_factory.pyx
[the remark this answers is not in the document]
ndgrigorian: I see the confusion.

`_get_devices` in `_sycl_device.pyx` is a different function, which does consume `DVRef`

`_get_devices` in `_sycl_device_factory.pyx` does not

Now everywhere `_get_devices` is used, the deletion is called right after... and these are more or less the same function. So maybe this should be factored out at some point into `.pxi`.

dpctl/_sycl_device.pyx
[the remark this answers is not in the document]
antonwolfy: Docstring of other function needs to be update also, since there is no `SyclSubdeviceCreationError` but `SyclSubDeviceCreationError` raised (like in `create_sub_devices_equally`).

antonwolfy (approved):
Thank you @ndgrigorian, LGTM!

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -849,3 +849,57 @@ DPCTLDevice_GetSubGroupSizes(__dpctl_keep const DPCTLSyclDeviceRef DRef,
     }
     return sizes;
 }
+
+__dpctl_give DPCTLDeviceVectorRef
+DPCTLDevice_GetComponentDevices(__dpctl_keep const DPCTLSyclDeviceRef DRef)
+{
+    using vecTy = std::vector<DPCTLSyclDeviceRef>;
+    vecTy *ComponentDevicesVectorPtr = nullptr;
```
AlexanderKalistratov, line 857: You can use `unique_ptr` here to avoid calling `delete` by hands. 

```C++
auto ComponentDevicesVectorPtr = std::unique_ptr<vecTy>();
if (DRef) {
    ...
    try {
        ...
        ComponentDevicesVectorPtr.reset(new VecTy());
        ...
    }
    catch(...) {
            error_handler(e, __FILE__, __func__, __LINE__);
            return nullptr;
    }


return wrap<vecTy>(ComponentDevicesVectorPtr.release());

File: libsyclinterface/source/dpctl_sycl_device_interface.cpp [hunk 7]
```diff
@@ -849,3 +849,57 @@ DPCTLDevice_GetSubGroupSizes(__dpctl_keep const DPCTLSyclDeviceRef DRef,
...
+            return nullptr;
+        }
+    }
+    return wrap<vecTy>(ComponentDevicesVectorPtr);
+}
+
+__dpctl_give DPCTLSyclDeviceRef
+DPCTLDevice_GetCompositeDevice(__dpctl_keep const DPCTLSyclDeviceRef DRef)
+{
+    auto D = unwrap<device>(DRef);
```
AlexanderKalistratov, line 882: This function feels a bit overcomplicated. Is this done for purpose?

```C++
__dpctl_give DPCTLSyclDeviceRef
DPCTLDevice_GetCompositeDevice(__dpctl_keep const DPCTLSyclDeviceRef DRef)
{
    auto D = unwrap<device>(DRef);
    if (D) {
        try {
            bool is_component = D->has(sycl::aspect::ext_oneapi_is_component);
            if (is_component) {
                const auto &compositeDevice =
                D->get_info<sycl::ext::oneapi::experimental::info::device::
                                composite_device>();

                return wrap<device>(new device(compositeDevice));
            }
        } catch (std::exception const &e) {
            error_handler(e, __FILE__, __func__, __LINE__);
        }
    }

    return nullptr;
}
```

libsyclinterface/source/dpctl_sycl_platform_interface.cpp
```diff
@@ -316,3 +316,39 @@ DPCTLPlatform_GetDevices(__dpctl_keep const DPCTLSyclPlatformRef PRef,
         return nullptr;
     }
 }
+
+__dpctl_give DPCTLDeviceVectorRef
+DPCTLPlatform_GetCompositeDevices(__dpctl_keep const DPCTLSyclPlatformRef PRef)
```
AlexanderKalistratov, line 321: Same here:

```C++
__dpctl_give DPCTLDeviceVectorRef
DPCTLPlatform_GetCompositeDevices(__dpctl_keep const DPCTLSyclPlatformRef PRef)
{
    if (!PRef) {
        error_handler("Cannot retrieve composite devices from "
                      "DPCTLSyclPlatformRef as input is a nullptr.",
                      __FILE__, __func__, __LINE__);
        return nullptr;
    }

    auto P = unwrap<platform>(PRef);

    using vecTy = std::vector<DPCTLSyclDeviceRef>;
    try {
        auto DevicesVectorPtr = std::unique_ptr<>(new vecTy());

        auto composite_devices = P->ext_oneapi_get_composite_devices();
        DevicesVectorPtr->reserve(composite_devices.size());
        for (const auto &Dev : composite_devices) {
            DevicesVectorPtr->emplace_back(
                wrap<device>(new device(std::move(Dev))));
        }

        return wrap<vecTy>(DevicesVectorPtr.release());
    } catch (std::exception const &e) {
        error_handler(e, __FILE__, __func__, __LINE__);
    }


    return nullptr;
}
```

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -849,3 +849,57 @@ DPCTLDevice_GetSubGroupSizes(__dpctl_keep const DPCTLSyclDeviceRef DRef,
     }
     return sizes;
 }
+
+__dpctl_give DPCTLDeviceVectorRef
+DPCTLDevice_GetComponentDevices(__dpctl_keep const DPCTLSyclDeviceRef DRef)
+{
+    using vecTy = std::vector<DPCTLSyclDeviceRef>;
+    vecTy *ComponentDevicesVectorPtr = nullptr;
+    if (DRef) {
```
AlexanderKalistratov, line 858: In some functions we are checking for `XRef` to be non `nullptr`. (https://github.com/IntelPython/dpctl/pull/1993/files#diff-93f8a41da82dd5dd2ae6b05e019d8104c87c244e6077d5a819625a2fc5e3a2f1R858)

In others we are checking that result of `auto X = unwrap<...>(XRef)` is non `nullptr`. (https://github.com/IntelPython/dpctl/pull/1993/files#diff-93f8a41da82dd5dd2ae6b05e019d8104c87c244e6077d5a819625a2fc5e3a2f1R883, https://github.com/IntelPython/dpctl/pull/1993/files#diff-2cee49fe41f858d191076f530e6c6ce6ca841be844b8bca525509ed3deddbbaeR324)

Is that intentionally?
Are these checks interchangeable?
Can we `unwrap<...>(XRef)` if `XRef` is nullptr?
Can result of `unwrap<...>(XRef)` be nullptr if `XRef` is not nullptr?

AlexanderKalistratov (commented):
A bit late, but I have some comments:
1. try to avoid calling `delete` by hands and use `unique_ptr` instead
2. In most of the functions two `try...catch` blocks could be combined into one.
3. I have question about equivalence of two checks:
```C++
if (XRef) {
    auto X = unwrap<...>(XRef);
    ...
}
```
vs
```C++
auto X = unwrap<...>(XRef);
if (X) {
    ...
}
```

libsyclinterface/source/dpctl_sycl_device_interface.cpp
ndgrigorian: This is something that might be smart to go back and do throughout the libsyclinterface codebase

libsyclinterface/source/dpctl_sycl_device_interface.cpp
ndgrigorian: They are treated as more or less interchangeable throughout the file.

As to if we can `unwrap` a nullptr, yes, there are a few tests which more or less do this throughout the libsyclinterface tests.

And the last question, I can't think of such a case. Given it's treated as interchangeable: probably not.

ndgrigorian:
> A bit late, but I have some comments:
> 
> 1. try to avoid calling `delete` by hands and use `unique_ptr` instead
> 2. In most of the functions two `try...catch` blocks could be combined into one.
> 3. I have question about equivalence of two checks:
> 
> ```c++
> if (XRef) {
>     auto X = unwrap<...>(XRef);
>     ...
> }
> ```
> 
> vs
> 
> ```c++
> auto X = unwrap<...>(XRef);
> if (X) {
>     ...
> }
> ```

Suggestions themselves make sense, I think it may be most sensible to open up (a) PR(s) doing some refactoring and clean-up in libsyclinterface

PR #2003, opened 2025-02-20
Fix possible `itemsize` overflows in `usm_ndarray` from-pointer constructors

ndgrigorian:
C-API functions which make a `usm_ndarray` from a pointer (`UsmNDArray_MakeSimpleFromPtr` and `UsmNDArray_MakeFromPtr`) take an integral `typenum` argument

If this argument is invalid, the `itemsize` declared in the functions would be returned from `type_bytesize` as -1 and could potentially overflow

### DIFF
--- a/dpctl/tensor/_usmarray.pyx
+++ b/dpctl/tensor/_usmarray.pyx
@@ -1777,8 +1777,10 @@ cdef api object UsmNDArray_MakeSimpleFromPtr(
     Returns:
         Created usm_ndarray instance
     """
-    cdef size_t itemsize = type_bytesize(typenum)
-    cdef size_t nbytes = itemsize * nelems
+    cdef int itemsize = type_bytesize(typenum)
+    if (itemsize < 1):
+        raise ValueError("dtype with typenum=" + str(typenum) + " is not supported.")
+    cdef size_t nbytes = (<size_t> itemsize) * nelems
     cdef c_dpmem._Memory mobj = c_dpmem._Memory.create_from_usm_pointer_size_qref(
         ptr, nbytes, QRef, memory_owner=owner
     )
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_usmarray.pyx
```diff
@@ -1777,8 +1777,10 @@ cdef api object UsmNDArray_MakeSimpleFromPtr(
     Returns:
         Created usm_ndarray instance
     """
-    cdef size_t itemsize = type_bytesize(typenum)
-    cdef size_t nbytes = itemsize * nelems
+    cdef int itemsize = type_bytesize(typenum)
+    if (itemsize < 1):
+        raise ValueError("dtype with typenum=" + str(typenum) + " is not supported.")
```
antonwolfy, line 1782: As I can see there is `test_pyx_capi_make_from_ptr` test to cover `UsmNDArray_MakeSimpleFromPtr` call. Should it be extended or the new one added to cover that negative scenario?

dpctl/tensor/_usmarray.pyx
```diff
@@ -1830,6 +1832,8 @@ cdef api object UsmNDArray_MakeFromPtr(
     cdef object obj_shape
     cdef object obj_strides
 
+    if (itemsize < 1):
+        raise ValueError("dtype with typenum=" + str(typenum) + " is not supported.")
```
antonwolfy, line 1836: And also the same about `UsmNDArray_MakeFromPtr`.

dpctl/tensor/_usmarray.pyx
ndgrigorian: I've added a test

antonwolfy (approved):
LGTM! Thank you @ndgrigorian

Issue #2017, opened 2025-02-24
`dpt.cumulative_sum` returns incorrect result for `dtype=dpt.bool`


ndgrigorian:
Similar to previously closed issue for reductions #1944 

```python
import dpctl.tensor as dpt
a = dpt.asarray([-1, 1], dtype=dpt.int32)
dpt.cumulative_sum(a)
# usm_ndarray([-1, 0])
dpt.cumulative_sum(a, dtype=dpt.bool)
# usm_ndarray([True, False])

# expected
dpt.cumulative_sum(dpt.astype(a, dpt.bool), dtype=dpt.bool)
# usm_ndarray([True, True])
```

PR #2018, opened 2025-02-25
Fix incorrect results from `cumulative_sum` and `cumulative_prod` with `dtype=bool`

ndgrigorian:
This PR resolves gh-2017 by introducing missing dedicated kernels for `bool->bool` in the type matrix of `cumulative_sum` and `cumulative_prod`.

Also slips in the addition of `int8->int8`

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
@@ -86,6 +86,7 @@ template <typename srcTy, typename dstTy> struct CastTransformer
 
 template <typename ScanOpT, typename T> struct needs_workaround
 {
+    // work-around needed due to crash in JITing on CPU
     static constexpr bool value =
         std::is_same_v<ScanOpT, sycl::logical_or<T>> ||
         std::is_same_v<ScanOpT, sycl::logical_and<T>>;
--- a/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
@@ -88,7 +88,8 @@ template <typename srcTy, typename dstTy> struct CastTransformer
 
 template <typename ScanOpT, typename T> struct needs_workaround
 {
-    // work-around needed due to crash in JITing on CPU
+    // workaround needed due to crash in JITing on CPU
+    // remove when CMPLRLLVM-65813 is resolved
     static constexpr bool value = su_ns::IsSyclLogicalAnd<T, ScanOpT>::value ||
                                   su_ns::IsSyclLogicalOr<T, ScanOpT>::value;
 };
--- a/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/accumulators.hpp
@@ -84,9 +84,17 @@ template <typename srcTy, typename dstTy> struct CastTransformer
     }
 };
 
+template <typename ScanOpT, typename T> struct needs_workaround
+{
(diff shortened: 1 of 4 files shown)


dpctl/tensor/libtensor/include/kernels/accumulators.hpp
```diff
@@ -84,9 +84,18 @@ template <typename srcTy, typename dstTy> struct CastTransformer
     }
 };
 
+template <typename ScanOpT, typename T> struct needs_workaround
+{
+    // work-around needed due to crash in JITing on CPU
```
antonwolfy, line 89: Would it be helpful to post a link on any issue describing that?

dpctl/tensor/libtensor/include/kernels/accumulators.hpp
```diff
@@ -84,9 +84,18 @@ template <typename srcTy, typename dstTy> struct CastTransformer
...
+    static constexpr bool value =
+        std::is_same_v<ScanOpT, sycl::logical_or<T>> ||
```
antonwolfy, line 91: There are `IsSyclLogicalOr` and `IsSyclLogicalAnd` declared in `dpctl/tensor/libtensor/include/utils/sycl_utils.hpp`.
Should they be used here?

dpctl/tensor/libtensor/source/accumulators/cumulative_prod.cpp
```diff
@@ -138,7 +140,9 @@ struct CumProd1DContigFactory
         if constexpr (TypePairSupportDataForProdAccumulation<srcTy,
                                                              dstTy>::is_defined)
         {
-            using ScanOpT = sycl::multiplies<dstTy>;
+            using ScanOpT = std::conditional_t<std::is_same_v<dstTy, bool>,
+                                               sycl::logical_and<dstTy>,
+                                               sycl::multiplies<dstTy>>;
```
antonwolfy, line 145: It seems the same per every factory.
Should we define an alias for
```cpp
std::conditional_t<std::is_same_v<dstTy, bool>,
                                               sycl::logical_and<dstTy>,
                                               sycl::multiplies<dstTy>>
```
and to reuse in every factory to reduce duplication?

dpctl/tensor/libtensor/source/accumulators/cumulative_prod.cpp
```diff
@@ -171,7 +175,9 @@ struct CumProd1DIncludeInitialContigFactory
...
```
antonwolfy, line 180: Also here is `IsSyclMultiplies` declared in `dpctl/tensor/libtensor/include/utils/sycl_utils.hpp`. Should it be reused? Or is that intended only for kernel code?

ndgrigorian: It isn't obvious to me how it's useful here, for these conditional checks you mean?

dpctl/tensor/libtensor/source/accumulators/cumulative_prod.cpp
ndgrigorian: Added this change, good suggestion

dpctl/tensor/libtensor/source/accumulators/cumulative_prod.cpp
antonwolfy: I misread the code, please ignore the comment.

antonwolfy (dismissed):
Looks good for me, thank you @ndgrigorian

antonwolfy (approved):
LGTM! Thank you @ndgrigorian

Issue #2022, opened 2025-03-04
Level Zero Loader detection does not work with most recent DPC++ nightly


sommerlukas:
https://github.com/intel/llvm/pull/17104 changed the way the Level Zero library is linked into the DPC++/UR Level Zero adapter. 

This means that the logic to detect the name of the Level Zero loader library [here](https://github.com/IntelPython/dpctl/blob/master/libsyclinterface/CMakeLists.txt#L69-L76), which relies on Level Zero being linked dynamically to the adapter and showing up in `readelf -d` will stop working.

The first nightly of DPC++ with which this behavior can be triggered is the [nightly from 2025-03-04](https://github.com/intel/llvm/releases/download/nightly-2025-03-04/sycl_linux.tar.gz).

PR #2025, opened 2025-03-18
Update Level Zero loader detection for newer compiler builds

ndgrigorian:
As of https://github.com/intel/llvm/pull/17104 `libur_adapter_level_zero.so` is statically linked to the Level Zero loader, meaning that the shared object can no longer be read for the loader filename

To resolve this, a default filename is now set in CMake, as it is unlikely to change. For backwards compatibility, it is only set to the default value if the object file is read and the loader is not found through the original approach.

Resolves https://github.com/IntelPython/dpctl/issues/2022


ndgrigorian:
@sommerlukas 
This should resolve the need for a work-around

The only thing left is to make sure it works on Windows with nightly compiler

libsyclinterface/CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ option(DPCTL_ENABLE_GLOG
 
 set(DPCTL_LEVEL_ZERO_INCLUDE_DIR "" CACHE STRING "Level zero header path")
 
+set(LIBZE_DEFAULT_LOADER_FILENAME "libze_loader.so.1" CACHE STRING "Default Level Zero Loader filename")
```
sommerlukas, line 47: I'd recommend to guard setting this variable with an `if` and only set it if it isn't already set by the user, like I did [here](https://github.com/sommerlukas/dpctl/commit/c90bc708b7af3b46298de64d9e6d7ac189c4c589). 

This allows users to override the name through CMake options supplied to the build.

sommerlukas (commented):
Solution looks good, we could additionally allow users to set the variable.

libsyclinterface/CMakeLists.txt
ndgrigorian: I considered a similar approach. I think it would be best to add another variable for the user to set instead, because then we can also guard the `readelf` process, since if the user is setting it directly, there's no reason to attempt to inspect for the filename.

ndgrigorian: Added `LIBZE_SET_LOADER_FILENAME` for this purpose

sommerlukas (approved):

libsyclinterface/CMakeLists.txt
```diff
@@ -60,20 +64,28 @@ if(DPCTL_ENABLE_L0_PROGRAM_CREATION)
...
+                ERROR_STRIP_TRAILING_WHITESPACE
+            )
+            # if libze_loader is statically linked, LIBZE_LOADER_FILENAME will be an empty string
+            if ("x${LIBZE_LOADER_FILENAME}" STREQUAL "x")
+                set(LIBZE_LOADER_FILENAME ${LIBZE_DEFAULT_LOADER_FILENAME})
+            endif()
+        else ()
+            set(LIBZE_LOADER_FILENAME ${LIBZE_SET_LOADER_FILENAME})
+        endif()
         set(LIBZE_LOADER_FILENAME ${LIBZE_LOADER_FILENAME} PARENT_SCOPE)
```
antonwolfy, line 89: Should we remove that line? It seems duplicating with [one](https://github.com/IntelPython/dpctl/pull/2025/files#diff-df76a50a4c90962d8ff51b821a9f862a80054464e8880eda6aa648fa5bcbdc80R91) below.

antonwolfy (dismissed):
Thank you @ndgrigorian, LGTM! Just a small nit.

libsyclinterface/CMakeLists.txt
ndgrigorian: Difficult to say, it does duplicate, but it was already present in the code before, so I would want to be sure that it never served a purpose

antonwolfy: I don't mind to handle that separately if needed.

antonwolfy (approved):

PR #2032, opened 2025-03-26
Update integer advanced indexing for array API 2024.12 spec

ndgrigorian:
The array API specification now requires advanced integer indexing (i.e., with arrays) to promote Python integer scalars to arrays when mixed with integer arrays.

This PR implements this behavior, promoting any integer indices that occur in a sequence with array indices of dimension 1 or greater. 

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -800,6 +800,79 @@ def _nonzero_impl(ary):
     return res
 
 
+def _validate_indices(inds, queue_list, usm_type_list):
+    """
+    Utility for validating indices are usm_ndarray of integral dtype or Python
+    integers. At least one must be an array.
+
+    For each array, the queue and usm type are appended to `queue_list` and
+    `usm_type_list`, respectively.
+    """
+    any_usmarray = False
+    for ind in inds:
+        if isinstance(ind, dpt.usm_ndarray):
+            any_usmarray = True
+            if ind.dtype.kind not in "ui":
+                raise IndexError(
+                    "arrays used as indices must be of integer (or boolean) "
+                    "type"
+                )
+            queue_list.append(ind.sycl_queue)
+            usm_type_list.append(ind.usm_type)
+        elif not isinstance(ind, Integral):
+            raise TypeError(
+                "all elements of `ind` expected to be usm_ndarrays "
+                f"or integers, found {type(ind)}"
+            )
+    if not any_usmarray:
+        raise TypeError(
+            "at least one element of `inds` expected to be a usm_ndarray"
+        )
+    return inds
+
+
+def _prepare_indices_arrays(inds, q, usm_type):
+    """
+    Utility taking a mix of usm_ndarray and possibly Python int scalar indices,
+    a queue (assumed to be common to arrays in inds), and a usm type.
+
+    Python scalar integers are promoted to arrays on the provided queue and
+    with the provided usm type. All arrays are then promoted to a common
+    integral type (if possible) before being broadcast to a common shape.
+    """
+    # scalar integers -> arrays
+    inds = tuple(
+        map(
+            lambda ind: (
+                ind
+                if isinstance(ind, dpt.usm_ndarray)
+                else dpt.asarray(ind, usm_type=usm_type, sycl_queue=q)
+            ),
+            inds,
+        )
+    )
+
+    # promote to a common integral type if possible
+    ind_dt = dpt.result_type(*inds)
+    if ind_dt.kind not in "ui":
+        raise ValueError(
+            "cannot safely promote indices to an integer data type"
+        )
+    inds = tuple(
+        map(
+            lambda ind: (
+                ind if ind.dtype == ind_dt else dpt.astype(ind, ind_dt)
+            ),
+            inds,
+        )
+    )
+
+    # broadcast
+    inds = dpt.broadcast_arrays(*inds)
+
+    return inds
+
+
 def _take_multi_index(ary, inds, p, mode=0):
     if not isinstance(ary, dpt.usm_ndarray):
         raise TypeError(
@@ -820,26 +893,8 @@ def _take_multi_index(ary, inds, p, mode=0):
     ]
     if not isinstance(inds, (list, tuple)):
         inds = (inds,)
-    any_usmarray = False
-    for ind in inds:
-        if isinstance(ind, dpt.usm_ndarray):
-            any_usmarray = True
-            if ind.dtype.kind not in "ui":
-                raise IndexError(
-                    "arrays used as indices must be of integer (or boolean) "
-                    "type"
-                )
-            queues_.append(ind.sycl_queue)
-            usm_types_.append(ind.usm_type)
-        elif not isinstance(ind, Integral):
-            raise TypeError(
-                "all elements of `ind` expected to be usm_ndarrays "
-                "or integers"
-            )
-    if not any_usmarray:
-        raise TypeError(
-            "at least one element of `ind` expected to be a usm_ndarray"
-        )
+
+    _validate_indices(inds, queues_, usm_types_)
     res_usm_type = dpctl.utils.get_coerced_usm_type(usm_types_)
     exec_q = dpctl.utils.get_execution_queue(queues_)
     if exec_q is None:
@@ -849,34 +904,10 @@ def _take_multi_index(ary, inds, p, mode=0):
             "Use `usm_ndarray.to_device` method to migrate data to "
             "be associated with the same queue."
         )
+
     if len(inds) > 1:
-        inds = tuple(
-            map(
-                lambda ind: (
-                    ind
-                    if isinstance(ind, dpt.usm_ndarray)
-                    else dpt.asarray(
-                        ind, usm_type=res_usm_type, sycl_queue=exec_q
-                    )
-                ),
-                inds,
-            )
-        )
-        ind_dt = dpt.result_type(*inds)
-        # ind arrays have been checked to be of integer dtype
-        if ind_dt.kind not in "ui":
-            raise ValueError(
-                "cannot safely promote indices to an integer data type"
-            )
-        inds = tuple(
-            map(
-                lambda ind: (
-                    ind if ind.dtype == ind_dt else dpt.astype(ind, ind_dt)
-                ),
-                inds,
-            )
-        )
-        inds = dpt.broadcast_arrays(*inds)
+        inds = _prepare_indices_arrays(inds, exec_q, res_usm_type)
+
     ind0 = inds[0]
     ary_sh = ary.shape
     p_end = p + len(inds)
@@ -992,26 +1023,9 @@ def _put_multi_index(ary, inds, p, vals, mode=0):
         ]
     if not isinstance(inds, (list, tuple)):
         inds = (inds,)
-    any_usmarray = False
-    for ind in inds:
-        if isinstance(ind, dpt.usm_ndarray):
-            any_usmarray = True
-            if ind.dtype.kind not in "ui":
-                raise IndexError(
-                    "arrays used as indices must be of integer (or boolean) "
-                    "type"
-                )
-            queues_.append(ind.sycl_queue)
-            usm_types_.append(ind.usm_type)
-        elif not isinstance(ind, Integral):
-            raise TypeError(
-                "all elements of `ind` expected to be usm_ndarrays "
-                "or integers"
-            )
-    if not any_usmarray:
-        raise TypeError(
-            "at least one element of `ind` expected to be a usm_ndarray"
-        )
+
+    _validate_indices(inds, queues_, usm_types_)
+
     vals_usm_type = dpctl.utils.get_coerced_usm_type(usm_types_)
     exec_q = dpctl.utils.get_execution_queue(queues_)
     if exec_q is not None:
@@ -1028,34 +1042,10 @@ def _put_multi_index(ary, inds, p, vals, mode=0):
             "Use `usm_ndarray.to_device` method to migrate data to "
             "be associated with the same queue."
         )
+
     if len(inds) > 1:
-        inds = tuple(
-            map(
-                lambda ind: (
-                    ind
-                    if isinstance(ind, dpt.usm_ndarray)
-                    else dpt.asarray(
-                        ind, usm_type=vals_usm_type, sycl_queue=exec_q
-                    )
-                ),
-                inds,
-            )
-        )
-        ind_dt = dpt.result_type(*inds)
-        # ind arrays have been checked to be of integer dtype
-        if ind_dt.kind not in "ui":
-            raise ValueError(
-                "cannot safely promote indices to an integer data type"
-            )
-        inds = tuple(
-            map(
-                lambda ind: (
-                    ind if ind.dtype == ind_dt else dpt.astype(ind, ind_dt)
-                ),
-                inds,
-            )
-        )
-        inds = dpt.broadcast_arrays(*inds)
+        inds = _prepare_indices_arrays(inds, exec_q, vals_usm_type)
+
     ind0 = inds[0]
     ary_sh = ary.shape
     p_end = p + len(inds)
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -15,6 +15,7 @@
 #  limitations under the License.
 import builtins
 import operator
+from numbers import Integral
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


dpctl/tensor/_slicing.pxi
```diff
@@ -255,26 +258,44 @@ def _basic_slice_meta(ind, shape : tuple, strides : tuple, offset : int):
...
+                    else:
+                        raise IndexError(
+                            ("Index {0} is out of range for "
+                            "axes {1} with size {2}").format(ind_i, k, shape[k]))
             elif isinstance(ind_i, usm_ndarray):
+                if not array_streak:
+                    array_streak = True
                 if not advanced_start_pos_set:
                     new_advanced_start_pos = len(new_shape)
                     advanced_start_pos_set = True
```
antonwolfy, line 301: It seems there is `IndexError` below without any message. Should it be updated?

dpctl/tensor/_usmarray.pyx
```diff
@@ -962,28 +961,30 @@ cdef class usm_ndarray:
...
-            if not matching:
-                raise IndexError("boolean index did not match indexed array in dimensions")
-            res = _extract_impl(res, key_, axis=adv_ind_start_p)
-            res.flags_ = _copy_writable(res.flags_, self.flags_)
-            return res
 
-        if any(ind.dtype == dpt_bool for ind in adv_ind):
+        # if len(adv_ind == 1), the (only) element is always an array
+        if len(adv_ind) == 1 and adv_ind[0].dtype == dpt_bool:
+                key_ = adv_ind[0]
```
antonwolfy, line 967: I guess the code needs to be formatted, too many spaces in the indentation.

dpctl/tensor/_usmarray.pyx
```diff
@@ -962,28 +961,30 @@ cdef class usm_ndarray:
...
+                res.flags_ = _copy_writable(res.flags_, self.flags_)
+                return res
+
+        if any((isinstance(ind, usm_ndarray) and ind.dtype == dpt_bool) for ind in adv_ind):
             adv_ind_int = list()
             for ind in adv_ind:
-                if ind.dtype == dpt_bool:
-                    adv_ind_int.extend(_nonzero_impl(ind))
+                if isinstance(ind, usm_ndarray) and ind.dtype == dpt_bool:
+                        adv_ind_int.extend(_nonzero_impl(ind))
```
antonwolfy, line 987: the same here:
```suggestion
                    adv_ind_int.extend(_nonzero_impl(ind))
```

dpctl/tensor/_usmarray.pyx
```diff
@@ -962,28 +961,30 @@ cdef class usm_ndarray:
             return res
 
         from ._copy_utils import _extract_impl, _nonzero_impl, _take_multi_index
```
antonwolfy, line 963: I noticed some legacy behavior changed:
```python
a = dpt.ones(4)
idx = dpt.asarray(2)

dpt.sum(a[idx:], out=a[idx])
# Out: usm_ndarray(2.)

a
# Out: usm_ndarray([1., 1., 1., 1.])

# in legacy it was:
usm_ndarray([1., 1., 2., 1.])
```
I wonder if it is expected or not?

dpctl/tensor/_copy_utils.py
```diff
@@ -819,15 +820,26 @@ def _take_multi_index(ary, inds, p, mode=0):
...
+            any_usmarray = True
+            if ind.dtype.kind not in "ui":
+                raise IndexError(
+                    "arrays used as indices must be of integer (or boolean) "
+                    "type"
+                )
+            queues_.append(ind.sycl_queue)
+            usm_types_.append(ind.usm_type)
+        elif not isinstance(ind, Integral):
+            raise TypeError(
```
antonwolfy, line 835: Seems not covered by a test

dpctl/tensor/_copy_utils.py
```diff
@@ -819,15 +820,26 @@ def _take_multi_index(ary, inds, p, mode=0):
...
+                "all elements of `ind` expected to be usm_ndarrays "
+                "or integers"
             )
+    if not any_usmarray:
+        raise TypeError(
```
antonwolfy, line 840: And that negative one is not covered

dpctl/tensor/_usmarray.pyx
ndgrigorian: This is interesting and I would expect it given the changes here. Previously, 0D arrays were being treated as Python scalars and brought to the host, meaning that in this case, it would never pass through advanced indexing.

Now this is treated as an advanced indexing case in all circumstances. So `a[idx]` here is a different array.

antonwolfy: It's clear now, thank you.

ndgrigorian: For dpnp's sake, it may be worth noting that in NumPy's case, `a[idx]` for some 0-D integer array `idx` always returns a NumPy scalar, which can't be written to, so effectively, it may as well be a new array.

When `idx` is 1-D and greater, the behaviors match.

dpctl/tensor/_copy_utils.py
```diff
@@ -819,15 +820,26 @@ def _take_multi_index(ary, inds, p, mode=0):
...
+            if ind.dtype.kind not in "ui":
+                raise IndexError(
+                    "arrays used as indices must be of integer (or boolean) "
+                    "type"
+                )
+            queues_.append(ind.sycl_queue)
+            usm_types_.append(ind.usm_type)
+        elif not isinstance(ind, Integral):
+            raise TypeError(
+                "all elements of `ind` expected to be usm_ndarrays "
```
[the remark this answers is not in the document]
ndgrigorian: good idea for the sake of consistency

dpctl/tensor/_slicing.pxi
ndgrigorian: It can be, with message
```
                    "Only integers, slices (`:`), ellipsis (`...`), "
                    "dpctl.tensor.newaxis (`None`) and integer and "
                    "boolean arrays are valid indices."
```
but it was never added because the first validation already checks this condition, it makes more sense (to me) to just remove it.

ndgrigorian:
@antonwolfy 
a couple of new tests added, I think PR is ready for another review pass

dpctl/tensor/_copy_utils.py
```diff
@@ -799,6 +800,79 @@ def _nonzero_impl(ary):
...
+                ind
+                if isinstance(ind, dpt.usm_ndarray)
+                else dpt.asarray(ind, usm_type=usm_type, sycl_queue=q)
+            ),
+            inds,
+        )
+    )
+
+    # promote to a common integral type if possible
+    ind_dt = dpt.result_type(*inds)
```
antonwolfy, line 856: Would it be helpful to comment somewhere that `ind_dt` will never has boolean kind (at least it is not expected by the code logic), because in case of bool dtype of any input `ind` in `inds` list there will be `_nonzero_impl(ind)` call for it done in `__getitem__`?

dpctl/tests/test_usm_ndarray_indexing.py
```diff
@@ -395,6 +402,33 @@ def test_advanced_slice13():
     assert (dpt.asnumpy(y) == dpt.asnumpy(expected)).all()
 
 
+def test_advanced_slice14():
+    q = get_queue_or_skip()
+    ii = dpt.asarray([1, 2], sycl_queue=q)
+    x = dpt.reshape(dpt.arange(3**5, dtype="i4", sycl_queue=q), (3,) * 5)
```
antonwolfy, line 408: Is it expected to have the exception raised in below example?
```python
x = dpt.reshape(dpt.arange(3**5, dtype="i4"), (3,) * 5)
ii = dpt.asarray([1, 2])
i0 = dpt.asarray(0, dtype='?')
i1 = dpt.asarray(1, dtype='?')

x[ii, i0, ii, i1, :]
IndexError                                Traceback (most recent call last)
Cell In[30], line 1
----> 1 x[ii, i0, ii, i1, :]

File /localdisk/work/antonvol/code/dpctl/dpctl/tensor/_usmarray.pyx:936, in dpctl.tensor._usmarray.usm_ndarray.__getitem__()

File /localdisk/work/antonvol/code/dpctl/dpctl/tensor/_slicing.pxi:190, in dpctl.tensor._usmarray._basic_slice_meta()

IndexError: Advanced indexing array specs may not be separated by basic slicing specs.
```
while that works fine
```python
i0 = dpt.asarray(0)
i1 = dpt.asarray(1)

x[ii, i0, ii, i1, :]
# Out:
# usm_ndarray([[ 93,  94,  95],
#              [183, 184, 185]], dtype=int32)
```

dpctl/tests/test_usm_ndarray_indexing.py
```diff
@@ -252,8 +253,14 @@ def test_advanced_slice5():
     q = get_queue_or_skip()
     ii = dpt.asarray([1, 2], sycl_queue=q)
     x = _make_3d("i4", q)
-    with pytest.raises(IndexError):
-        x[ii, 0, ii]
+    y = x[ii, 0, ii]
```
antonwolfy, line 256: Do we need to have any test, where we are testing different combination of dtypes for indices arrays?

dpctl/tests/test_usm_ndarray_indexing.py
antonwolfy: in case of numpy there will be broadcast error raised (which is expected I guess):
```python
x = np.reshape(np.arange(3**5, dtype="i4"), (3,) * 5)
ii = np.asarray([1, 2])
i0 = np.asarray(0, dtype='?')
i1 = np.asarray(1, dtype='?')

x[ii, i0, ii, i1, :]
IndexError                                Traceback (most recent call last)
Cell In[30], line 1
----> 1 x[ii, i0, ii, i1, :]

IndexError: shape mismatch: indexing arrays could not be broadcast together with shapes (2,) (0,) (2,) (1,)
```
but if we update the example, like:
```python
ii = np.asarray(1)

x[ii, i0, ii, i1, :]
# Out: array([], shape=(0, 3, 3, 3), dtype=int32)
```
it will work in numpy, but still failing in dpctl with the same error:
```python
ii = dpt.asarray(1)

x[ii, i0, ii, i1, :]
IndexError                                Traceback (most recent call last)
Cell In[37], line 1
----> 1 x[ii, i0, ii, i1, :]

File /localdisk/work/antonvol/code/dpctl/dpctl/tensor/_usmarray.pyx:936, in dpctl.tensor._usmarray.usm_ndarray.__getitem__()

File /localdisk/work/antonvol/code/dpctl/dpctl/tensor/_slicing.pxi:190, in dpctl.tensor._usmarray._basic_slice_meta()

IndexError: Advanced indexing array specs may not be separated by basic slicing specs.
```

dpctl/tests/test_usm_ndarray_indexing.py
ndgrigorian: Such tests were already added, just only using integer arrays (see: `test_advanced_integer_indexing_cast_indices`). I don't think we need one for Python scalars because `dpt.asarray` is called on them, followed by the same coercion as in the array case.

dpctl/tests/test_usm_ndarray_indexing.py
ndgrigorian: This is a very strange edge case, I guess it's just never been looked at, but in theory, this would have been an issue since gh-1136

Integral arrays are treated as advanced indexing specs, while 0D boolean arrays are (always) treated as Python scalars and are brought back to the host, treated as basic slicing specs. We didn't permit non-consecutive advanced slicing specs because the behavior is ill-defined (NumPy and CuPy make an arbitrary choice with respect to this).

0D boolean arrays are treated as Python scalars because of NumPy's odd behavior, mentioned in gh-1136

ndgrigorian: This gets even uglier and more convoluted than just this though...

```python
# x = dpt.reshape(dpt.arange(3**5, dtype="i4"), (3,) * 5)
# ii = dpt.asarray(1)
# i0 = dpt.asarray(0, dtype='?')
# i1 = dpt.asarray(0, dtype='?')
# x_np, ii_np, i0_np, i1_np are the same arrays after dpt.asnumpy calls
x_np[ii_np, i0_np[np.newaxis], ii_np, i1_np[np.newaxis], :]
IndexError                                Traceback (most recent call last)
Cell In[145], line 1
----> 1 x_np[1, i0_np[np.newaxis], 1, i1_np[np.newaxis], :]

IndexError: boolean index did not match indexed array along axis 1; size of axis is 3 but size of corresponding boolean axis is 1
```
this is clearly a bug in NumPy, but works fine for us
```python
In [146]: x[ii, i0[dpt.newaxis], ii, i1[dpt.newaxis], :]
Out[146]: usm_ndarray([], dtype=int32)

In [147]: _.shape
Out[147]: (0, 3)

In [148]: x_np[ii_np, i0_np[np.newaxis].nonzero(), ii_np, i1_np[np.newaxis].nonzero(), :]
Out[148]: array([], shape=(1, 0, 3), dtype=int32)
```

...except the shape is wrong.

digging a bit further

```python
In [149]: x[1, False, 1, True, :]
Out[149]: usm_ndarray([], dtype=int32)

In [150]: _.shape
Out[150]: (0, 1, 3, 3, 3)

In [151]: x_np[1, False, 1, True, :]
Out[151]: array([], shape=(0, 3, 3, 3), dtype=int32)
```

ndgrigorian: I will go back and verify but I doubt any of these bugs are related to the changes in this PR. They can be fixed in a follow-up PR.

dpctl/tensor/_slicing.pxi
```diff
@@ -179,10 +175,12 @@ def _basic_slice_meta(ind, shape : tuple, strides : tuple, offset : int):
                 if array_streak_started:
                     array_streak_interrupted = True
             elif _is_integral(i):
-                explicit_index += 1
                 axes_referenced += 1
-                if array_streak_started:
-                    array_streak_interrupted = True
+                if array_streak_started and not array_streak_interrupted:
```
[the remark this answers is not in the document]
ndgrigorian: Also, it's identical to NumPy's text
```
In [152]: x_np[1, False, 1, dict(), :]
IndexError                                Traceback (most recent call last)
Cell In[152], line 1
----> 1 x_np[1, False, 1, dict(), :]

IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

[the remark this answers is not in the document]
antonwolfy: No, I meant `IndexError` raised at line#191. To trace ind which brakes the sequence or something similar..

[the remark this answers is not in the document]
ndgrigorian: I see, that makes more sense, but would still require us to add an enumerate the loop to get index of the current indices object

dpctl/tests/test_usm_ndarray_indexing.py
ndgrigorian: I think it would be excessive to add boolean to it, since `nonzero` converts to integer anyway. I think adding a comment/documenting it in `_prepare_indices_arrays` makes a lot of sense though.

dpctl/tensor/_slicing.pxi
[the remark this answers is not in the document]
antonwolfy: I see, then I don't mind to leave that unchanged, it might be improved, but not as a part of that PR.

dpctl/tests/test_usm_ndarray_indexing.py
ndgrigorian: The case
```python
x[ii, i0, ii, i1, :]
```
failing was introduced by this PR, when the change was made that 0D integer arrays would be treated as arrays unconditionally.

This might be fine, but I will look at tweaking this behavior to make it more conditional. Maybe 0D integer arrays can never be considered the _start_ of advanced array indexing, but in the middle of it, will be treated on the same level as arrays.

The shape in this case is still a mismatch with NumPy however. Figuring out the correct behavior can be deferred for now though.

dpctl/tensor/_copy_utils.py
ndgrigorian: I just realized that `_validate_indices` already checks this above, and is called before `_prepare_indices_arrays` is ever called. So in that case, I think it's fine that `_prepare_indices_arrays` doesn't mention anything, as it's an internal function anyway.

dpctl/tests/test_usm_ndarray_indexing.py
ndgrigorian: @antonwolfy 
Partly reverted 87fa69a4f5edbf0c194225c45b82770e87ff319c which introduced the change in behavior

```python

In [2]: x = dpt.reshape(dpt.arange(3**5, dtype="i4"), (3,) * 5)

In [3]: ii, i0, i1 = dpt.asarray(1), dpt.asarray(False), dpt.asarray(True)

In [4]: x[ii, i0, ii, i1, :]
Out[4]: usm_ndarray([], dtype=int32)

In [5]: _.shape
Out[5]: (0, 1, 3, 3, 3)
```
now 0D integral arrays of treated as Python scalars unless they're part of consecutive (integral) arrays

antonwolfy: Would it make sense to add a test for that?

antonwolfy (approved):
Thank you @ndgrigorian, LGTM

PR #2038, opened 2025-04-02
Add support for raw_kernel_arg extension

sommerlukas:
Add support for the DPC++ SYCL extension `raw_kernel_arg` that allows to pass a binary blob as kernel arguments.

This is for example useful if the kernel uses a `struct` as kernel parameter, but is loaded from SPIR-V and the argument type is unknown for `set_arg`.

The implementation follows #1984.


libsyclinterface/include/syclinterface/dpctl_sycl_extension_interface.h
```diff
@@ -53,4 +53,22 @@ void DPCTLWorkGroupMemory_Delete(__dpctl_take DPCTLSyclWorkGroupMemoryRef Ref);
 DPCTL_API
 bool DPCTLWorkGroupMemory_Available();
 
+typedef struct RawKernelArgDataTy
+{
+    void *bytes;
```
AlexanderKalistratov, line 58: RawKernelArgDataTy doesn't owns `bytes` which rises a lot of questions.
Who's owning `bytes`?
Who's responsible for deleting `bytes`?
How long do we need to keep `bytes` alive?

It is probably better to make `RawKernelArgDataTy` to own `bytes` and allocate and copy data on construction of `RawKernelArgDataTy`.

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
```diff
@@ -62,3 +62,33 @@ bool DPCTLWorkGroupMemory_Available()
...
 #endif
 }
+
+DPCTL_API
+__dpctl_give DPCTLSyclRawKernelArgRef DPCTLRawKernelArg_Create(void *bytes,
+                                                               size_t count)
+{
+    DPCTLSyclRawKernelArgRef rka = nullptr;
+    try {
+        auto RawKernelArg = new RawKernelArgData{bytes, count};
```
AlexanderKalistratov, line 72: Could you please use `unique_ptr` here?
```C++
auto RawKernelArg = std::make_unique(new RawKernelArgData{bytes, count});
rka = wrap<RawKernelArgData>(RawKernelArg.release());
```

sommerlukas: I'm not sure I understand the advantage here. We're still allocating with `new` and we're immediately releasing the `unique_ptr` in the second line. Is there any benefit of this, e.g., exception handling?

libsyclinterface/include/syclinterface/dpctl_sycl_extension_interface.h
sommerlukas: I had considered having `RawKernelArgDataTy` own the money and copy on construction. My reasoning to not do that was that the current approach allows the modification of the data in the time between construction of the raw kernel arg and the kernel launch. If you prefer to have raw kernel arg own the data, I can change that, no problem.

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
sommerlukas: It also deviates from the code style used in the rest of `libsyclinterface`, e.g., [here](https://github.com/IntelPython/dpctl/blob/master/libsyclinterface/source/dpctl_sycl_queue_interface.cpp#L326).

libsyclinterface/include/syclinterface/dpctl_sycl_extension_interface.h
ndgrigorian: I think I favor copying as well, especially because, while the overload from an object exposing buffer is pretty unambiguous (though the GC is a concern), the nbytes and pointer constructor or using the C API directly have a lot of ownership concerns.

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
ndgrigorian: Yes, biggest advantage would be not having to delete it in case an exception is caught.

If an exception did occur here, we would need `delete rka` below in the catch block.

libsyclinterface/include/syclinterface/dpctl_sycl_extension_interface.h
sommerlukas: Okay, I switched to copying in the latest commit.

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
sommerlukas: Using `unique_ptr` know.

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
```diff
@@ -62,3 +62,37 @@ bool DPCTLWorkGroupMemory_Available()
...
+    } catch (std::exception const &e) {
+        error_handler(e, __FILE__, __func__, __LINE__);
+    }
+    return rka;
+}
+
+DPCTL_API
+void DPCTLRawKernelArg_Delete(__dpctl_take DPCTLSyclRawKernelArgRef Ref)
+{
+    delete unwrap<RawKernelArgData>(Ref);
```
AlexanderKalistratov, line 87:
```C++
auto RawArg = unwrap<RawKernelArgData>(Ref);
delete[] RawArg->bytes;
delete RawArg;
```

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
```diff
@@ -62,3 +62,37 @@ bool DPCTLWorkGroupMemory_Available()
...
+                                                               size_t count)
+{
+    DPCTLSyclRawKernelArgRef rka = nullptr;
+    try {
+        auto rawData =
+            std::unique_ptr<unsigned char[]>(new unsigned char[count]);
+        std::memcpy(rawData.get(), bytes, count);
+        auto RawKernelArg = std::unique_ptr<RawKernelArgData>(
+            new RawKernelArgData{rawData.release(), count});
+        rka = wrap<RawKernelArgData>(RawKernelArg.release());
```
AlexanderKalistratov, line 77: Sorry. I've misled you in my previous comment about `unique_ptr`. To make it work it should be something like this:
```C++
        auto rawData =
            std::unique_ptr<unsigned char[]>(new unsigned char[count]);
        std::memcpy(rawData.get(), bytes, count);
        auto RawKernelArg = std::unique_ptr<RawKernelArgData>(
            new RawKernelArgData{rawData.get(), count});
        rka = wrap<RawKernelArgData>(RawKernelArg.get());
        RawKernelArg.release();
        rawData.release();
```

Otherwise `rawData` would leak if `new RawKernelArgData{rawData.release(), count}` throws.

ndgrigorian:
@sommerlukas 
Master now has a fix for nightly build, so please merge or rebase on master so the tests run on nightly compiler

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
sommerlukas: I changed the approach slightly to hold the data in a `std::vector` inside `RawKernelArgDataTy`.

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
sommerlukas: I changed the approach slightly to hold the data in a `std::vector` inside `RawKernelArgDataTy`, because `delete[] RawArg->bytes` gave a warning about incomplete delete.

libsyclinterface/source/dpctl_sycl_extension_interface.cpp
AlexanderKalistratov: ```std::vector``` is used indeed. The same as ```std::unique_ptr```. But it is in ```.cpp``` file. Not in header file which actually has ```.h``` extension and not ```.hpp``` and ```extern "C"``` declaration (https://github.com/IntelPython/dpctl/pull/2038/files#diff-bc26687692ae358c908d9587e3abdfab35454bc0fcab36f47cf59af185e6b5fdR40)

libsyclinterface/include/syclinterface/dpctl_sycl_extension_interface.h
```diff
@@ -34,6 +34,9 @@
 #include "dpctl_sycl_enum_types.h"
 #include "dpctl_sycl_types.h"
 
+#include <cstring>
+#include <vector>
```
ndgrigorian, line 38: These seem to be causing the test failures

libsyclinterface/include/syclinterface/dpctl_sycl_extension_interface.h
```diff
@@ -53,4 +56,31 @@ void DPCTLWorkGroupMemory_Delete(__dpctl_take DPCTLSyclWorkGroupMemoryRef Ref);
 DPCTL_API
 bool DPCTLWorkGroupMemory_Available();
 
+typedef class RawKernelArgDataTy
+{
+public:
+    RawKernelArgDataTy(void *bytes, size_t count) : data(count)
+    {
+        std::memcpy(data.data(), bytes, count);
+    }
```
ndgrigorian, line 65: Unfortunately, the use of copying and `cstring` also brought about another problem, namely, that this is now a class with a constructor under `extern "C"` linkage.

Builds fail but also, it won't be constructible from C with this constructor due to name-mangling issues.

sommerlukas:
Thanks for the feedback @ndgrigorian @AlexanderKalistratov! 

After taking a look at how the opaque pointers for other `std::vector` work, I've changed the approach: I've removed `RawKernelArgDataTy` entirely, so that the opaque pointer (`DPCTLSyclRawKernelArgRef`) now directly points to `std::vector<unsigned char>`. That way, we can avoid defining a struct with `std::vector` in the header, so the offending includes of `<vector>` and `<cstring>` are now removed.

AlexanderKalistratov (approved):
@sommerlukas PR looks great for me, thank you!
I'm not code owner, so we need to wait @ndgrigorian review as well

ndgrigorian:
> @ndgrigorian @AlexanderKalistratov Friendly ping for reviews. CI is looking quite good now, I can unfortunately not see the logs for the single failing Jenkins job.

Sorry, forgot to put my final review on this.

Jenkins job is unrelated. I'll give it a last look over but I think it's ready to go in.

ndgrigorian (approved):
LGTM!

PR #2041, opened 2025-04-07
Updated expected dtypes of input arrays in docstrings of elementwise functions

antonwolfy:
This PR updates docstrings of elementwise functions to add expected dtype of input arrays where missing, to align the dtypes with implementation and to fix some typos which were identified.

### DIFF
--- a/dpctl/tensor/_elementwise_funcs.py
+++ b/dpctl/tensor/_elementwise_funcs.py
@@ -33,7 +33,7 @@ Calculates the absolute value for each element `x_i` of input array `x`.
 
 Args:
     x (usm_ndarray):
-        Input array, expected to have a boolean or numeric data type.
+        Input array. May have any data type.
     out (Union[usm_ndarray, None], optional):
         Output array to populate.
         Array must have the correct shape and the expected data type.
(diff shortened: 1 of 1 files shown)


dpctl/tensor/_elementwise_funcs.py
```diff
@@ -33,7 +33,7 @@
 
 Args:
     x (usm_ndarray):
-        Input array.
+        Input array, expected to have a boolean or numeric data type.
```
ndgrigorian, line 36: technically correct language but a bit wordy. I personally left 'input array' because the type matrix is dense.

Maybe `Input array of any data type.` is acceptable? Or just leave it as `Input array`

Note that the reason behind the `expected to have [...]` language is to indicate to users what types won't have temporaries made without querying `types` from the ufunc object. This should be clarified in a document that covers the element-wise functions more in-depth.

ndgrigorian:
Other than the above comment I think this would be fine to go in, thanks for taking the time for this @antonwolfy

dpctl/tensor/_elementwise_funcs.py
antonwolfy: I noticed there is `Input array. May have any data type.` used in array API for some ufunc (like [here](https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.equal.html)).
We can align and use exactly the same wording if you don't mind.

ndgrigorian: I think that makes sense

antonwolfy: Updated

PR #2049, opened 2025-04-11
Support compilation from SYCL source code

sommerlukas:
This adds support to create a program/executable kernel_bundle from SYCL source code.

It uses the DPC++ `kernel_compiler` extension. As this is only an extension and not all backends are supported, a query on the device was added to check for support for compilation. 

### DIFF
--- a/libsyclinterface/source/dpctl_sycl_device_interface.cpp
+++ b/libsyclinterface/source/dpctl_sycl_device_interface.cpp
@@ -982,3 +982,28 @@ void DPCTLDevice_DisablePeerAccess(__dpctl_keep const DPCTLSyclDeviceRef DRef,
     }
     return;
 }
+
+bool DPCTLDevice_CanCompileSPIRV(__dpctl_keep const DPCTLSyclDeviceRef DRef)
+{
+    auto Dev = unwrap<device>(DRef);
+    auto Backend = Dev->get_platform().get_backend();
+    return Backend == backend::opencl ||
+           Backend == backend::ext_oneapi_level_zero;
+}
+
+bool DPCTLDevice_CanCompileOpenCL(__dpctl_keep const DPCTLSyclDeviceRef DRef)
+{
+    auto Dev = unwrap<device>(DRef);
+    return Dev->get_platform().get_backend() == backend::opencl;
+}
+
+bool DPCTLDevice_CanCompileSYCL(__dpctl_keep const DPCTLSyclDeviceRef DRef)
+{
+#ifdef SYCL_EXT_ONEAPI_KERNEL_COMPILER
+    auto Dev = unwrap<device>(DRef);
+    return Dev->ext_oneapi_can_compile(
+        ext::oneapi::experimental::source_language::sycl);
+#else
+    return false;
+#endif
+}
--- a/dpctl/_backend.pxd
+++ b/dpctl/_backend.pxd
@@ -287,9 +287,12 @@ cdef extern from "syclinterface/dpctl_sycl_device_interface.h":
                                         _peer_access PT)
     cdef void DPCTLDevice_EnablePeerAccess(const DPCTLSyclDeviceRef DRef,
                                            const DPCTLSyclDeviceRef PDRef)
-
     cdef void DPCTLDevice_DisablePeerAccess(const DPCTLSyclDeviceRef DRef,
                                             const DPCTLSyclDeviceRef PDRef)
+    cdef bool DPCTLDevice_CanCompileSPIRV(const DPCTLSyclDeviceRef DRef)
+    cdef bool DPCTLDevice_CanCompileOpenCL(const DPCTLSyclDeviceRef DRef)
+    cdef bool DPCTLDevice_CanCompileSYCL(const DPCTLSyclDeviceRef DRef)
+
 
 cdef extern from "syclinterface/dpctl_sycl_device_manager.h":
     cdef DPCTLDeviceVectorRef DPCTLDeviceVector_CreateFromArray(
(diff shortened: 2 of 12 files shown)


sommerlukas:
The `kernel_compiler` extension also supports SPIR-V and OpenCL as input languages, so it would be possible to replace the existing interop for SPIR-V and OpenCL with the same extension. I however did not do this as part of this PR for three reasons:
* To keep the PR scope limited
* The resulting programs wouldn't be interop anymore.
* It should probably be part of a bigger refactor that also renames program to kernel_bundle then.

ndgrigorian:
Builds seem to be failing on the current compiler version, so this will probably need to be gated like the others behind utilities that check for the availability of the "kernel_compiler" extension

ndgrigorian:
On the plus side, it's passing with 2025.2, and OS builds work—tests won't run until rebased on/merged with master though

sommerlukas:
SYCL runtime compilation should also be supported with 2025.1, but the `registered_names` property used to be called `registered_kernel_names` in that release.
Also, 2025.0 might define the feature test macro, because it had support for OpenCL and SPIR-V in `kernel_compiler` (but no SYCL yet), but doesn't define all properties yet.

@ndgrigorian Do you know how to distinguish the different minor versions? I tried with `__LIBSYCL_MAJOR_VERSION` and `__LIBSYCL_MINOR_VERSION`, but I don't seem to get different versions.

ndgrigorian:
> SYCL runtime compilation should also be supported with 2025.1, but the `registered_names` property used to be called `registered_kernel_names` in that release. Also, 2025.0 might define the feature test macro, because it had support for OpenCL and SPIR-V in `kernel_compiler` (but no SYCL yet), but doesn't define all properties yet.
> 
> @ndgrigorian Do you know how to distinguish the different minor versions? I tried with `__LIBSYCL_MAJOR_VERSION` and `__LIBSYCL_MINOR_VERSION`, but I don't seem to get different versions.

Compiler also defines `__LIBSYCL_PATCH_VERSION `, that's the only other way I would know immediately.

sommerlukas:
I ended up using `__SYCL_COMPILER_VERSION` to distinguish the different DPC++ minor versions. Based on feedback in my other PR #2038, I've also switched to using opaque pointers here and removed `std::vector` and other STL types from the header.

ndgrigorian:
Coverage failure relates to two of the newly-added tests
```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] KernelBundleCreationFromSYCL/TestSYCLKernelBundleFromSource.CheckCreateFromSYCLSource/1, where GetParam() = "opencl:cpu"
[  FAILED  ] KernelBundleCreationFromSYCL/TestSYCLKernelBundleFromSource.CheckGetKernelSYCLSource/1, where GetParam() = "opencl:cpu"
```

sommerlukas:
> Coverage failure relates to two of the newly-added tests
> 
> ```
> [  FAILED  ] 2 tests, listed below:
> [  FAILED  ] KernelBundleCreationFromSYCL/TestSYCLKernelBundleFromSource.CheckCreateFromSYCLSource/1, where GetParam() = "opencl:cpu"
> [  FAILED  ] KernelBundleCreationFromSYCL/TestSYCLKernelBundleFromSource.CheckGetKernelSYCLSource/1, where GetParam() = "opencl:cpu"
> ```

Which version of the Intel OpenCL for CPU runtime is that setup using?

Locally, I had segfaults with these tests in `libintelocl.so`, in the step where the compiler reads in SPIR-V. Upgrading my Intel OpenCL for CPU runtime from `2023.x` to `2025.19.x` fixed the issue, so this may be a case of an older OpenCL CPU runtime not being able to correctly ingest the SPIR-V.

If you have control over the setup, we could try updating the driver. Alternatively, we could deactivate the tests (and RTC support) for OpenCL CPU for now.

ndgrigorian:
@sommerlukas 
master now has change to OS workflow, we can see if picking up the change fixes failures

ndgrigorian:
@sommerlukas 
seems Linux builds are still building on 2025.0.4, which is actually why it's passing

failures in others (on 2025.1.1) are all the same:
```
/home/runner/work/dpctl/dpctl/libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp:854:45: error: no type named 'registered_names' in namespace 'sycl::ext::oneapi::experimental'
  854 | using registered_names_property_t = syclex::registered_names;
      |                                     ~~~~~~~~^
/home/runner/work/dpctl/dpctl/libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp:908:9: error: unknown type name 'registered_names_property_t'
  908 |         registered_names_property_t RegisteredNames;
      |         ^
```

sommerlukas:
> @sommerlukas seems Linux builds are still building on 2025.0.4, which is actually why it's passing
> 
> failures in others (on 2025.1.1) are all the same:
> 
> ```
> /home/runner/work/dpctl/dpctl/libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp:854:45: error: no type named 'registered_names' in namespace 'sycl::ext::oneapi::experimental'
>   854 | using registered_names_property_t = syclex::registered_names;
>       |                                     ~~~~~~~~^
> /home/runner/work/dpctl/dpctl/libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp:908:9: error: unknown type name 'registered_names_property_t'
>   908 |         registered_names_property_t RegisteredNames;
>       |         ^
> ```

I used the version reported by 2025.1(.0), which is the release date, to distinguish between versions `<= 2025.1`, which define `registered_kernel_names` and `>= 2025.2`, which defines `registered_names` instead.
The patch release 2025.1.1 probably still defines the old `registered_kernel_names`, but ofc has a version/release date higher than 2025.1.0. This is either a question of increasing the version number of the distinction or, if there is some overlap in release dates between 2025.2 and 2025.1.1, of finding an alternative approach to determining which property is defined by that release. I'll look into it.

ndgrigorian:
> @ndgrigorian I fixed another issue that caused the coverage testing to fail in the previous CI run. On 2025.1, querying template kernels by template name is not yet supported, they can only be queried by mangled name. I've adapted the test to account for that. Coverage testing is running correctly for me locally, but the CI workflows need approval.
> 
> I'm not quite sure what's causing the Jenkins runs to fail.

Started them up.

And internal tests cite `test_create_program_from_sycl_source` as the source, with `dpctl.program._program.SyclProgramCompilationError`. I'll check the builds as well tomorrow.

sommerlukas:
> [Coverage Status](https://coveralls.io/builds/73639623)
> 
> coverage: 85.335% (-1.0%) from 86.372% when pulling **[153ec69](https://github.com/IntelPython/dpctl/commit/153ec69eb12bca62c7a45412ec34e1b53506666d) on sommerlukas:sycl-source-compilation** into **[ae1e532](https://github.com/IntelPython/dpctl/commit/ae1e532fbb07cf9aad93d4447481a96c175fb056) on IntelPython:master**.

I think the coverage went down, because the Python test that I added runs on LevelZero and is skipped in the coverage testing. Does the coverage machine have LevelZero?

ndgrigorian:
> I think the coverage went down, because the Python test that I added runs on LevelZero and is skipped in the coverage testing. Does the coverage machine have LevelZero?

You're correct, none of our runners have LevelZero. I don't mind the coverage going down in these cases, because there's really not much we can do about it.

sommerlukas:
> > I think the coverage went down, because the Python test that I added runs on LevelZero and is skipped in the coverage testing. Does the coverage machine have LevelZero?
> 
> You're correct, none of our runners have LevelZero. I don't mind the coverage going down in these cases, because there's really not much we can do about it.

Does the runner have OpenCL? We could have the test run on OpenCL instead.

sommerlukas:
I've changed the test to run on `opencl` now.

I've also added support for inspecting the build log in case compilation fails through the message in the `SyclProgramCompilationError` reported back.

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -982,3 +982,28 @@ void DPCTLDevice_DisablePeerAccess(__dpctl_keep const DPCTLSyclDeviceRef DRef,
     }
     return;
 }
+
+bool DPCTLDevice_CanCompileSPIRV(__dpctl_keep const DPCTLSyclDeviceRef DRef)
```
ndgrigorian, line 986: Seems to not be used in SyclDevice

ndgrigorian:
> @ndgrigorian Is the Jenkins failure still related to the tests or just flaky? The rest of the CI is green now.

The build failures are unrelated, but the test failures still seem to be related. On Windows there's a stack overflow in at least one of the Jenkins runners, and on Linux there's `Device linking failed`

ndgrigorian:
Superseded by gh-2206

PR #2054, opened 2025-04-15
task: benchmarking with asv

jharlow-intel:
This PR introduces ASV benchmarks to the repo.
You can find links to ASV's documentation in the added README

Once this is merged, the infra team can start finalizing the work to add these executions to the CI, allowing instant feedback on benchmarks and performance regressions during development.

The benchmarks were tested pretty exhaustively, but please let me know if I need to tweak any of the parameters or add additional benchmark cases. They work for both CPU and GPU SYCL devices as well.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,0 +1,22 @@
+# dpctl benchmarks
+
+Benchmarking dpctl using Airspeed Velocity
+Read more about ASV [here](https://asv.readthedocs.io/en/stable/index.html)
+
+## Usage
+The benchmarks were made with using an existing environment in-mind before execution. You will see the `asv.conf.json` is minimal without any environmental information supplied.
+The expectation is for users to execute `asv run` with an existing environment.
+
+As such, you should have conda or mamba installed, and create an environment [following these instructions](https://intelpython.github.io/dpctl/latest/beginners_guides/installation.html#dpctl-installation)
+Additionally, install `asv` and `libmambapy` to the environment.
+
+Then, you may activate the environment and instruct `asv run` to use this existing environment for the benchmarks by pointing it to the environment's python binary, like so:
+```
+conda activate dpctl_env
+asv run --environment existing:/full/mamba/path/envs/dpctl_env/bin/python
+```
+
+For `level_zero` devices, you might see `USM Allocation` errors unless you use the `asv run` command with `--launch-method spawn`
+
+## Writing new benchmarks
+Read ASV's guidelines for writing benchmarks [here](https://asv.readthedocs.io/en/stable/writing_benchmarks.html)
--- a/benchmarks/asv.conf.json
+++ b/benchmarks/asv.conf.json
@@ -1,0 +1,53 @@
+{
+    // The version of the config file format.  Do not change, unless
+    // you know what you are doing.
+    "version": 1,
+
+    // The name of the project being benchmarked
+    "project": "dpctl",
+
+    // The project's homepage
+    "project_url": "https://github.com/IntelPython/dpctl",
+
+    // The URL or local path of the source code repository for the
+    // project being benchmarked
+    "repo": "..",
+
+    // Customizable commands for building the project.
+    // See asv.conf.json documentation.
+    "build_command": [],
+
+    // List of branches to benchmark. If not provided, defaults to "main"
+    // (for git) or "default" (for mercurial).
+    "branches": ["HEAD"],
+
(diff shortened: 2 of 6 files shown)


jharlow-intel:
fixing the formatting issues now

benchmarks/README.md
```diff
@@ -0,0 +1,24 @@
...
+## Usage
+This was made for use within our CI in mind. As such, you will see the `asv.conf.json` is minimal without any environmental information supplied.
+The expectation is for users to execute `asv run` with an existing environment.
+
+As such, you should have conda or mamba installed, and create a `dpctl-benchmarking` environment as so:
+```
+conda create --name dpctl-benchmarking python=$PYTHON_VERSION dpctl asv libmambapy conda dpcpp_linux-64 --override-channels \
+        -c https://software.repos.intel.com/python/conda \
+        -c conda-forge -y
+```
```
ndgrigorian, line 15: I think this should be re-written from more of a user perspective, as in, I think we should encourage users to run these benchmarks locally, regardless of our intention to use in CI.

The `conda` command also only handles case where the user does not have oneAPI locally installed. We can just refer to creating an environment using the [installation instructions](https://intelpython.github.io/dpctl/latest/beginners_guides/installation.html#dpctl-installation)

then instruct to install addition dependencies (asv, libmamabapy)

benchmarks/asv.conf.json
```diff
@@ -0,0 +1,182 @@
...
+
+    // The Pythons you'd like to test against.  If not provided, defaults
+    // to the current version of Python used to run `asv`.
+    // "pythons": ["3.8", "3.12"],
+
+    // The list of conda channel names to be searched for benchmark
+    // dependency packages in the specified order
+    // "conda_channels": [
+    //     "https://af01p-igk.devtools.intel.com/artifactory/api/conda/idp-conda-pkgserver-igk-local/gold",
+    //     "https://af01p-igk.devtools.intel.com/artifactory/api/conda/idp-conda-pkgserver-igk-local/tools",
```
ndgrigorian, line 64: This goes with my previous comment, external users won't have access to these channels, so the conf won't work for them.

Maybe we should stick to just the Intel channel and benchmark against the version of dpctl on dppy/label/dev, or maybe somehow we can have two configurations? One which we use in the CI and one which is exposed for users.

benchmarks/benchmarks/binary.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+# Dynamically create benchmark methods at the module level
+def generate_benchmark_functions():
+    """Dynamically create benchmark functions for each
+    function and dtype combination.
+    """
+    for fn in function_list:
+        fn_name = fn.name_
+        for dtype1, dtype2 in dtypes[fn]:
```
ndgrigorian, line 110: Mentioned offline but we will need utilities to skip `fp64` and `fp16`, as some devices lack support for these types

benchmarks/benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,15 @@
...
+    Skip the benchmark if the device does not support the given data type.
+    """
+    if (
+        (dtype.name == "float64" or dtype.name == "complex128")
+        and not queue.sycl_device.has_aspect_fp64
+    ) or (dtype.name == "float16" and not queue.sycl_device.has_aspect_fp16):
+        raise SkipNotImplemented(
+            f"Skipping benchmark for {dtype.name} on this device"
+            + " as it is not supported."
+        )
```
ndgrigorian, line 15:
```suggestion
import dpctl.tensor as dpt
from asv_runner.benchmarks.mark import SkipNotImplemented


def skip_unsupported_dtype(q, dtype):
    """
    Skip the benchmark if the device does not support the given data type.
    """
    if (
        (dtype == dpt.float64 or dtype.name == dpt.complex128)
        and not q.sycl_device.has_aspect_fp64
    ) or (dtype == dpt.float16 and not q.sycl_device.has_aspect_fp16):
        raise SkipNotImplemented(
            f"Skipping benchmark for {dtype} on this device"
            + " as it is not supported."
        )
```

I suspect object comparison will be faster than string comparison, also changes make this more consistent with the rest of dpctl

jharlow-intel: Implemented

ndgrigorian (approved):
Regardless of outcome of OS compiler build this LGTM

Thank you @jharlow-intel for working on this

PR #2062, opened 2025-04-25
Use sycl experimental complex namespace for `real` and `imag`

ndgrigorian:
This PR proposes
* `imag` using static constant value of 0 for real inputs
* `real` and `imag` both using the experimental complex namespace calls 

### DIFF
--- a/dpctl/tensor/libtensor/include/kernels/elementwise_functions/imag.hpp
+++ b/dpctl/tensor/libtensor/include/kernels/elementwise_functions/imag.hpp
@@ -31,6 +31,7 @@
 #include <sycl/sycl.hpp>
 #include <type_traits>
 
+#include "sycl_complex.hpp"
 #include "vec_size_util.hpp"
 
 #include "kernels/dpctl_tensor_types.hpp"
@@ -56,11 +57,11 @@ using dpctl::tensor::type_utils::is_complex;
 
 template <typename argT, typename resT> struct ImagFunctor
 {
-
     // is function constant for given argT
-    using is_constant = typename std::false_type;
+    using is_constant =
+        typename std::is_same<is_complex<argT>, std::false_type>;
     // constant value, if constant
-    // constexpr resT constant_value = resT{};
+    static constexpr resT constant_value = resT{0};
     // is function defined for sycl::vec
     using supports_vec = typename std::false_type;
     // do both argTy and resTy support sugroup store/load operation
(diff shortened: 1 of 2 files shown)


dpctl/tensor/libtensor/include/kernels/elementwise_functions/imag.hpp
```diff
@@ -70,11 +71,13 @@ template <typename argT, typename resT> struct ImagFunctor
     resT operator()(const argT &in) const
     {
         if constexpr (is_complex<argT>::value) {
-            return std::imag(in);
+            using realT = typename argT::value_type;
+            using sycl_complexT = typename exprm_ns::complex<realT>;
+            return exprm_ns::imag(sycl_complexT(in));
```
antonwolfy, line 76: It might be unclear from the PR description what was the motivation to switch to experimental namespace and why other place where `std::imag` used are not impacted.. Would it make sense to add clarity on that?

dpctl/tensor/libtensor/include/kernels/elementwise_functions/imag.hpp
```diff
@@ -70,11 +71,13 @@ template <typename argT, typename resT> struct ImagFunctor
     resT operator()(const argT &in) const
     {
         if constexpr (is_complex<argT>::value) {
```
antonwolfy, line 73: There is `is_complex_v` available in the `type_utils.hpp`
```suggestion
        if constexpr (is_complex_v<argT>) {
```

ndgrigorian: yes, did not exist at the time these functions were made. I can switch to it.

dpctl/tensor/libtensor/include/kernels/elementwise_functions/imag.hpp
ndgrigorian: The initial motivation was old technical debt to add the constant value for `imag`, since we may as well use it for real-valued types.

In terms of why add in the experimental namespace change, `complex` header is not supported on AMD and this has been an issue with recent AMD builds. I've pushed another branch which removes (almost) every use of `complex` in kernel code aside from conversion to SYCL complex type. I want to see if this resolves issues with AMD builds on OS compiler.

ndgrigorian: In general, I think it would be best to slowly transition to full use of the sycl complex extension. We also have benchmarking now and can see if this PR brings regressions.

antonwolfy: I see now, thank you. I'm totally with that. Just it was a bit unclear from the PR description.

antonwolfy (approved):
LGTM

PR #2071, opened 2025-05-01
Change `"max dimensions"` to `None` in array API capabilities

ndgrigorian:
`tensor.usm_ndarray` can support potentially arbitrary dimensions

Changes will need to be made where NumPy arrays are created from `usm_ndarray` (printing, dlpack with `kDLCPU`, `asnumpy`/`from_numpy`)

### DIFF
--- a/dpctl/tensor/_array_api.py
+++ b/dpctl/tensor/_array_api.py
@@ -75,7 +75,7 @@ class Info:
         self._capabilities = {
             "boolean indexing": True,
             "data-dependent shapes": True,
-            "max dimensions": 64,
+            "max dimensions": None,
         }
         self._all_dtypes = {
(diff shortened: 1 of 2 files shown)


antonwolfy (approved):
LGTM!
Do we need to not the change in `CHANGELOG.md`? Or it will be done before the release directly?

ndgrigorian:
> LGTM! Thank you @ndgrigorian Do we need to note the change in `CHANGELOG.md`? Or it will be done before the release directly?

It will be done before the release.

Adding to changelog as a task was added to simplify the process at release, but it was found that it didn't really help avoid a number of ugly merge conflicts, etc.

Really the chore should be removed and we should switch to towncrier.

PR #2077, opened 2025-05-05
Support peer access DPC++ extension

ndgrigorian:
This PR adds full support for the [DPC++ `sycl_ext_oneapi_peer_access` extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_peer_access.asciidoc)

This adds a Python interface to calls that check if a `dpctl.SyclDevice` can enable peer access to another `dpctl.SyclDevice` and, if so, to enable or disable it.

### DIFF
--- a/dpctl/_sycl_device.pyx
+++ b/dpctl/_sycl_device.pyx
@@ -1855,7 +1855,7 @@ cdef class SyclDevice(_SyclDevice):
             raise ValueError("Internal error: NULL device vector encountered")
         return _get_devices(cDVRef)
 
-    def can_access_peer_access_supported(self, peer):
+    def can_access_peer(self, peer, value="access_supported"):
         """ Returns ``True`` if this device (``self``) can enable peer access
         to USM device memory on ``peer``, ``False`` otherwise.
 
@@ -1869,6 +1869,25 @@ cdef class SyclDevice(_SyclDevice):
             peer (:class:`dpctl.SyclDevice`):
                 The :class:`dpctl.SyclDevice` instance to check for peer access
                 by this device.
+            value (str, optional):
+                Specifies the kind of peer access being queried
+
+                - ``"access_supported"``
+                    Returns ``True`` if it is possible for this device to
+                    enable peer access to USM device memory on ``peer``.
+
+                - ``"atomics_supported"``
+                    Returns ``True`` if it is possible for this device to
+                    concurrently access and atomically modify USM device
+                    memory on ``peer`` when enabled.
+
+                    If ``False`` is returned, these operations result in
+                    undefined behavior.
+
+                    Note: atomics must have ``memory_scope::system`` when
+                    modifying memory on a peer device.
+
+                Default: ``"access_supported"``
 
         Returns:
             bool:
@@ -1881,6 +1900,19 @@ cdef class SyclDevice(_SyclDevice):
         """
         cdef SyclDevice p_dev
 
+        if not isinstance(value, str):
+            raise TypeError(
+                f"Expected `value` to be of type str, got {type(value)}"
+            )
+        if value == "access_supported":
+            access_type = _peer_access._access_supported
+        elif value == "atomics_supported":
+            access_type = _peer_access._atomics_supported
+        else:
+            raise ValueError(
+                "`value` must be 'access_supported' or 'atomics_supported', "
+                f"got {value}"
+            )
         if not isinstance(peer, SyclDevice):
             raise TypeError(
                 "peer device must be a `dpctl.SyclDevice`, got "
@@ -1891,49 +1923,7 @@ cdef class SyclDevice(_SyclDevice):
             return DPCTLDevice_CanAccessPeer(
                 self._device_ref,
                 p_dev.get_device_ref(),
-                _peer_access._access_supported
-            )
-        return False
-
-    def can_access_peer_atomics_supported(self, peer):
-        """ Returns ``True`` if this device (``self``) can concurrently access
-        and modify USM device memory on ``peer`` when peer access is enabled,
-        ``False`` otherwise.
-
-        If peer access is supported, it may be enabled by calling
-        :meth:`.enable_peer_access`.
-
-        For details, see
-        :oneapi_peer_access:`DPC++ peer access SYCL extension <>`.
-
-        Args:
-            peer (:class:`dpctl.SyclDevice`):
-                The :class:`dpctl.SyclDevice` instance to check for concurrent
-                peer access and modification by this device.
-
-        Returns:
-            bool:
-                ``True`` if this device may concurrently access and modify USM
-                device memory on ``peer`` when peer access is enabled,
-                otherwise ``False``.
-
-        Raises:
-            TypeError:
-                If ``peer`` is not :class:`dpctl.SyclDevice`.
-        """
-        cdef SyclDevice p_dev
-
-        if not isinstance(peer, SyclDevice):
-            raise TypeError(
-                "peer device must be a `dpctl.SyclDevice`, got "
-                f"{type(peer)}"
-            )
-        p_dev = <SyclDevice>peer
-        if _check_peer_access(self, p_dev):
-            return DPCTLDevice_CanAccessPeer(
-                self._device_ref,
-                p_dev.get_device_ref(),
-                _peer_access._atomics_supported
+                access_type
             )
         return False
 
@@ -2002,11 +1992,10 @@ cdef class SyclDevice(_SyclDevice):
             )
         p_dev = <SyclDevice>peer
         _raise_invalid_peer_access(self, p_dev)
-        if _check_peer_access(self, p_dev):
-            DPCTLDevice_DisablePeerAccess(
-                self._device_ref,
-                p_dev.get_device_ref()
-            )
+        DPCTLDevice_DisablePeerAccess(
+            self._device_ref,
+            p_dev.get_device_ref()
+        )
         return
 
     @property
--- a/dpctl/_sycl_device.pyx
+++ b/dpctl/_sycl_device.pyx
@@ -1980,6 +1980,8 @@ cdef class SyclDevice(_SyclDevice):
     def disable_peer_access(self, peer):
         """ Disables peer access to ``peer`` from this device (``self``).
 
+        Peer access may be enabled by calling :meth:`.enable_peer_access`.
+
         For details, see
         :oneapi_peer_access:`DPC++ peer access SYCL extension <>`.
 
--- a/dpctl/_sycl_device.pyx
+++ b/dpctl/_sycl_device.pyx
@@ -239,6 +239,43 @@ cdef inline bint _check_peer_access(SyclDevice dev, SyclDevice peer) except *:
     return False
 
 
+cdef inline void _raise_invalid_peer_access(
+    SyclDevice dev,
+    SyclDevice peer,
+) except *:
+    """
+    Check peer access ahead of time and raise errors for invalid cases.
+    """
+    cdef list _peer_access_backends = [
+        _backend_type._CUDA,
+        _backend_type._HIP,
+        _backend_type._LEVEL_ZERO
+    ]
+    cdef _backend_type BTy1 = DPCTLDevice_GetBackend(dev._device_ref)
+    cdef _backend_type BTy2 = DPCTLDevice_GetBackend(peer.get_device_ref())
+    if (BTy1 != BTy2):
+        raise ValueError(
+            f"Device with backend {_backend_type_to_filter_string_part(BTy1)} "
+            "cannot peer access device with backend "
+            f"{_backend_type_to_filter_string_part(BTy2)}"
+        )
+    if (BTy1 not in _peer_access_backends):
+        raise ValueError(
+            "Peer access not supported for backend "
+            f"{_backend_type_to_filter_string_part(BTy1)}"
+        )
+    if (BTy2 not in _peer_access_backends):
+        raise ValueError(
+            "Peer access not supported for backend "
+            f"{_backend_type_to_filter_string_part(BTy2)}"
+        )
+    if (dev == peer):
+        raise ValueError(
+            "Peer access cannot be enabled between a device and itself"
+        )
+    return
+
+
 @functools.lru_cache(maxsize=None)
 def _cached_filter_string(d : SyclDevice):
     """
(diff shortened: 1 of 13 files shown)


libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -903,3 +902,117 @@ DPCTLDevice_GetCompositeDevice(__dpctl_keep const DPCTLSyclDeviceRef DRef)
...
+            return false;
+        }
+
+        if (BE2 != sycl::backend::ext_oneapi_level_zero &&
+            BE2 != sycl::backend::ext_oneapi_cuda &&
+            BE2 != sycl::backend::ext_oneapi_hip)
+        {
+            error_handler("Backend " + std::to_string(static_cast<int>(BE2)) +
+                              " does not support peer access",
+                          __FILE__, __func__, __LINE__);
```
antonwolfy, line 933: `sycl::backend` supports stream insertion overloading, so it can be simplified, like:
```suggestion
            std::ostringstream os;
            os << "Backend " << BE1 << " does not support peer access";
            error_handler(os.str(), __FILE__, __func__, __LINE__);
```
or in similar way

antonwolfy: While printing integer value does not seem informative about backend name

dpctl/_sycl_device.pyx
```diff
@@ -1792,6 +1796,245 @@ cdef class SyclDevice(_SyclDevice):
...
+                "second argument must be a `dpctl.SyclDevice`, got "
+                f"{type(peer)}"
+            )
+        p_dev = <SyclDevice>peer
+        BTy1 = DPCTLDevice_GetBackend(self._device_ref)
+        if (
+            BTy1 != _backend_type._CUDA and
+            BTy1 != _backend_type._HIP and
+            BTy1 != _backend_type._LEVEL_ZERO
+        ):
```
antonwolfy, line 1841: It can be slightly simplified:
```suggestion
        if BTy1 not in [_backend_type._CUDA, _backend_type._HIP, _backend_type._LEVEL_ZERO]:
```

dpctl/_sycl_device.pyx
```diff
@@ -1792,6 +1796,245 @@ cdef class SyclDevice(_SyclDevice):
...
 
+    def can_access_peer_access_supported(self, peer):
+        """ Returns ``True`` if this device (``self``) can enable peer access
+        to USM device memory on ``peer``, ``False`` otherwise.
+
+        If peer access is supported, it may be enabled by calling
+        :meth:`.enable_peer_access`.
+
+        For details, see
+        :oneapi_peer_access:`DPC++ peer access SYCL extension <>`.
```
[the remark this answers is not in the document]
ndgrigorian: Fixed, was in the legacy `urls.json` instead of the current one

dpctl/tests/test_sycl_device.py
```diff
@@ -341,3 +341,69 @@ def test_get_component_devices_from_composite():
...
+    devices = platform.get_devices()
+    if len(devices) < 2:
+        pytest.skip(
+            f"Platform {platform_name} does not have enough devices to "
+            "test peer access"
+        )
+    dev0 = devices[0]
+    dev1 = devices[1]
+    if dev0.can_access_peer_access_supported(dev1):
+        dev0.enable_peer_access(dev1)
```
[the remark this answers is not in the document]
ndgrigorian: Unfortunately there isn't

dpctl/tests/test_sycl_device.py
```diff
@@ -341,3 +341,69 @@ def test_get_component_devices_from_composite():
...
+        pytest.skip(f"{str(e)} {platform_name}")
+    devices = platform.get_devices()
+    if len(devices) < 2:
+        pytest.skip(
+            f"Platform {platform_name} does not have enough devices to "
+            "test peer access"
+        )
+    dev0 = devices[0]
+    dev1 = devices[1]
+    if dev0.can_access_peer_access_supported(dev1):
```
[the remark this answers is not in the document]
ndgrigorian: No, because there's no way to distinguish between the two. When you enable, `can_access_peer_access_supported` is a weaker condition than the atomics one. So atomics is true, this is true, but not always the converse.

dpctl/tests/test_sycl_device.py
```diff
@@ -341,3 +341,69 @@ def test_get_component_devices_from_composite():
...
+        pytest.skip("No default device available")
+    bad_dev = dict()
+    with pytest.raises(TypeError):
+        dev.can_access_peer_access_supported(bad_dev)
+    with pytest.raises(TypeError):
+        dev.can_access_peer_atomics_supported(bad_dev)
+    with pytest.raises(TypeError):
+        dev.enable_peer_access(bad_dev)
+    with pytest.raises(TypeError):
+        dev.disable_peer_access(bad_dev)
```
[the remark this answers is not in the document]
ndgrigorian: good idea

ndgrigorian:
Noting here that for future reference, we will probably need to document how this extension interacts with sub-devices and composite devices.

For example, per composite device documentation
```
Interaction with the default context

If the implementation supports the sycl_ext_oneapi_default_context extension, then each platform has the notion of a default context.
This default context contains only the platform’s root devices, which are those returned by platform::get_devices.
Any composite devices for the platform are not considered root devices, and thus the default context does not include these composite devices.
If an application wants a context that contains the composite devices, then it must explicitly create such a context.
```
this means that while the peer access methods will work between composite device and its component devices, and vice versa, to actually leverage it, the user needs to explicit construct a context containing component and composite devices.

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -903,3 +902,86 @@ DPCTLDevice_GetCompositeDevice(__dpctl_keep const DPCTLSyclDeviceRef DRef)
     else
         return nullptr;
 }
+
+bool _CallPeerAccess(device dev, device peer)
```
antonwolfy, line 906:
```suggestion
static inline bool _CallPeerAccess(device dev, device peer)
```

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -903,3 +902,86 @@ DPCTLDevice_GetCompositeDevice(__dpctl_keep const DPCTLSyclDeviceRef DRef)
...
+                               __dpctl_keep const DPCTLSyclDeviceRef PDRef,
+                               DPCTLPeerAccessType PT)
+{
+    bool canAccess = false;
+    auto D = unwrap<device>(DRef);
+    auto PD = unwrap<device>(PDRef);
+    if (D && PD) {
+        if (_CallPeerAccess(*D, *PD)) {
+            try {
+                canAccess = D->ext_oneapi_can_access_peer(
```
antonwolfy, line 935: Do we need to wrap with `#ifdef SYCL_EXT_ONEAPI_PEER_ACCESS` to check the api is available?

libsyclinterface/tests/test_sycl_peer_access.cpp
```diff
@@ -0,0 +1,139 @@
...
+    ASSERT_TRUE(D1 != nullptr);
+    bool canEnable = false;
+    EXPECT_NO_FATAL_FAILURE(canEnable = DPCTLDevice_CanAccessPeer(
+                                D0, D1, DPCTLPeerAccessType::access_supported));
+    if (canEnable) {
+        EXPECT_NO_FATAL_FAILURE(DPCTLDevice_EnablePeerAccess(D0, D1));
+        EXPECT_NO_FATAL_FAILURE(DPCTLDevice_DisablePeerAccess(D0, D1));
+    }
+}
+
```
antonwolfy, line 106: Do we need to check the use case when passing the same device as a peer?

libsyclinterface/source/dpctl_sycl_device_interface.cpp
ndgrigorian: Hard to say—if we did, then it opens questions like what we should do about the Python APIs. It has also been supported for some time (since 2023, so well before min. version of 2024.2). The biggest advantage to the `ifdef` from my perspective would be to support other SYCL compilers, but I have a feeling we are not really doing that anyway right now.

If it seems like it wouldn't be too bad, I'll make the change.

libsyclinterface/tests/test_sycl_peer_access.cpp
ndgrigorian: Yes, though I'm not going to add a check for whether it can be accessed as a peer or not—I want to wait and see what the compiler ends up doing. With any luck, a lot of this backend validation logic, etc. will go away after some compiler changes.

libsyclinterface/source/dpctl_sycl_device_interface.cpp
antonwolfy: I thought it has to be aligned with min required DPC++ compiler version in `meta.yaml`.
One we are bumping it we will be able to remove the redundant ifdefs. But currently it sets to `2024.2`.

Or depending on what we are stating in the documentation about min supported version (currently we didn't, but probably should).

antonwolfy: But initially I thought if did that with `SYCL_EXT_ONEAPI_WORK_GROUP_MEMORY` and `SYCL_EXT_ONEAPI_RAW_KERNEL_ARG`, it would make sense to keep consistency here with other new extensions we are adding.

ndgrigorian: > I thought it has to be aligned with min required DPC++ compiler version in `meta.yaml`. One we are bumping it we will be able to remove the redundant ifdefs. But currently it sets to `2024.2`.

That's correct, what I meant was—the extension was supported well before `2024.2` so it should be available no matter what. In general I agree on the consistency, it may be sensible to go back and add similar for the composite devices extension in a separate PR.

ndgrigorian: I forgot—compiler is currently not defining `SYCL_EXT_ONEAPI_PEER_ACCESS` either.

Maybe we are best suited waiting for the extension to be updated, as opposed to doing it ourself...

libsyclinterface/tests/test_sycl_peer_access.cpp
```diff
@@ -0,0 +1,150 @@
...
+TEST_P(TestDPCTLPeerAccess, ChkPeerAccessToSelf)
+{
+    auto D0 = DPCTLDeviceVector_GetAt(DV, 0);
+    ASSERT_TRUE(D1 != nullptr);
+    bool canEnable = false;
+    EXPECT_NO_FATAL_FAILURE(canEnable = DPCTLDevice_CanAccessPeer(
```
antonwolfy, line 112: Do we need to add `ASSERT_FALSE(canEnable)` after that?

libsyclinterface/tests/test_sycl_peer_access.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                D0, D0, DPCTLPeerAccessType::access_supported));
```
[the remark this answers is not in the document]
antonwolfy: In general it seems access type can be also parametrized to reduce duplication, but not sure if it is easy to implement with gtest f/w.

libsyclinterface/tests/test_sycl_peer_access.cpp
ndgrigorian: I think it would be better not to test for a specific value, because the compiler might change soon.

libsyclinterface/tests/test_sycl_peer_access.cpp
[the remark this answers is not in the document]
ndgrigorian: don't think it adds much to repeat the call—passes through the same function and logic anyway. There aren't any branches based on the enum.

antonwolfy (dismissed):

ndgrigorian:
@antonwolfy 
made one more quick round of changes—combined the `can_access_peer` methods into one since extension defaults to `access_supported` anyway. This seems more convenient for users.

Take a look, it should have overall small impact on the implementation.

libsyclinterface/helper/source/dpctl_utils_helper.cpp
```diff
@@ -452,6 +452,32 @@ DPCTLPartitionAffinityDomainType DPCTL_SyclPartitionAffinityDomainToDPCTLType(
...
+ext::oneapi::peer_access
+DPCTL_DPCTLPeerAccessTypeToSycl(DPCTLPeerAccessType PeerAccessTy)
+{
+    switch (PeerAccessTy) {
+    case DPCTLPeerAccessType::access_supported:
+        return ext::oneapi::peer_access::access_supported;
+    case DPCTLPeerAccessType::atomics_supported:
+        return ext::oneapi::peer_access::atomics_supported;
+    default:
+        throw std::runtime_error("Unsupported peer_access type");
```
antonwolfy, line 464: The error case is not covered by a test

libsyclinterface/helper/source/dpctl_utils_helper.cpp
```diff
@@ -452,6 +452,32 @@ DPCTLPartitionAffinityDomainType DPCTL_SyclPartitionAffinityDomainToDPCTLType(
...
+DPCTLPeerAccessType
+DPCTL_SyclPeerAccessToDPCTLType(ext::oneapi::peer_access PeerAccess)
+{
+    switch (PeerAccess) {
+    case ext::oneapi::peer_access::access_supported:
+        return DPCTLPeerAccessType::access_supported;
+    case ext::oneapi::peer_access::atomics_supported:
+        return DPCTLPeerAccessType::atomics_supported;
+    default:
+        throw std::runtime_error("Unsupported peer_access type");
```
antonwolfy, line 477: The error case is not covered by a test

antonwolfy (approved):
In overall LGTM, few small comments which might be address separately.

dpctl/_sycl_device.pyx
```diff
@@ -1792,6 +1855,150 @@ cdef class SyclDevice(_SyclDevice):
...
+
+                - ``"atomics_supported"``
+                    Returns ``True`` if it is possible for this device to
+                    concurrently access and atomically modify USM device
+                    memory on ``peer`` when enabled. Atomics must have
+                    ``memory_scope::system`` when modifying memory on a peer
+                    device.
+
+                    If ``False`` is returned, these operations result in
+                    undefined behavior.
```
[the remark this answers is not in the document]
ndgrigorian: The specification for the extension only lists the undefined behavior under this case, which is why I put it here.

Truthfully, I'm not sure what happens if you try to access without peer access enabled, so in that sense, I think this comment makes sense.

ndgrigorian:
Addressing other comments in follow-up

PR #2080, opened 2025-05-14
DOC: Clarify details about conda installs

david-cortes-intel:
From internal discussions, installing this package requires pulling all non-intel dependencies from conda-forge. This PR fixes the instructions so that following them wouldn't result in installing packages from the `anaconda` channel.


ndgrigorian (approved):
LGTM! Thank you for the contribution @david-cortes-intel

PR #2095, opened 2025-06-05
Add passing dependent events to `copy` calls

antonwolfy:
The reshape and roll kernels have a special handling for 1-element input array, where `exec_q.copy()` is used.
But it is still possible that the array data depends on submitted events even in that case with only 1-element.

Thus this PR is fixing the issue with unsafe copying of 1-element input array.
The issue was initially detected in dpnp tests on ARL BMG Linux machine in scope of internal testing.


ndgrigorian (approved):
This is a good catch

LGTM

ndgrigorian:
`copy_usm_ndarray_for_roll_nd` has the same issue, I will push a commit to fix it

PR #2096, opened 2025-06-05
Enable architecture selection for `DPCTL_TARGET_CUDA`

vlad-perevezentsev:
This PR proposes to change `DPCTL_TARGET_CUDA` CMake option from a boolean to a string allowing users to specify a CUDA architecture (e.g. `sm_80`). If not specified, it defaults to `sm_50`.

```bash
$ python scripts/build_locally.py --verbose --cmake-opts="-DDPCTL_TARGET_CUDA=<cuda_arch>"
# or
$ python scripts/build_locally.py --verbose --cmake-opts="-DDPCTL_TARGET_CUDA=ON"

```

The specified architecture is used to construct a SYCL alias target (e.g. `nvidia_gpu_sm_80`) and passed via `-fsycl-targets` option, following [OneAPI for NVIDIA GPUs](https://developer.codeplay.com/products/oneapi/nvidia/2025.0.0/guides/get-started-guide-nvidia#use-dpc-to-target-nvidia-gpus) 

Additionally removing `DPCTL_TARGET_CUDA` env handling logic


CMakeLists.txt
```diff
@@ -51,14 +52,21 @@ set(_dpctl_sycl_target_compile_options)
...
+set(_dpctl_cuda_arch)
 set(_dpctl_amd_targets)
 if ("x${DPCTL_SYCL_TARGETS}" STREQUAL "x")
-   if (DPCTL_TARGET_CUDA)
-      set(_dpctl_sycl_targets "nvptx64-nvidia-cuda,spir64-unknown-unknown")
-   else()
-      if (DEFINED ENV{DPCTL_TARGET_CUDA})
-          set(_dpctl_sycl_targets "nvptx64-nvidia-cuda,spir64-unknown-unknown")
+   if (NOT "x${DPCTL_TARGET_CUDA}" STREQUAL "x")
+      if (DPCTL_TARGET_CUDA STREQUAL "ON")
```
ndgrigorian: I think this can be made a bit more robust, per [CMake documentation](https://cmake.org/cmake/help/latest/command/if.html#constant)
```
True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number
```
then for `if(<variable>)`
```
True if given a variable that is defined to a value that is not a false constant.
```
if this is accurate, then `if(DPCTL_TARGET_CUDA)` can be used to check for true/false, and if true, set to default value, then we can check for a match with `"sm_*"`

then we can also allow `DPCTL_TARGET_CUDA=FALSE` or `=0`, etc. and just pass if given this way

vlad-perevezentsev: CMake treats any non-false value as `True` including invalid inputs such as `qwe`. This will incorrectly set the default value to `sm_50`

To handle this correctly we still have to explicitly check that `DPCTL_TARGET_CUDA MATCHES “^(ON|TRUE|YES|1)$”` before assigning default value and throw `FATAL_ERROR in `else` branch for invalid inputs.

If it is ok I will implement it

ndgrigorian: Good point, I think that's the most sensible approach

vlad-perevezentsev: done

CMakeLists.txt
```diff
@@ -51,15 +52,24 @@ set(_dpctl_sycl_target_compile_options)
...
+set(_dpctl_cuda_arch)
 set(_dpctl_amd_targets)
+
 if ("x${DPCTL_SYCL_TARGETS}" STREQUAL "x")
-   if (DPCTL_TARGET_CUDA)
-      set(_dpctl_sycl_targets "nvptx64-nvidia-cuda,spir64-unknown-unknown")
-   else()
-      if (DEFINED ENV{DPCTL_TARGET_CUDA})
-          set(_dpctl_sycl_targets "nvptx64-nvidia-cuda,spir64-unknown-unknown")
+   if (NOT "x${DPCTL_TARGET_CUDA}" STREQUAL "x")
```
antonwolfy: Is it fair to validate `DPCTL_TARGET_CUDA` only in case when empty `DPCTL_SYCL_TARGETS`?

CMakeLists.txt
```diff
@@ -25,9 +25,10 @@ option(DPCTL_GENERATE_COVERAGE_FOR_PYBIND11_EXTENSIONS
     "Build dpctl pybind11 offloading extensions with coverage instrumentation"
     OFF
 )
-option(DPCTL_TARGET_CUDA
-    "Build DPCTL to target CUDA devices"
-    OFF
+set(DPCTL_TARGET_CUDA
+    ""
+    CACHE STRING
+    "Build DPCTL to target CUDA devices. Set to a truthy value (e.g., ON, TRUE) to use default architecture (sm_50), or to a specific architecture like sm_80."
```
antonwolfy: Is it possible to split the line? It's too long.

antonwolfy: Or is it possible to pass several value to `DPCTL_TARGET_CUDA` variable?
```suggestion
    "Build DPCTL to target CUDA device. Set to a truthy value (e.g., ON, TRUE) to use default architecture (sm_50), or to a specific architecture like sm_80."
```

docs/doc_sources/beginners_guides/installation.rst
```diff
@@ -159,13 +159,32 @@ The following plugins from CodePlay are supported:
...
     python scripts/build_locally.py --verbose --cmake-opts="-DDPCTL_TARGET_CUDA=ON"
 
-And for AMD devices
+Note that kernels are built for ``sm_50`` by default, allowing them to work on a wider
+range of architectures, but limiting the usage of more recent CUDA features.
+
+For reference, compute architecture strings like ``sm_80`` are based on
+CUDA Compute Capability. A complete mapping between NVIDIA GPU models and their
+respective ``sm_XX`` values can be found in the official
+`CUDA GPU Compute Capability <https://developer.nvidia.com/cuda-gpus>`_.
```
antonwolfy: The mapping is not clear from the reference doc.

ndgrigorian: Yes, seems they aren't necessarily related either ([see here](https://docs.nvidia.com/cuda/volta-compatibility-guide/index.html#building-volta-compatible-apps-using-cuda-8-0) and below it)

CMakeLists.txt
ndgrigorian: it was how we were doing it before—but it looks like current logical flow will add HIP targets even when `DPCTL_SYCL_TARGETS` is not none, but not CUDA

so that should probably be changed, either make `DPCTL_SYCL_TARGETS` exclusive from both or check `DPCTL_TARGET_CUDA` as well

CMakeLists.txt
vlad-perevezentsev: done

vlad-perevezentsev: only 1 device can be passed. I applied it

CMakeLists.txt
vlad-perevezentsev: In my understanding, when the user passes `DPCTL_SYCL_TARGETS` he is responsible for the correctness of the flags.

The logic of checking `if (NOT “x${DPCTL_TARGET_HIP}” STREQUAL “x”)` when `DPCTL_SYCL_TARGETS` is `not none` was added to pass the correct compile and link options.
```
if(_dpctl_amd_targets)
      list(APPEND _dpctl_sycl_target_compile_options -Xsycl-target-backend=amdgcn-amd-amdhsa --offload-arch=${_dpctl_amd_targets})
      list(APPEND _dpctl_sycl_target_link_options -Xsycl-target-backend=amdgcn-amd-amdhsa --offload-arch=${_dpctl_amd_targets})
   endif()
```
I am already working on PR that will refresh the logic for AMD build using aliases to remove `if(_dpctl_amd_targets)` branch.

docs/doc_sources/beginners_guides/installation.rst
vlad-perevezentsev: A CUDA developer [notes](https://forums.developer.nvidia.com/t/how-should-i-use-correctly-the-sm-xx-and-compute-xx/219160/2) that `sm_XX` refers to machine code for a specific GPU hardware architecture. Since each Compute Capability version corresponds to a particular architecture (CC 8.0 -> Ampere A100) it is reasonable to say that `sm_80` corresponds to `CC 8.0`

I changed the text a bit

ndgrigorian (approved):
LGTM

ndgrigorian (approved):

Issue #1046, opened 2023-01-26
Copying from device with float64 support to device without float64 support failed


npolina4:
``` python
import dpctl.tensor as dpt

#Default type is float64.
Xcpu = dpt.ones(4, device="cpu")

#GPU without float64 support. Should have been cast to dtype float32, but failed.
dpt.asarray(Xcpu, device="gpu")
```

PR #2097, opened 2025-06-05
Resolve gh-1046

ndgrigorian:
Closes gh-1046 

This PR resolves `dpt.asarray(x, device=dev)` failing for some array `x` with dtype not supported by device `dev`

### DIFF
--- a/dpctl/tensor/_ctors.py
+++ b/dpctl/tensor/_ctors.py
@@ -111,8 +111,6 @@ def _asarray_from_usm_ndarray(
         raise TypeError(
             f"Expected dpctl.tensor.usm_ndarray, got {type(usm_ndary)}"
         )
-    if dtype is None:
-        dtype = usm_ndary.dtype
     if usm_type is None:
         usm_type = usm_ndary.usm_type
     if sycl_queue is not None:
@@ -122,6 +120,8 @@ def _asarray_from_usm_ndarray(
         copy_q = normalize_queue_device(sycl_queue=sycl_queue, device=exec_q)
     else:
         copy_q = usm_ndary.sycl_queue
+    if dtype is None:
+        dtype = _map_to_device_dtype(usm_ndary.dtype, copy_q)
     # Conditions for zero copy:
     can_zero_copy = copy is not True
     #    dtype is unchanged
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_ctors.py
```diff
@@ -122,6 +120,8 @@ def _asarray_from_usm_ndarray(
         copy_q = normalize_queue_device(sycl_queue=sycl_queue, device=exec_q)
     else:
         copy_q = usm_ndary.sycl_queue
+    if dtype is None:
```
antonwolfy, line 123: Is there any way to add a test to cover that change?

antonwolfy (dismissed):
In overall, LGTM
Thank you @ndgrigorian

dpctl/tensor/_ctors.py
ndgrigorian: I've added such a test, the logic was complex and brings up the question of a `dpctl.select_device_with_aspects` that permits only finding a device without a given aspect

dpctl/tests/test_tensor_asarray.py
```diff
@@ -623,3 +623,33 @@ def test_asarray_support_for_usm_ndarray_protocol(usm_type):
...
+
+@pytest.mark.parametrize("dt", [dpt.float16, dpt.float64, dpt.complex128])
+def test_asarray_to_device_with_unsupported_dtype(dt):
+    aspect = "fp16" if dt == dpt.float16 else "fp64"
+    try:
+        d0 = dpctl.select_device_with_aspects(aspect)
+    except dpctl.SyclDeviceCreationError:
+        pytest.skip("No device with aspect for test")
+    d1 = None
+    try:
```
antonwolfy: Would it be better to do through the loop?
```python
    for d in dpctl.get_devices():
        if d.default_selector_score < 0:
            pass
        try:
            d1 = dpctl.select_device_with_aspects(d.device_type.name, excluded_aspects=[aspect])
        except dpctl.SyclDeviceCreationError:
            pass
```

ndgrigorian: sure, it's definitely less ugly

PR #2098, opened 2025-06-06
Implement `tensor.isin`

ndgrigorian:
This PR proposes an implementation for `isin`, a function likely coming to a future array API specification, which leverages a similar kernel to the implementation of `searchsorted`

This implementation uses the `searchsorted` kernel to check if the value has a position in the array. If that position is the number of elements in the array, it is not a member. Otherwise, if `arr[pos] == val` for some array `arr` being searched for value `val`, then `val` is a member.

### DIFF
--- a/dpctl/tensor/CMakeLists.txt
+++ b/dpctl/tensor/CMakeLists.txt
@@ -112,6 +112,7 @@ set(_reduction_sources
     ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/reductions/sum.cpp
 )
 set(_sorting_sources
+    ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/sorting/isin.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/sorting/merge_sort.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/sorting/merge_argsort.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/sorting/radix_sort.cpp
--- a/dpctl/tensor/_set_functions.py
+++ b/dpctl/tensor/_set_functions.py
@@ -744,7 +744,9 @@ def isin(
 
     if test_dt != dt:
         # copy into C-contiguous memory, because the array will be flattened
-        test_buf = dpt.empty_like(test_arr, dtype=dt, order="C")
+        test_buf = dpt.empty_like(
+            test_arr, dtype=dt, order="C", usm_type=res_usm_type
+        )
         ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
             src=test_arr, dst=test_buf, sycl_queue=exec_q, depends=dep_evs
         )
--- a/dpctl/tensor/_set_functions.py
+++ b/dpctl/tensor/_set_functions.py
@@ -698,8 +698,11 @@ def isin(
     du.validate_usm_type(res_usm_type, allow_none=False)
     sycl_dev = exec_q.sycl_device
 
-    if invert not in [True, False]:
-        raise ValueError(f"`invert` must be `True` or `False`, got {invert}")
+    if not isinstance(invert, bool):
+        raise TypeError(
+            "`invert` keyword argument must be of boolean type, "
+            f"got {type(invert)}"
+        )
 
     x_dt = _get_dtype(x, sycl_dev)
     test_dt = _get_dtype(test_elements, sycl_dev)
--- a/dpctl/tensor/_set_functions.py
+++ b/dpctl/tensor/_set_functions.py
@@ -698,6 +698,9 @@ def isin(
     du.validate_usm_type(res_usm_type, allow_none=False)
     sycl_dev = exec_q.sycl_device
 
+    if invert not in [True, False]:
+        raise ValueError(f"`invert` must be `True` or `False`, got {invert}")
+
     x_dt = _get_dtype(x, sycl_dev)
     test_dt = _get_dtype(test_elements, sycl_dev)
     if not all(_validate_dtype(dt) for dt in (x_dt, test_dt)):
--- a/dpctl/tensor/_set_functions.py
+++ b/dpctl/tensor/_set_functions.py
@@ -16,10 +16,12 @@
 
 from typing import NamedTuple
 
+import dpctl
 import dpctl.tensor as dpt
 import dpctl.utils as du
 
 from ._copy_utils import _empty_like_orderK
+from ._scalar_utils import _get_dtype, _get_queue_usm_type, _validate_dtype
 from ._tensor_elementwise_impl import _not_equal, _subtract
 from ._tensor_impl import (
     _copy_usm_ndarray_into_usm_ndarray,
@@ -36,8 +38,10 @@ from ._tensor_sorting_impl import (
     _searchsorted_left,
     _sort_ascending,
 )
+from ._type_utils import _resolve_weak_types_all_py_ints
 
 __all__ = [
+    "isin",
     "unique_values",
     "unique_counts",
     "unique_inverse",
@@ -629,61 +633,101 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
 
 
 def isin(x, test_elements, /, *, assume_unique=False, invert=False):
+    """
+    Tests `x in test_elements` for each element of `x`. Returns a boolean array
+    with the same shape as `x` that is `True` where the element is in
+    `test_elements`, `False` otherwise.
+
+    Args:
+        x (usm_ndarray):
+            input array.
+        test_elements (Union[usm_ndarray, bool, int, float, complex]):
+            elements against which to test each value of `x`.
+            Default: `None`.
+        assume_unique (Optional[bool]):
+            if `True`, the input arrays are both assumed to be unique, which
+            currently has no effect.
+            Default: `False`.
+        invert (Optional[bool]):
+            if `True`, the output results are inverted, i.e., are equivalent to
+            testing `x not in test_elements` for each element of `x`.
+            Default: `False`.
+
+    Returns:
+        usm_ndarray:
+            an array of the inclusion test results. The returned array has a
+            boolean data type and the same shape as `x`.
+    """
     if not isinstance(x, dpt.usm_ndarray):
         raise TypeError(f"Expected dpctl.tensor.usm_ndarray, got {type(x)}")
+    q1, x_usm_type = x.sycl_queue, x.usm_type
+    q2, test_usm_type = _get_queue_usm_type(test_elements)
+    if q2 is None:
+        exec_q = q1
+        res_usm_type = x_usm_type
+    else:
+        exec_q = dpctl.utils.get_execution_queue((q1, q2))
+        if exec_q is None:
+            raise du.ExecutionPlacementError(
+                "Execution placement can not be unambiguously inferred "
+                "from input arguments."
+            )
+        res_usm_type = dpctl.utils.get_coerced_usm_type(
+            (
+                x_usm_type,
+                test_usm_type,
+            )
+        )
+    dpctl.utils.validate_usm_type(res_usm_type, allow_none=False)
+    sycl_dev = exec_q.sycl_device
+
+    x_dt = x.dtype
+    test_dt = _get_dtype(test_elements, sycl_dev)
+    if not _validate_dtype(test_dt):
+        raise ValueError("`test_elements` has unsupported dtype")
+
+    dt = dpt.result_type(
+        *_resolve_weak_types_all_py_ints(x_dt, test_dt, sycl_dev)
+    )
+
+    _manager = du.SequentialOrderManager[exec_q]
+
+    if x_dt != dt:
+        x_buf = _empty_like_orderK(x, dt)
+        dep_evs = _manager.submitted_events
+        ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
+            src=x, dst=x_buf, sycl_queue=exec_q, depends=dep_evs
+        )
+        _manager.add_event_pair(ht_ev, ev)
+    else:
+        x_buf = x
+
     if not isinstance(test_elements, dpt.usm_ndarray):
-        raise TypeError(
-            f"Expected dpctl.tensor.usm_ndarray, got {type(test_elements)}"
+        test_buf = dpt.asarray(test_elements, dtype=dt, sycl_queue=exec_q)
+    elif test_dt != dt:
+        # copy into C-contiguous memory, because the array will be flattened
+        test_buf = dpt.empty_like(test_elements, dt, order="C")
+        dep_evs = _manager.submitted_events
+        ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
+            src=test_elements, dst=test_buf, sycl_queue=exec_q, depends=dep_evs
         )
+        _manager.add_event_pair(ht_ev, ev)
+    else:
+        test_buf = test_elements
 
-    q = du.get_execution_queue([x.sycl_queue, test_elements.sycl_queue])
-    if q is None:
-        raise du.ExecutionPlacementError(
-            "Execution placement can not be unambiguously "
-            "inferred from input arguments."
-        )
+    test_buf = dpt.reshape(test_buf, -1)
+    test_buf = dpt.sort(test_buf)
 
-    x1 = x
-    x2 = dpt.reshape(test_elements, -1)
-
-    x1_dt = x1.dtype
-    x2_dt = x2.dtype
-
-    _manager = du.SequentialOrderManager[q]
-    dep_evs = _manager.submitted_events
-
-    if x1_dt != x2_dt:
-        dt = dpt.result_type(x1, x2)
-        if x1_dt != dt:
-            x1_buf = _empty_like_orderK(x1, dt)
-            dep_evs = _manager.submitted_events
-            ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
-                src=x1, dst=x1_buf, sycl_queue=q, depends=dep_evs
-            )
-            _manager.add_event_pair(ht_ev, ev)
-            x1 = x1_buf
-        if x2_dt != dt:
-            x2_buf = _empty_like_orderK(x2, dt)
-            dep_evs = _manager.submitted_events
-            ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
-                src=x2, dst=x2_buf, sycl_queue=q, depends=dep_evs
-            )
-            _manager.add_event_pair(ht_ev, ev)
-            x2 = x2_buf
-
-    x2 = dpt.sort(x2)
-
-    dst_usm_type = du.get_coerced_usm_type([x1.usm_type, x2.usm_type])
-    dst = _empty_like_orderK(x1, dpt.bool, usm_type=dst_usm_type)
+    dst = _empty_like_orderK(x_buf, dpt.bool, usm_type=res_usm_type)
 
     dep_evs = _manager.submitted_events
     ht_ev, s_ev = _isin(
-        needles=x1,
-        hay=x2,
+        needles=x_buf,
+        hay=test_buf,
         dst=dst,
-        sycl_queue=q,
+        sycl_queue=exec_q,
         invert=invert,
         depends=dep_evs,
     )
     _manager.add_event_pair(ht_ev, s_ev)
-    return dpt.reshape(dst, x.shape)
+    return dst
--- a/dpctl/tensor/_set_functions.py
+++ b/dpctl/tensor/_set_functions.py
@@ -19,6 +19,7 @@ from typing import NamedTuple
 import dpctl.tensor as dpt
 import dpctl.utils as du
 
+from ._copy_utils import _empty_like_orderK
 from ._tensor_elementwise_impl import _not_equal, _subtract
 from ._tensor_impl import (
     _copy_usm_ndarray_into_usm_ndarray,
@@ -31,6 +32,7 @@ from ._tensor_impl import (
 )
 from ._tensor_sorting_impl import (
     _argsort_ascending,
+    _isin,
     _searchsorted_left,
     _sort_ascending,
 )
@@ -624,3 +626,64 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
         inv,
         _counts,
     )
+
+
+def isin(x, test_elements, /, *, assume_unique=False, invert=False):
+    if not isinstance(x, dpt.usm_ndarray):
+        raise TypeError(f"Expected dpctl.tensor.usm_ndarray, got {type(x)}")
+    if not isinstance(test_elements, dpt.usm_ndarray):
+        raise TypeError(
+            f"Expected dpctl.tensor.usm_ndarray, got {type(test_elements)}"
+        )
+
+    q = du.get_execution_queue([x.sycl_queue, test_elements.sycl_queue])
+    if q is None:
+        raise du.ExecutionPlacementError(
+            "Execution placement can not be unambiguously "
+            "inferred from input arguments."
+        )
+
+    x1 = x
+    x2 = dpt.reshape(test_elements, -1)
+
+    x1_dt = x1.dtype
+    x2_dt = x2.dtype
+
+    _manager = du.SequentialOrderManager[q]
+    dep_evs = _manager.submitted_events
+
+    if x1_dt != x2_dt:
+        dt = dpt.result_type(x1, x2)
+        if x1_dt != dt:
+            x1_buf = _empty_like_orderK(x1, dt)
+            dep_evs = _manager.submitted_events
+            ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
+                src=x1, dst=x1_buf, sycl_queue=q, depends=dep_evs
+            )
+            _manager.add_event_pair(ht_ev, ev)
+            x1 = x1_buf
+        if x2_dt != dt:
+            x2_buf = _empty_like_orderK(x2, dt)
+            dep_evs = _manager.submitted_events
+            ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
+                src=x2, dst=x2_buf, sycl_queue=q, depends=dep_evs
+            )
+            _manager.add_event_pair(ht_ev, ev)
+            x2 = x2_buf
+
+    x2 = dpt.sort(x2)
+
+    dst_usm_type = du.get_coerced_usm_type([x1.usm_type, x2.usm_type])
+    dst = _empty_like_orderK(x1, dpt.bool, usm_type=dst_usm_type)
+
+    dep_evs = _manager.submitted_events
+    ht_ev, s_ev = _isin(
+        needles=x1,
+        hay=x2,
+        dst=dst,
+        sycl_queue=q,
+        invert=invert,
+        depends=dep_evs,
+    )
+    _manager.add_event_pair(ht_ev, s_ev)
+    return dpt.reshape(dst, x.shape)
--- a/dpctl/tensor/_set_functions.py
+++ b/dpctl/tensor/_set_functions.py
@@ -646,7 +646,8 @@ def isin(
     *,
     invert: Optional[bool] = False,
 ) -> dpt.usm_ndarray:
-    """
+    """isin(x, test_elements, /, *, invert=False)
+
     Tests `x in test_elements` for each element of `x`. Returns a boolean array
     with the same shape as `x` that is `True` where the element is in
     `test_elements`, `False` otherwise.
--- a/dpctl/tensor/_set_functions.py
+++ b/dpctl/tensor/_set_functions.py
@@ -21,7 +21,12 @@ import dpctl.tensor as dpt
 import dpctl.utils as du
 
 from ._copy_utils import _empty_like_orderK
-from ._scalar_utils import _get_dtype, _get_queue_usm_type, _validate_dtype
+from ._scalar_utils import (
+    _get_dtype,
+    _get_queue_usm_type,
+    _get_shape,
+    _validate_dtype,
+)
 from ._tensor_elementwise_impl import _not_equal, _subtract
 from ._tensor_impl import (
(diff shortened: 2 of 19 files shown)


dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +626,64 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+    if x1_dt != x2_dt:
+        dt = dpt.result_type(x1, x2)
+        if x1_dt != dt:
+            x1_buf = _empty_like_orderK(x1, dt)
+            dep_evs = _manager.submitted_events
+            ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
+                src=x1, dst=x1_buf, sycl_queue=q, depends=dep_evs
+            )
+            _manager.add_event_pair(ht_ev, ev)
+            x1 = x1_buf
```
antonwolfy: it can be `x1 = dpt.astype(dt)` here

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +626,64 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+        )
+
+    x1 = x
+    x2 = dpt.reshape(test_elements, -1)
+
+    x1_dt = x1.dtype
+    x2_dt = x2.dtype
+
+    _manager = du.SequentialOrderManager[q]
+    dep_evs = _manager.submitted_events
```
antonwolfy, line 741: seems unused

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +626,64 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+        )
+
+    q = du.get_execution_queue([x.sycl_queue, test_elements.sycl_queue])
+    if q is None:
+        raise du.ExecutionPlacementError(
+            "Execution placement can not be unambiguously "
+            "inferred from input arguments."
+        )
+
+    x1 = x
```
antonwolfy: Do we need to flatten `x` array also?

dpctl/tensor/_set_functions.py
antonwolfy: or probably the point was to do copying of both arrays in parallel, then we should pass the same `dep_evs` to both copy calls

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +626,64 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+            x2 = x2_buf
+
+    x2 = dpt.sort(x2)
+
+    dst_usm_type = du.get_coerced_usm_type([x1.usm_type, x2.usm_type])
+    dst = _empty_like_orderK(x1, dpt.bool, usm_type=dst_usm_type)
+
+    dep_evs = _manager.submitted_events
+    ht_ev, s_ev = _isin(
+        needles=x1,
```
antonwolfy: It seems the only case when the strided implementation (which assumes slower) will be used when `x1` array is not contiguous (we sort `test_elements` array and no `out` keyword in `isin` function).
Would it make sense to flatten input array `x` and to pass `order` keyword there?

But, it makes sense also to keep strided implementation of `_isin` in case when it might be helpful in implementation of other set functions).

dpctl/tensor/_set_functions.py
ndgrigorian: no, `x` doesn't need to be flattened, just the second argument

dpctl/tensor/_set_functions.py
ndgrigorian: I can experiment and see values of flattening vs. not flattening

but in general, this implementation is going to be changed quite a bit soon, I have some local changes waiting

dpctl/tensor/_set_functions.py
ndgrigorian: yes, good point about `dep_evs`

this logic also needs to be redone—it's taken from `searchsorted` but array API is considering scalar values of `test_elements`

furthermore, we will need logic handling cases of `int64—uint64`, similar to comparisons

dpctl/tensor/_set_functions.py
antonwolfy: then it seems no need to reshape at the end of the function

ndgrigorian: good point, probably left over from an earlier iteration

dpctl/tensor/_set_functions.py
ndgrigorian: turns out that `int64-uint64` case is also not handled in `searchsorted`, including NumPy's implementation

```
In [2]: np.searchsorted(np.arange(np.iinfo(np.int64).max - 10, np.iinfo(np.int64).max, dtype="i8"), np.asarray([0, np.ii
   ...: nfo(np.int64).max + 1000], dtype="u8"))
Out[2]: array([0, 0])
```
our result is the same, and this is likely caused by casting into float

I think I will proceed with this implementation ignoring mixed integral kinds for now, and use a follow-up PR to fix this and searchsorted in general

ndgrigorian:
@antonwolfy 
implementation is updated and now permits Python scalars in second argument

tests still need to be added, but `isin` itself is ready for review

dpctl/tensor/CMakeLists.txt
```diff
@@ -112,6 +112,7 @@ set(_reduction_sources
     ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/reductions/sum.cpp
 )
 set(_sorting_sources
+    ${CMAKE_CURRENT_SOURCE_DIR}/libtensor/source/sorting/isin.cpp
```
antonwolfy, line 115: It doesn't seem relating to sorting routine

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +630,104 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+    Tests `x in test_elements` for each element of `x`. Returns a boolean array
+    with the same shape as `x` that is `True` where the element is in
+    `test_elements`, `False` otherwise.
+
+    Args:
+        x (usm_ndarray):
+            input array.
+        test_elements (Union[usm_ndarray, bool, int, float, complex]):
+            elements against which to test each value of `x`.
+            Default: `None`.
```
antonwolfy: It can't be `None` I assume

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +630,104 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+    else:
+        x_buf = x
+
+    if not isinstance(test_elements, dpt.usm_ndarray):
+        test_buf = dpt.asarray(test_elements, dtype=dt, sycl_queue=exec_q)
+    elif test_dt != dt:
+        # copy into C-contiguous memory, because the array will be flattened
+        test_buf = dpt.empty_like(test_elements, dt, order="C")
+        dep_evs = _manager.submitted_events
+        ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
```
antonwolfy, line 757: This copy and the one at line 698 can be done independently if any.

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +630,104 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+            src=x, dst=x_buf, sycl_queue=exec_q, depends=dep_evs
+        )
+        _manager.add_event_pair(ht_ev, ev)
+    else:
+        x_buf = x
+
+    if not isinstance(test_elements, dpt.usm_ndarray):
+        test_buf = dpt.asarray(test_elements, dtype=dt, sycl_queue=exec_q)
```
antonwolfy: Do we need to pass `usm_type` keyword?

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +630,104 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+            src=test_elements, dst=test_buf, sycl_queue=exec_q, depends=dep_evs
+        )
+        _manager.add_event_pair(ht_ev, ev)
+    else:
+        test_buf = test_elements
+
+    test_buf = dpt.reshape(test_buf, -1)
+    test_buf = dpt.sort(test_buf)
+
+    dst = _empty_like_orderK(x_buf, dpt.bool, usm_type=res_usm_type)
```
antonwolfy: Do we need to keep the same order like used by `x`? Wouldn't it be better to return C-contig array?

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +630,104 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+            )
+        )
+    dpctl.utils.validate_usm_type(res_usm_type, allow_none=False)
+    sycl_dev = exec_q.sycl_device
+
+    x_dt = x.dtype
+    test_dt = _get_dtype(test_elements, sycl_dev)
+    if not _validate_dtype(test_dt):
+        raise ValueError("`test_elements` has unsupported dtype")
+
```
antonwolfy, line 722: Do we need to handle the corner case separately at some point?
```python
    if isinstance(test_elements, dpt.usm_ndarray) and test_elements.size == 0:
        if invert:
            return dpt.ones_like(ar1, dtype=bool, usm_type=res_usm_type)
        else:
            return dpt.zeros_like(ar1, dtype=bool, usm_type=res_usm_type)
```

dpctl/tensor/libtensor/source/sorting/isin.cpp
```diff
@@ -0,0 +1,322 @@
...
+    isin_contig_impl_dispatch_vector[td_ns::num_types];
+
+template <typename fnT, typename argTy> struct IsinContigFactory
+{
+    constexpr IsinContigFactory() {}
+
+    fnT get() const
+    {
+        using dpctl::tensor::kernels::isin_contig_impl;
+        using Compare = typename AscendingSorter<argTy>::type;
```
antonwolfy: Do we need to have `Compare` templated here? Is there any use case possible when another one will be required to be used by `isin` kernel?

dpctl/tensor/libtensor/include/kernels/sorting/isin.hpp
```diff
@@ -0,0 +1,240 @@
...
+                const OutIndexerT &out_indexer_)
+        : invert(invert_), hay_tp(hay_), needles_tp(needles_), out_tp(out_),
+          hay_nelems(hay_nelems_), hay_indexer(hay_indexer_),
+          needles_indexer(needles_indexer_), out_indexer(out_indexer_)
+    {
+    }
+
+    void operator()(sycl::id<1> id) const
+    {
+        const Compare comp{};
```
antonwolfy:
```suggestion
        static constexpr Compare comp{};
```

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +630,104 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+        )
+        _manager.add_event_pair(ht_ev, ev)
+    else:
+        x_buf = x
+
+    if not isinstance(test_elements, dpt.usm_ndarray):
+        test_buf = dpt.asarray(test_elements, dtype=dt, sycl_queue=exec_q)
+    elif test_dt != dt:
+        # copy into C-contiguous memory, because the array will be flattened
+        test_buf = dpt.empty_like(test_elements, dt, order="C")
```
antonwolfy:
```suggestion
        test_buf = dpt.empty_like(test_elements, dtype=dt, order="C")
```

dpctl/tensor/libtensor/source/sorting/isin.cpp
ndgrigorian: It's not strictly necessary, but was done to reduce code duplication—these sorters are defined in `tensor/source`.

I can look at the normal sort implementation to refresh myself on what was done there, but if that isn't sufficient, it may be preferable to template and pass the sorter here as opposed to duplicating in `isin.hpp`

dpctl/tensor/libtensor/include/kernels/sorting/isin.hpp
ndgrigorian: good point!

dpctl/tensor/_set_functions.py
ndgrigorian: good catch, copy-paste error

dpctl/tensor/CMakeLists.txt
ndgrigorian: it uses common utilities with `searchsorted` (i.e., from `rich_comparisons.hpp`) which is why it lives there

ndgrigorian: if the code from `rich_comparisons` gets factored out, I can go ahead and move it elsewhere, I guess to `_tensor_impl` for now

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +638,130 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+                "Execution placement can not be unambiguously inferred "
+                "from input arguments."
+            )
+        res_usm_type = dpctl.utils.get_coerced_usm_type(
+            (
+                x_usm_type,
+                test_usm_type,
+            )
+        )
+    dpctl.utils.validate_usm_type(res_usm_type, allow_none=False)
```
antonwolfy:
```suggestion
    du.validate_usm_type(res_usm_type, allow_none=False)
```

antonwolfy: `import dpctl` [here](https://github.com/IntelPython/dpctl/pull/2098/files#diff-7baf58a822588f3094fcddb77f35b92e447bc55ef87d1fbbaffe18c0a1d440eeR19) looks unused then

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +638,130 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
         inv,
         _counts,
     )
+
+
+def isin(x, test_elements, /, *, invert=False):
```
antonwolfy: Do we need to use typing? (I found it was previously used for unique functions)
```suggestion
def isin(x: Union[dpt.usm_ndarray, int, float, complex, bool], test_elements: Union[dpt.usm_ndarray, int, float, complex, bool], /, *, invert: Optional[bool] = False) -> dpt.usm_ndarray:
```

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +638,130 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+        ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
+            src=x_arr, dst=x_buf, sycl_queue=exec_q, depends=dep_evs
+        )
+        _manager.add_event_pair(ht_ev, ev)
+    else:
+        x_buf = x_arr
+
+    if test_dt != dt:
+        # copy into C-contiguous memory, because the array will be flattened
+        test_buf = dpt.empty_like(test_arr, dtype=dt, order="C")
```
antonwolfy: Would it make sense to pass `usm_type` here?
```suggestion
        test_buf = dpt.empty_like(test_arr, dtype=dt, order="C", usm_type=res_usm_type)
```

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +638,130 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+            return dpt.zeros(
+                x_sh, dtype=dpt.bool, usm_type=res_usm_type, sycl_queue=exec_q
+            )
+
+    dt1, dt2 = _resolve_weak_types_all_py_ints(x_dt, test_dt, sycl_dev)
+    dt = _to_device_supported_dtype(dpt.result_type(dt1, dt2), sycl_dev)
+
+    if not isinstance(x, dpt.usm_ndarray):
+        x_arr = dpt.asarray(
+            x, dtype=dt1, usm_type=res_usm_type, sycl_queue=exec_q
```
antonwolfy, line 728: Do we need to cast here to result dtype to avoid unnecessary copy [below](https://github.com/IntelPython/dpctl/pull/2098/files#diff-7baf58a822588f3094fcddb77f35b92e447bc55ef87d1fbbaffe18c0a1d440eeR733)?
```suggestion
            x, dtype=dt, usm_type=res_usm_type, sycl_queue=exec_q
```

antonwolfy: The same with `test_arr` casting [here](https://github.com/IntelPython/dpctl/pull/2098/files#diff-7baf58a822588f3094fcddb77f35b92e447bc55ef87d1fbbaffe18c0a1d440eeR722)

antonwolfy: And probably then it would sense to combine the checks, like:
```python
    if not isinstance(x, dpt.usm_ndarray):
        x_buf = dpt.asarray(
            x, dtype=dt, usm_type=res_usm_type, sycl_queue=exec_q
        )
    elif x_dt != dt:
        x_buf = _empty_like_orderK(x, dt, res_usm_type, sycl_dev)
        ht_ev, ev = _copy_usm_ndarray_into_usm_ndarray(
            src=x, dst=x_buf, sycl_queue=exec_q, depends=dep_evs
        )
        _manager.add_event_pair(ht_ev, ev)
    else:
        x_buf = x
```

dpctl/tests/test_tensor_isin.py
```diff
@@ -0,0 +1,168 @@
...
+    assert res3.dtype == dpt.bool
+    assert not dpt.all(res3)
+
+    res4 = dpt.isin(x, test, invert=True)
+    assert isinstance(res4, dpt.usm_ndarray)
+    assert res4.shape == x.shape
+    assert res4.dtype == dpt.bool
+    assert dpt.all(res4)
+
+
```
antonwolfy, line 209: I believe it would make sense to test `nan` and `±0` values, per documented behavior in Notes of array-api spec,

dpctl/tests/test_tensor_isin.py
```diff
@@ -0,0 +1,168 @@
...
+def test_isin_strided_bool():
+    dt = dpt.bool
+    n, m = 100, 20
+    x = dpt.ones((n, m), dtype=dt, order="F")
+    x[:, ::2] = False
+    test = dpt.zeros((), dtype=dt)
+    r1 = dpt.isin(x, test)
+    assert dpt.all(r1[:, ::2])
+    assert not dpt.all(r1[:, 1::2])
+    assert r1.shape == x.shape
```
antonwolfy: It might be helpful to check the result is contiguous 
```suggestion
    assert r1.shape == x.shape
    assert r1.flags.c_contiguous
```

dpctl/tests/test_tensor_isin.py
```diff
@@ -0,0 +1,168 @@
...
+    r2 = dpt.isin(x, test, invert=True)
+    assert not dpt.all(r2[:, ::2])
+    assert dpt.all(r2[:, 1::2])
+    assert r2.shape == x.shape
+
+
+def test_isin_strided_bool():
+    dt = dpt.bool
+    n, m = 100, 20
+    x = dpt.ones((n, m), dtype=dt, order="F")
```
antonwolfy: Would it make sense to consider completely non-contiguous use case?
```suggestion
    x = dpt.ones((n, m), dtype=dt, order="F")
    x = x[:, ::2]
```

dpctl/tests/test_tensor_isin.py
```diff
@@ -0,0 +1,168 @@
...
+
+import pytest
+
+import dpctl.tensor as dpt
+from dpctl.tests.helper import get_queue_or_skip, skip_if_dtype_not_supported
+from dpctl.utils import ExecutionPlacementError
+
+
+@pytest.mark.parametrize(
+    "dtype",
```
antonwolfy: What is about case when inputs have different dtypes and casting is required?

dpctl/tensor/_set_functions.py
ndgrigorian: may as well, we're pretty inconsistent throughout dpctl with this in general, it would be good someday to go back and add type hints everywhere

dpctl/tests/test_tensor_isin.py
ndgrigorian: yes, this was the intention of the test but it got lost at some point 🤔 I'll rewrite it

dpctl/tensor/_set_functions.py
ndgrigorian: I've left it this way because this is how element-wise functions handle it, as well: scalars are put first into the appropriate array type, then the array is cast into another type for computation

I'm not sure if it's strictly necessary, but may avoid some edge cases producing incorrect results

dpctl/tensor/_sorting.py
```diff
@@ -85,6 +85,8 @@ def sort(x, /, *, axis=-1, descending=False, stable=True, kind=None):
         return dpt.copy(x, order="C")
     else:
         axis = normalize_axis_index(axis, ndim=nd, msg_prefix="axis")
+    if x.size == 1:
```
antonwolfy: It seems already covered by #2106. Do we need to have the fix on python side as well?

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +637,138 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
         inv,
         _counts,
     )
+
+
+def isin(
```
antonwolfy, line 642: The new function is missing in the rendered documentation

dpctl/tests/test_tensor_isin.py
```diff
@@ -0,0 +1,258 @@
...
+    n, m = 100, 20
+    x = dpt.zeros((n, m), dtype=dtype, order="F", sycl_queue=q)
+    x[:, ::2] = dpt.arange(1, (m / 2) + 1, dtype=dtype, sycl_queue=q)
+    x_s = x[:, ::2]
+    test = dpt.arange(1, (m / 2), dtype=dtype, sycl_queue=q)
+    r1 = dpt.isin(x_s, test)
+    assert dpt.all(r1[:, :-1])
+    assert not dpt.any(r1[:, -1])
+    assert not dpt.any(x[:, 1::2])
+    assert r1.shape == x_s.shape
```
antonwolfy, line 113: Would it make sense to check `c_contiguous` flag here also?

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +637,138 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+
+    Returns:
+        usm_ndarray:
+            an array of the inclusion test results. The returned array has a
+            boolean data type and the same shape as `x`.
+    """
+    q1, x_usm_type = _get_queue_usm_type(x)
+    q2, test_usm_type = _get_queue_usm_type(test_elements)
+    if q1 is None and q2 is None:
+        raise du.ExecutionPlacementError(
```
antonwolfy, line 673: That isn't covered by any test

dpctl/tensor/_sorting.py
ndgrigorian: No, it's a temporary fix to keep coverage builds from failing, when #2106 is merged I'm going to drop the commit

dpctl/tensor/_set_functions.py
ndgrigorian: It's covered by `test_isin_validation` but the later `ExecutionPlacementError` is not covered per the coverage, even though there's a test for it too

possibly just a bug in the coverage but I'll check again

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +637,138 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+
+def isin(
+    x: Union[dpt.usm_ndarray, int, float, complex, bool],
+    test_elements: Union[dpt.usm_ndarray, int, float, complex, bool],
+    /,
+    *,
+    invert: Optional[bool] = False,
+) -> dpt.usm_ndarray:
+    """
+    Tests `x in test_elements` for each element of `x`. Returns a boolean array
```
antonwolfy: It seems the function signature needs to be duplicated in the docstring comment for better formatting
```suggestion
    """isin(x, test_elements, /, *, invert=False)

    Tests `x in test_elements` for each element of `x`. Returns a boolean array
```

dpctl/tensor/libtensor/source/sorting/isin.cpp
ndgrigorian: I've factored this keyword out

dpctl/tensor/CMakeLists.txt
ndgrigorian: I think we can leave it in sorting for the time being and consider a follow-up PR to reorganize some of the sub-modules (for example, `where` and `clip` exist in the `tensor_ctors` sub-module, but neither is a constructor or needed for use by constructors...)

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +637,139 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+        x_buf, dtype=dpt.bool, usm_type=res_usm_type, order="C"
+    )
+
+    dep_evs = _manager.submitted_events
+    ht_ev, s_ev = _isin(
+        needles=x_buf,
+        hay=test_buf,
+        dst=dst,
+        sycl_queue=exec_q,
+        invert=invert,
```
antonwolfy, line 777: There is no validation on input type of `invert` keyword. We expect to have it as a bool, do we need to check that?

dpctl/tensor/_sorting.py
antonwolfy: The PR has been merged. Do we need to undo the change?

dpctl/tensor/libtensor/include/utils/rich_comparisons.hpp
```diff
@@ -31,7 +31,7 @@ namespace dpctl
 {
 namespace tensor
 {
-namespace py_internal
+namespace rich_comparisons
```
antonwolfy, line 37: Seems missing `#include <cmath>` above, which required for `std::isnan`. And `#include <complex>` is needed by `std::real`.
But it can be done in the follow up PR.

dpctl/tensor/libtensor/include/kernels/sorting/isin.hpp
```diff
@@ -0,0 +1,242 @@
...
+        // position of the needle_v in the hay array
+        std::size_t pos{};
+
+        static constexpr std::size_t zero(0);
+        // search in hay in left-closed interval, give `pos` such that
+        // hay[pos - 1] < needle_v <= hay[pos]
+
+        // lower_bound returns the first pos such that bool(hay[pos] <
+        // needle_v) is false, i.e. needle_v <= hay[pos]
+        pos = static_cast<std::size_t>(
```
antonwolfy: Why do we need type casting here? `search_sorted_detail::lower_bound_indexed_impl` returns `std::size_t`.

dpctl/tensor/libtensor/include/kernels/sorting/isin.hpp
```diff
@@ -0,0 +1,242 @@
...
+///
+/// \file
+/// This file defines kernels for tensor membership operations.
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
```
antonwolfy: Seems unused includes:
```suggestion
#include <cstddef>
```

dpctl/tensor/_set_functions.py
ndgrigorian: in general we don't, since Python `bool` operator is usually sufficient, but in this case we are passing it to pybind11, which is a bit less predictable. I'll make sure behavior is defined but after that I'm fine with it

dpctl/tensor/_sorting.py
ndgrigorian: yes, good point

dpctl/tensor/libtensor/include/utils/rich_comparisons.hpp
ndgrigorian: don't see a reason not to add it in here

dpctl/tensor/libtensor/include/kernels/sorting/isin.hpp
ndgrigorian: I believe it was a generalized index type at first. Since it isn't anymore, it's an artifact, I'll remove it.

dpctl/tensor/_set_functions.py
ndgrigorian: added validation

dpctl/tensor/_set_functions.py
```diff
@@ -624,3 +637,142 @@ def unique_all(x: dpt.usm_ndarray) -> UniqueAllResult:
...
+        res_usm_type = du.get_coerced_usm_type(
+            (
+                x_usm_type,
+                test_usm_type,
+            )
+        )
+    du.validate_usm_type(res_usm_type, allow_none=False)
+    sycl_dev = exec_q.sycl_device
+
+    if invert not in [True, False]:
```
antonwolfy: Would it be safer to check for the type instead?
```suggestion
    if not isinstance(invert, bool):
```
While the current check is much less wicker, i.e. it accepts `numpy.array([1]) in [True, False]` and results `True`.
In that case it does not cause any pybind11 error, but I wonder if there is any value which might.

ndgrigorian: I wanted to preserve this behavior, but I guess expecting `bool(x)` for some size-1 array `x` is fine as well for a user, and it can be a little stricter since, generally, NumPy and CuPy are a bit more lenient.

antonwolfy: In that case it'd be probably better to make it explicit, like:
```python
    if not isinstance(invert, bool):
        invert = bool(invert)
```

antonwolfy (approved):

antonwolfy (approved):

dpctl/tensor/_set_functions.py
ndgrigorian: I recall flattening seemed to slow down the implementation a bit, I will check again after merging and add details

PR #2099, opened 2025-06-06
Update HIP target handling using alias

vlad-perevezentsev:
This PR suggests simplifying `DPCTL_TARGET_HIP` handling by using target aliases `amd_gpu_<arch>` and removing support for `amdgcn-amd-amdhsa` target.
Also updates `docs` to clarify how to determine and use the correct AMD GPU architecture.


ndgrigorian (dismissed):
This LGTM overall

After this and the CUDA architecture update are merged, I think allowing setting CUDA and HIP architecture as arguments to `scripts/build_locally.py` script would be convenient

ndgrigorian (approved):

PR #2106, opened 2025-06-18
Fix sort for size 1 axes

ndgrigorian:
Some sorting algorithms used (in particular `radix_sort`) assert that elements to be sorted must be greater than 1, and this can cause failures in debug builds, like coverage

This PR proposes a fix: copy in sort and memset to 0 in argsort when `sort_nelems` is 1


ndgrigorian:
@antonwolfy 
I've added assertions that `iter_nelems` is the count of elements in the destination array, for safety

antonwolfy (approved):

PR #2109, opened 2025-06-20
Extend build_locally.py with `--target-cuda` and `--target-hip` arguments

vlad-perevezentsev:
This PR suggests extending `build_locally.py` script by adding `--target-cuda` and `--target-hip` arguments
to simplify configuration of CUDA and AMD targets without manually passing CMake options.

```
python scripts/build_locally.py --target-cuda --target-hip=gfx1030
```

Also updates the documentation to reflect the new changes

### DIFF
--- a/docs/doc_sources/beginners_guides/installation.rst
+++ b/docs/doc_sources/beginners_guides/installation.rst
@@ -192,12 +192,11 @@ Compute Capabilities can be found in the official
 AMD build
 ~~~~~~~~~
 
-``dpctl`` can be built for AMD devices using the ``DPCTL_TARGET_HIP`` CMake option,
-which requires specifying a compute architecture string:
+``dpctl`` can be built for AMD devices using the  ``--target-hip`` argument.
 
 .. code-block:: bash
 
-    python scripts/build_locally.py --verbose --cmake-opts="-DDPCTL_TARGET_HIP=<arch>"
+    python scripts/build_locally.py --verbose --target-hip=<arch>
 
 Note that the `oneAPI for AMD GPUs` plugin requires the architecture be specified and only
 one architecture can be specified at a time.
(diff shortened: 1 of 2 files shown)


docs/doc_sources/beginners_guides/installation.rst
```diff
@@ -166,19 +166,19 @@ A full list of available SYCL alias targets is available in the
...
-    python scripts/build_locally.py --verbose --cmake-opts="-DDPCTL_TARGET_CUDA=sm_80"
+    python scripts/build_locally.py --verbose --target-cuda=sm_80
 
-To use the default architecture (``sm_50``),
-set ``DPCTL_TARGET_CUDA`` to a value such as ``ON``, ``TRUE``, ``YES``, ``Y``, or ``1``:
+To use the default architecture (``sm_50``), omit the value:
 
 .. code-block:: bash
 
-    python scripts/build_locally.py --verbose --cmake-opts="-DDPCTL_TARGET_CUDA=ON"
```
ndgrigorian: I think we should keep this option listed, even if the other way is simpler now

vlad-perevezentsev: done
thank you

scripts/build_locally.py
```diff
@@ -65,6 +67,23 @@ def run(
...
         cmake_args += cmake_opts.split()
+    if target_cuda is not None:
+        if not target_cuda.strip():
+            raise ValueError(
+                "--target-cuda can not be an empty string. "
+                "Use --target-cuda=<arch> or --target-cuda"
+            )
+        cmake_args += [
+            f"-DDPCTL_TARGET_CUDA={target_cuda}",
+        ]
```
ndgrigorian, line 84: what would happen if `-DDPCTL_TARGET_CUDA={target_cuda}` and `--target-cuda` both happen to be specified?

vlad-perevezentsev: @ndgrigorian good question!
When both are passed, CMake usually uses the first value.
To avoid confusion, I have added a check that prevents using both `--target-cuda/--target-hip` and the matching CMake option in `--cmake-opts`
Thank you

PR #2110, opened 2025-06-20
Add Python 3.13 to metadata

ndgrigorian:
This PR proposes adding Python 3.13 to the pyproject.toml metadata

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ classifiers = [
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
   "Programming Language :: Python :: Implementation :: CPython",
   "Topic :: Software Development",
   "Topic :: Scientific/Engineering",


vtavana (approved):

pyproject.toml
```diff
@@ -28,6 +28,7 @@ classifiers = [
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
```
antonwolfy, line 31: Do we need to have a upper limit on python version [below](https://github.com/IntelPython/dpctl/pull/2110/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R61)?
```toml
requires-python = ">=3.9,<3.14"
```

ndgrigorian: I think leaving the upper-bound undefined is also fine (for example, NumPy does the same) since we don't have reason to believe there would be compatibility issues with 3.14.

It also means we can test with the newer version without modification.

antonwolfy (approved):
LGTM

PR #2120, opened 2025-07-08
Add a fast path to copy to numpy.ndarray

antonwolfy:
This PR adds a fast path to copy the USM data to `numpy.ndarray`.

In case when the USM data has the zero size, there is no need to synchronize the data and to perform the copy. It will be enough to call `numpy.ndarray` constructor with appropriate `shape` and `dtype` keywords.


ndgrigorian (commented):
I wonder if we should still submit `q.wait()` for the sake of consistency. I doubt it would impact any workflows, but if one relies on this function to be a blocking call, it might be a surprise that it doesn't, even in this special case

perhaps we should also consider implementing asynchronous copies to NumPy and the blocking keyword, like CuPy does

Issue #2121, opened 2025-07-08
`dpt.astype` returns wrong result on boolean view of numpy.ndarray


antonwolfy:
Here is example with unexpected behavior:
```python
import dpctl, dpctl.tensor as dpt, numpy

dpctl.__version__
# Out: '0.21.0dev0+49.g5be8ef909c'

a = numpy.array([0, 3, 1, 2, 0, 1], dtype='u1')
a_view = a.view('?')

# create a view from numpy.ndarray
b = numpy.array(a_view)
b
# Out: array([False,  True,  True,  True, False,  True])

dpt_b = dpt.asarray(a_view)
dpt_b  # the same as b
# Out: usm_ndarray([False,  True,  True,  True, False,  True])

# cast to integer dtype
numpy.astype(b, int)
# Out: array([0, 1, 1, 1, 0, 1])

# the result is unexpected, since True and False have to casted to 0 and 1 only
dpt.astype(dpt_b, int)
# Out: usm_ndarray([0, 3, 1, 2, 0, 1])
```

sycloid:
@ndgrigorian @antonwolfy 

Notice that C++ also disagrees with NumPy here:

```cpp
// clang++ gh_2121.cpp -o tmp
#include <iostream>
#include <cstdint>
#include <cstring>

#include <string>
#include <vector>

int main(void) {
  std::vector<std::uint8_t> orig{{0, 3, 1, 2, 0, 1}};
  std::vector<std::uint8_t> conv(orig.size());

  std::cout << "Original : ";
  for(std::size_t i = 0; i < orig.size(); ++i) {
     std::cout << std::to_string(orig[i]) << " ";
  }
  std::cout << "\n";

  bool *p = reinterpret_cast<bool *>(orig.data());

  for(std::size_t i=0; i < orig.size(); ++i) {
      conv[i] = (p[i]) ? 1 : 0;
  }

  std::cout << "Converted: ";
  for(std::size_t i = 0; i < orig.size(); ++i) {
        std::cout << std::to_string(conv[i]) << " ";
  }
  std::cout << "\n";

  return 0;
}
``` 

```
$ clang++ gh_2121.cpp -o tmp
$ ./tmp
Original : 0 3 1 2 0 1
Converted: 0 1 1 0 0 1
```

To implement NumPy's behavior in C++ one needs to always work with bool by reference (otherwise a copy is made standardizing the underlying bit-set), bit-cast `bool` to `std::uint8_t` and convert as needed.

PR #2122, opened 2025-07-09
Fix for gh-2121

ndgrigorian:
This PR proposes a fix for gh-2121, which is caused by underlying memory of a boolean array not necessarily representing only 0s and 1s, in particular when it is a view from a non-boolean array

This PR adds a path to astype that guarantees `astype(a, "?")` will give an output array of only 0 and 1

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -739,6 +739,9 @@ def astype(
             order=copy_order,
             buffer_ctor_kwargs={"queue": usm_ary.sycl_queue},
         )
+    # see #2121
+    if ary_dtype == dpt.bool:
+        usm_ary = dpt.not_equal(usm_ary, 0, order=order)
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_copy_utils.py
```diff
@@ -739,6 +739,9 @@ def astype(
             order=copy_order,
             buffer_ctor_kwargs={"queue": usm_ary.sycl_queue},
         )
+    # see #2121
+    if ary_dtype == dpt.bool:
+        usm_ary = dpt.not_equal(usm_ary, 0, order=order)
```
antonwolfy: The issue is resolved. Thank you @ndgrigorian!
The only question is about `order` keyword, why do we pass it as `order=order` and would it be more efficient to use `order=copy_order` instead to be aligned with order of `R` array?

ndgrigorian: it was passed this way because of a typo, good catch

PR #2123, opened 2025-07-14
Allow type casting of zero-sized array to any dtype

antonwolfy:
This PR allows type casting zero-sized `usm_ndarray` to any dtype (including larger ones).

### DIFF
--- a/dpctl/tensor/_usmarray.pyx
+++ b/dpctl/tensor/_usmarray.pyx
@@ -382,7 +382,7 @@ cdef class usm_ndarray:
         else:
             self._cleanup()
             raise ValueError("buffer='{}' was not understood.".format(buffer))
-        if (_offset + ary_min_displacement < 0 or
+        if shape_to_elem_count(nd, shape_ptr) > 0 and (_offset + ary_min_displacement < 0 or
            (_offset + ary_max_displacement + 1) * itemsize > _buffer.nbytes):
             self._cleanup()
             raise ValueError(("buffer='{}' can not accommodate "
--- a/dpctl/tensor/_usmarray.pyx
+++ b/dpctl/tensor/_usmarray.pyx
@@ -382,8 +382,9 @@ cdef class usm_ndarray:
         else:
             self._cleanup()
             raise ValueError("buffer='{}' was not understood.".format(buffer))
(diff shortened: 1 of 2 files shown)


ndgrigorian (commented):
an interesting consequence of this is as follows:
```

In [2]: x1 = dpt.ones((0,), dtype="i4")

In [3]: x1.usm_data.nbytes
Out[3]: 4

In [4]: x2 = dpt.ones((0,), dtype="i8")

In [5]: x2.usm_data.nbytes
Out[5]: 8

In [6]: x2 = dpt.usm_ndarray(x1.shape, dtype="i8", buffer=x1)

In [7]: x2.usm_data.nbytes
Out[7]: 4
```
the underlying representation of a size-0 array is actually an array with a single element, which is generally in the size of the original data type

this is mostly an implementation detail, though

antonwolfy:
> hard to say—underlying tensor `usm_data` is always a valid USM pointer, which is the reason for this discrepancy. It's possible that the memory object could be adapted for this special case, but it may not be very important either. In general, working with the underlying memory of a size-0 array is ill-advised.

As per my understanding, it's also a valid pointer in NumPy. It [allocates](https://github.com/numpy/numpy/blob/ea869ecd6fa96c4c8a2ab5fc9c6dc015ea53f833/numpy/_core/src/multiarray/ctors.c#L863) a memory of size 1 in case of zero-space arrays and the same at the [deallocation](https://github.com/numpy/numpy/blob/ea869ecd6fa96c4c8a2ab5fc9c6dc015ea53f833/numpy/_core/src/multiarray/arrayobject.c#L442)). But in that case it returns `nbytes=0`.

My previous comment was mainly if we need to return `nbytes=0` also in that case.

ndgrigorian:
> As per my understanding, it's also a valid pointer in NumPy. It [allocates](https://github.com/numpy/numpy/blob/ea869ecd6fa96c4c8a2ab5fc9c6dc015ea53f833/numpy/_core/src/multiarray/ctors.c#L863) a memory of size 1 in case of zero-space arrays and the same at the [deallocation](https://github.com/numpy/numpy/blob/ea869ecd6fa96c4c8a2ab5fc9c6dc015ea53f833/numpy/_core/src/multiarray/arrayobject.c#L442). But in that case it returns `nbytes=0`.

Interesting, I believe CuPy does not, as far as I can tell its memory pointer object is wrapping a nullptr in this case.

PR #2128, opened 2025-07-24
Accept NumPy arrays in advanced indexing

### LINKED ISSUE #2053: Request: subsetting with numpy arrays/masks
ref https://github.com/uxlfoundation/scikit-learn-intelex/discussions/2350

Currently, it's not possible to subset dpctl or dpnp arrays with a numpy array of integers:
```python
import numpy as np
import dpnp
X = np.arange(16).reshape((4,4))
dpnp.array(X)[np.arange(2)]
```
```
IndexError                                Traceback (most recent call last)
Cell In[9], line 4
      2 import dpnp
      3 X = np.arange(16).reshape((4,4))
----> 4 dpnp.array(X)[np.arange(2)]

File /localdisk2/mkl/dcortes/miniforge3/envs/icxconda/lib/python3.12/site-packages/dpnp/dpnp_array.py:355, in dpnp_array.__getitem__(self, key)
    352 """Return ``self[key]``."""
    353 key = _get_unwrapped_index_key(key)
--> 355 item = self._array_obj.__getitem__(key)
    356 return dpnp_array._create_from_usm_ndarray(item)

File dpctl/tensor/_usmarray.pyx:937, in dpctl.tensor._usmarray.usm_ndarray.__getitem__()

File dpctl/tensor/_slicing.pxi:300, in dpctl.tensor._usmarray._basic_slice_meta()

IndexError: Only integers, slices (`:`), ellipsis (`...`), dpctl.tensor.newaxis (`None`) and integer and boolean arrays are valid indices.
```

Allowing this type of indexing, even if not very efficiently, would be very helpful for [scikit-learn-intelex](https://github.com/uxlfoundation/scikit-learn-intelex).

This library `scikit-learn-intelex` is meant to be compatible with the library `scikit-learn`, by being able to use classes from scikit-learn and scikit-learn-intelex interchangeably, but with `scikit-learn-intelex` additionally being able to work with `dpctl` arrays and offering algorithms that run on GPU.

Internally, library scikit-learn (on which scikit-learn-intelex relies) has many functionalities which simply create arrays of indices to subset a larger array, which currently do not work with `dpctl`/`dpnp` inputs, as the integer indices are from numpy and the library is not sycl-aware.

Enabling subsetting of dpctl arrays with numpy integer and boolean arrays (as scikit-learn does internally) would immediately enable a lot of useful GPU features on `scikit-learn-intelex`, such as tuning parameters of machine learning models on GPU, calculating cross-validated metrics, among many others, which currently are not possible.

ndgrigorian:
This PR proposes allowing NumPy arrays when indexing `usm_ndarray`

This can be useful where attempting to use `usm_ndarray` in a library which generates indices with NumPy, and which expects to only handle devices implicitly through tensor operations (i.e., a library which can take an arbitrary tensor input, but generates indices with `numpy.random` and expects these indices to work on the input tensor)

Both boolean and integer advanced indices can accept NumPy arrays with these changes 

Closes #2053 

### DIFF
--- a/dpctl/tensor/_copy_utils.py
+++ b/dpctl/tensor/_copy_utils.py
@@ -756,20 +756,28 @@ def _extract_impl(ary, ary_mask, axis=0):
         raise TypeError(
             f"Expecting type dpctl.tensor.usm_ndarray, got {type(ary)}"
         )
-    if not isinstance(ary_mask, dpt.usm_ndarray):
-        raise TypeError(
-            f"Expecting type dpctl.tensor.usm_ndarray, got {type(ary_mask)}"
+    if isinstance(ary_mask, dpt.usm_ndarray):
+        dst_usm_type = dpctl.utils.get_coerced_usm_type(
+            (ary.usm_type, ary_mask.usm_type)
         )
-    dst_usm_type = dpctl.utils.get_coerced_usm_type(
-        (ary.usm_type, ary_mask.usm_type)
-    )
-    exec_q = dpctl.utils.get_execution_queue(
-        (ary.sycl_queue, ary_mask.sycl_queue)
-    )
-    if exec_q is None:
-        raise dpctl.utils.ExecutionPlacementError(
-            "arrays have different associated queues. "
-            "Use `y.to_device(x.device)` to migrate."
+        exec_q = dpctl.utils.get_execution_queue(
+            (ary.sycl_queue, ary_mask.sycl_queue)
+        )
+        if exec_q is None:
+            raise dpctl.utils.ExecutionPlacementError(
+                "arrays have different associated queues. "
+                "Use `y.to_device(x.device)` to migrate."
+            )
+    elif isinstance(ary_mask, np.ndarray):
+        dst_usm_type = ary.usm_type
+        exec_q = ary.sycl_queue
+        ary_mask = dpt.asarray(
+            ary_mask, usm_type=dst_usm_type, sycl_queue=exec_q
+        )
+    else:
+        raise TypeError(
+            "Expecting type dpctl.tensor.usm_ndarray or numpy.ndarray, got "
+            f"{type(ary_mask)}"
         )
     ary_nd = ary.ndim
     pp = normalize_axis_index(operator.index(axis), ary_nd)
@@ -839,31 +847,32 @@ def _nonzero_impl(ary):
 
 def _validate_indices(inds, queue_list, usm_type_list):
     """
-    Utility for validating indices are usm_ndarray of integral dtype or Python
-    integers. At least one must be an array.
+    Utility for validating indices are NumPy ndarray or usm_ndarray of integral
+    dtype or Python integers. At least one must be an array.
 
     For each array, the queue and usm type are appended to `queue_list` and
     `usm_type_list`, respectively.
     """
-    any_usmarray = False
+    any_array = False
     for ind in inds:
-        if isinstance(ind, dpt.usm_ndarray):
-            any_usmarray = True
+        if isinstance(ind, (np.ndarray, dpt.usm_ndarray)):
+            any_array = True
             if ind.dtype.kind not in "ui":
                 raise IndexError(
                     "arrays used as indices must be of integer (or boolean) "
                     "type"
                 )
-            queue_list.append(ind.sycl_queue)
-            usm_type_list.append(ind.usm_type)
+            if isinstance(ind, dpt.usm_ndarray):
+                queue_list.append(ind.sycl_queue)
+                usm_type_list.append(ind.usm_type)
         elif not isinstance(ind, Integral):
             raise TypeError(
-                "all elements of `ind` expected to be usm_ndarrays "
-                f"or integers, found {type(ind)}"
+                "all elements of `ind` expected to be usm_ndarrays, "
+                f"NumPy arrays, or integers, found {type(ind)}"
             )
-    if not any_usmarray:
+    if not any_array:
         raise TypeError(
-            "at least one element of `inds` expected to be a usm_ndarray"
+            "at least one element of `inds` expected to be an array"
         )
     return inds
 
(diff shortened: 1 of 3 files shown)


dpctl/tensor/_copy_utils.py
```diff
@@ -756,20 +756,28 @@ def _extract_impl(ary, ary_mask, axis=0):
...
-            "Use `y.to_device(x.device)` to migrate."
+        exec_q = dpctl.utils.get_execution_queue(
+            (ary.sycl_queue, ary_mask.sycl_queue)
+        )
+        if exec_q is None:
+            raise dpctl.utils.ExecutionPlacementError(
+                "arrays have different associated queues. "
+                "Use `y.to_device(x.device)` to migrate."
+            )
+    elif isinstance(ary_mask, np.ndarray):
```
antonwolfy, line 771: `_extract_impl(arr, condition)` is also called inside `dpt.extract`. Should we allow passing `np.ndarray` there for `condition` array also?

dpctl/tensor/_copy_utils.py
```diff
@@ -756,20 +756,28 @@ def _extract_impl(ary, ary_mask, axis=0):
...
-        (ary.sycl_queue, ary_mask.sycl_queue)
-    )
-    if exec_q is None:
-        raise dpctl.utils.ExecutionPlacementError(
-            "arrays have different associated queues. "
-            "Use `y.to_device(x.device)` to migrate."
+        exec_q = dpctl.utils.get_execution_queue(
+            (ary.sycl_queue, ary_mask.sycl_queue)
+        )
+        if exec_q is None:
```
antonwolfy, line 766: I found a suspicious behavior:
```python
q1 = dpctl.SyclQueue(property="in_order")
q2 = dpctl.SyclQueue(property="enable_profiling")

a = dpt.ones(10, sycl_queue=q1)
mask = dpt.zeros(10, dtype='?', sycl_queue=q2)

# raised error as expected:
a[mask]
ExecutionPlacementError                   Traceback (most recent call last)
Cell In[30], line 1
----> 1 a[mask]

File dpctl/tensor/_usmarray.pyx:999, in dpctl.tensor._usmarray.usm_ndarray.__getitem__()

File /localdisk/work/antonvol/soft/miniforge3/envs/dpnp_dev/lib/python3.12/site-packages/dpctl/tensor/_copy_utils.py:767, in _extract_impl(ary, ary_mask, axis)
    763     exec_q = dpctl.utils.get_execution_queue(
    764         (ary.sycl_queue, ary_mask.sycl_queue)
    765     )
    766     if exec_q is None:
--> 767         raise dpctl.utils.ExecutionPlacementError(
    768             "arrays have different associated queues. "
    769             "Use `y.to_device(x.device)` to migrate."
    770         )
    771 elif isinstance(ary_mask, np.ndarray):
    772     dst_usm_type = ary.usm_type

ExecutionPlacementError: arrays have different associated queues. Use `y.to_device(x.device)` to migrate.

# no error for 0-d array:
mask = dpt.asarray(1, dtype='?', sycl_queue=q2)
a[mask]
# Out: usm_ndarray([[1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]])
```
There will be no `_extract_impl` called, but seems the check is missing in another implementation branch inside `__getitem__`.

dpctl/tensor/_copy_utils.py
```diff
@@ -1048,23 +1066,13 @@ def _put_multi_index(ary, inds, p, vals, mode=0):
...
 
-    _validate_indices(inds, queues_, usm_types_)
+    exec_q, vals_usm_type = _get_indices_queue_usm_type(
+        inds, ary.sycl_queue, ary.usm_type
+    )
 
-    vals_usm_type = dpctl.utils.get_coerced_usm_type(usm_types_)
-    exec_q = dpctl.utils.get_execution_queue(queues_)
     if exec_q is not None:
         if not isinstance(vals, dpt.usm_ndarray):
```
antonwolfy, line 1100: Is there missing coercion of usm_type for an input value?
```python
        else:
            exec_q = dpctl.utils.get_execution_queue((exec_q, vals.sycl_queue))
            vals_usm_type = dpctl.utils.get_coerced_usm_type((vals_usm_type, vals.usm_type))
```

dpctl/tests/test_usm_ndarray_indexing.py
```diff
@@ -440,6 +440,28 @@ def test_advanced_slice16():
...
 
 
+def test_integer_indexing_numpy_array():
+    q = get_queue_or_skip()
+    ii = np.asarray([1, 2])
+    x = dpt.arange(10, dtype="i4", sycl_queue=q)
+    y = x[ii]
+    assert isinstance(y, dpt.usm_ndarray)
+    assert y.shape == ii.shape
+    assert dpt.all(dpt.asarray(ii, sycl_queue=q) == y)
```
antonwolfy: Would it be better to use
```suggestion
    assert dpt.all(x[1:3] == y)
```
since currently it works only because there is `dpt.arange(10, ...)` used to allocate data

dpctl/tensor/_copy_utils.py
ndgrigorian: in `dpt.extract`? I think we it's best if we keep the function calls themselves the same

dpctl/tensor/_copy_utils.py
ndgrigorian: this is actually an expected behavior: when a zero-dim array is passed as mask (or index), it is brought to the host as a scalar. This especially necessary in the case of boolean masks

see: https://github.com/IntelPython/dpctl/pull/1136

dpctl/tensor/_copy_utils.py
ndgrigorian: yes, I've added it to here and added it in `_place_impl`

`_place_impl` never coerced because of uncertainty surrounding performance. I'll add it now and if we see regressions, I'll change it later.

antonwolfy (approved):
@ndgrigorian , no more comments from me.
I run the dpnp tests with that change and they passed.

PR #2144, opened 2025-09-10
Technical debt clean-up in test suite

ndgrigorian:
This PR proposes reducing the scope of the dpctl test suite by redacting tests which are redundant

For example, the dpctl test suite for element-wise functions repeatedly tests for things like `usm_type` of the output array. But the output array `usm_type` is determined in the function which dispatches to the element-wise kernel, and therefore, only needs to be tested once

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -221,15 +221,6 @@ jobs:
           . $CONDA/etc/profile.d/conda.sh
           conda activate ${{ env.TEST_ENV_NAME }}
           python -c "import dpctl; dpctl.lsplatform(verbosity=2)"
-      - name: Install gdb
-        run: |
-          sudo apt-get update --fix-missing
-          sudo apt-get install -y gdb
-      - name: Run test_elementwise under gdb
-        run: |
-          . $CONDA/etc/profile.d/conda.sh
-          conda activate ${{ env.TEST_ENV_NAME }}
-          gdb --batch -ex r -ex 'info sharedlibrary' -ex 'set print elements 1000' -ex bt --args ${CONDA_PREFIX}/bin/python -m pytest -q -ra --disable-warnings --pyargs dpctl.tests.elementwise.test_trigonometric::test_trig_order -vv || true
       - name: Create test temp dir
         # create temporary empty folder to runs tests from
         # https://github.com/pytest-dev/pytest/issues/11904
--- a/dpctl/tests/elementwise/test_bitwise_invert.py
+++ b/dpctl/tests/elementwise/test_bitwise_invert.py
@@ -20,7 +20,7 @@ import pytest
 import dpctl.tensor as dpt
 from dpctl.tests.helper import get_queue_or_skip, skip_if_dtype_not_supported
 
-from .utils import _compare_dtypes, _integral_dtypes, _usm_types
+from .utils import _compare_dtypes, _integral_dtypes
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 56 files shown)


.github/workflows/conda-package.yml
```diff
@@ -221,15 +221,6 @@ jobs:
           . $CONDA/etc/profile.d/conda.sh
           conda activate ${{ env.TEST_ENV_NAME }}
           python -c "import dpctl; dpctl.lsplatform(verbosity=2)"
-      - name: Install gdb
```
antonwolfy, line 224: It might be helpful in the future, when we will need to debug some test.
Would it be better to comment instead?

dpctl/tests/elementwise/test_bitwise_invert.py
```diff
@@ -73,52 +73,6 @@ def test_bitwise_invert_dtype_matrix(op_dtype):
     assert dpt.all(dpt.equal(r, r3))
 
 
-@pytest.mark.parametrize("op_usm_type", _usm_types)
-def test_bitwise_invert_usm_type_matrix(op_usm_type):
```
antonwolfy, line 77: Having dedicated test with different usm_type` per each elementwise function allows validating the kernel with different USM type of input array allocation.
It is not only about testing common dpctl f/w, but also each kernel.

.github/workflows/conda-package.yml
ndgrigorian: I figured we'd just write it back in later if it becomes necessary again, we could just comment though

dpctl/tests/elementwise/test_bitwise_invert.py
ndgrigorian: it's fair, but I guess there's also a question of how much redundancy is needed in it, since it's really more of a test for SYCL's operators functioning with different usm types (since there's no specialized behavior per-usm-type in the kernels), and comprises a large number of tests

antonwolfy: It might be good to test each elementwise function with simple input per different USM type.
At least this is how we test that in DPNP.

ndgrigorian: I'll drip-feed them back in while trying to maintain test suite stability in CI. My concern is that even simple and small inputs may bloat the device memory of the test suite back to crashing on devices with less memory if implemented in the same number. We'll see though.

ndgrigorian: I've added tests of this kind back in a new test file, `tests/elementwise/test_usm_types.py`

this covers all usm_type combinations for each elementwise function and the basic coercion/expected output type, using some test classes instead of individual tests

dpctl/tests/elementwise/test_usm_types.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+@pytest.mark.parametrize("usm_type", _usm_types)
+class TestUnaryUSMType:
+    def unary_elementwise(self, fn, usm_type, dtype="f4"):
+        q = get_queue_or_skip()
+        x = dpt.asarray(
+            [1, 2, 3, 4], dtype=dtype, usm_type=usm_type, sycl_queue=q
+        )
+        return getattr(dpt, fn)(x)
```
sycloid, line 33: Should an assertion that the result vector has `usm_type` be added here? Costs very little.
I would also make strided `x` inputs, such as `x[::2]` or `dpt.flip(x)`.

ndgrigorian: I opted to add two individual `usm_type` coercion tests, just to minimize redundancy, checking in each test would be fine too

I can also add for strided inputs.

ndgrigorian:
closing this PR as defunct, as tensor submodule was migrated to dpnp

work in this PR will be migrated

PR #2154, opened 2025-09-16
[MAINT] Make error checks in `.bat` files more general

ndgrigorian:
This PR proposes using `if %ERRORLEVEL% neq 0 exit 1` in bld.bat wheel copying, which is more general than `if errorlevel 1 exit 1`

A similar change was suggested when adding the recipe to conda-forge

### DIFF
--- a/conda-recipe/bld.bat
+++ b/conda-recipe/bld.bat
@@ -1,4 +1,3 @@
-
 REM A workaround for activate-dpcpp.bat issue to be addressed in 2021.4
 set "LIB=%BUILD_PREFIX%\Library\lib;%BUILD_PREFIX%\compiler\lib;%LIB%"
 set "INCLUDE=%BUILD_PREFIX%\include;%INCLUDE%"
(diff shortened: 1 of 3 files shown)


conda-recipe/bld.bat
```diff
@@ -65,5 +64,5 @@ for /f %%f in ('dir /b /S .\dist') do (
 :: Copy wheel package
 if NOT "%WHEELS_OUTPUT_FOLDER%"=="" (
     copy dist\dpctl*.whl %WHEELS_OUTPUT_FOLDER%
-    if errorlevel 1 exit 1
+    if %ERRORLEVEL% neq 0 exit 1
```
antonwolfy, line 67: What is about cmake file copying above? Should we use `%ERRORLEVEL% neq 0` there also?

antonwolfy: And probably we should update `conda-recipe/run_test.bat` and `libsyclinterface/dbg_build.bat` to use the same generic check everywhere.

ndgrigorian: done

antonwolfy (approved):
Thank you @ndgrigorian, LGTM

PR #2172, opened 2025-10-15
Remove use of `python setup.py develop/install` in the project

ndgrigorian:
A warning that `python setup.py develop` is deprecated and will no longer be supported at the end of October, 2025 has been ongoing for some time when using dpctl build driver scripts, and can be seen in the CI

This PR proposes instead relying on `pip` for installing dpctl in scripts, and reworks the scripts to maintain use of scikit-build

In the future, this will also simplify a transition to scikit-build-core

The PR also introduces options
* `--clean` and `--skip-editable` to `build_locally` driver
* `--clean` and `--skip-pytest` to `gen_coverage` driver
* `--clean` to `gen_docs` driver

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -75,8 +75,7 @@ jobs:
           source /opt/intel/oneapi/setvars.sh
           wget https://github.com/vovkos/doxyrest/releases/download/doxyrest-2.1.2/doxyrest-2.1.2-linux-amd64.tar.xz
           tar xf doxyrest-2.1.2-linux-amd64.tar.xz
-          python scripts/gen_docs.py --c-compiler=$(which icx) --cxx-compiler=$(which icpx) \
-              --doxyrest-root=`pwd`/doxyrest-2.1.2-linux-amd64 --verbose || exit 1
+          python scripts/gen_docs.py --doxyrest-root=`pwd`/doxyrest-2.1.2-linux-amd64 --verbose || exit 1
           python -c "import dpctl; print(dpctl.__version__)" || exit 1
           pushd "$(find _skbuild -name cmake-build)" || exit 1
           cmake --build . --target Sphinx || exit 1
--- a/scripts/build_locally.py
+++ b/scripts/build_locally.py
@@ -94,8 +94,7 @@ def parse_args():
         dest="no_level_zero",
         action="store_true",
         default=False,
-        help="Disable Level Zero backend (deprecated: use --target-level-zero "
-        "OFF)",
+        help="Disable Level Zero backend",
     )
 
     p.add_argument(
@@ -168,11 +167,6 @@ def main():
         cmake_args += [f"-DDPCTL_TARGET_CUDA={args.target_cuda}"]
     if args.target_hip:
         cmake_args += [f"-DDPCTL_TARGET_HIP={args.target_hip}"]
-
-    cmake_args += [
-        "-DDPCTL_ENABLE_L0_PROGRAM_CREATION="
-        f"{'ON' if level_zero_enabled else 'OFF'}"
-    ]
 
     log_cmake_args(cmake_args, "build_locally")
 
--- a/scripts/gen_coverage.py
+++ b/scripts/gen_coverage.py
@@ -87,8 +87,7 @@ def parse_args():
         dest="no_level_zero",
         action="store_true",
         default=False,
-        help="Disable Level Zero backend (deprecated: use --target-level-zero "
-        "OFF)",
+        help="Disable Level Zero backend",
     )
 
     p.add_argument(
--- a/scripts/gen_docs.py
+++ b/scripts/gen_docs.py
@@ -66,8 +66,7 @@ def parse_args():
         dest="no_level_zero",
         action="store_true",
         default=False,
-        help="Disable Level Zero backend (deprecated: use --target-level-zero "
-        "OFF)",
+        help="Disable Level Zero backend",
     )
 
     p.add_argument(


scripts/_build_helper.py
```diff
@@ -0,0 +1,127 @@
+#                      Data Parallel Control (dpctl)
+#
+# Copyright 2020-2025 Intel Corporation
```
antonwolfy:
```suggestion
# Copyright 2025 Intel Corporation
```

scripts/build_locally.py
```diff
@@ -14,225 +14,204 @@
...
+    setup_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    build_dir = os.path.join(setup_dir, args.build_dir)
+
+    c_compiler, cxx_compiler, compiler_root = resolve_compilers(
+        args.oneapi, args.c_compiler, args.cxx_compiler, args.compiler_root
+    )
+
+    # clean build dir if --clean set
+    if args.clean:
+        clean_build_dir(build_dir)
```
antonwolfy: I wonder if that works with default build directory. I used:
```bash
$ python scripts/build_locally.py
...
$ python scripts/build_locally.py --clean
[build_locally] CMake args:
 -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER:PATH=icx -DCMAKE_CXX_COMPILER:PATH=icpx -DDPCTL_ENABLE_L0_PROGRAM_CREATION=ON -DDPTL_ENABLE_GLOG:BOOL=OFF -GNinja -DDPCTL_ENABLE_L0_PROGRAM_CREATION=ON
[build_locally] Building extensions in-place...
+ /localdisk/work/antonvol/miniforge3/envs/dpctl_dev/bin/python setup.py build_ext --inplace
...
```
Seems no clean up of the build dir.

scripts/build_locally.py
```diff
@@ -14,225 +14,204 @@
...
-        help="Is one-API installation",
         dest="oneapi",
         action="store_true",
+        help="Use default oneAPI compiler layout",
     )
-    driver.add_argument(
+    p.add_argument(
         "--debug",
-        default="Release",
+        dest="build_type",
```
antonwolfy, line 62: The debug build produces a lot of the warning which probably has to be resolved in the follow-up PR.

scripts/build_locally.py
ndgrigorian: that's likely because I left in `--build-dir` which would only work if we moved to scikit-build-core, which I think should be done in a later PR

I'll remove the option so we just clean up `_skbuild`

antonwolfy: Do you mean removing `--build-dir` or `--clean`, or both options?
Currently, `--clean` option doesn't clean up `_skbuild` folder.

ndgrigorian: removing `--build-dir`

right now `--build-dir` does nothing, but is passed to the clean helper, and so, clean doesn't happen

scripts/build_locally.py
ndgrigorian: a number of them are
`icpx: warning: argument unused during compilation: '-Xsycl-target-frontend=spir64 -g0'`
which will need some looking at

the remainder are
`warning: comparison of integers of different signs: 'py::ssize_t' (aka 'long') and 'std::size_t'`
from
`assert(dst.get_size() == iter_nelems)`
in sorting functions

should be a simple enough fix

.github/workflows/conda-package.yml
```diff
@@ -633,7 +633,7 @@ jobs:
           do
             pushd $d
             conda activate --stack ${{ env.BUILD_ENV_NAME }}
-            CC=icx CXX=icpx python setup.py develop -G Ninja  || exit 1
+            CC=icx CXX=icpx python setup.py build_ext --inplace -G Ninja  || exit 1
```
antonwolfy, line 636: `CONTRIBUTING.md` needs to be updated also

.github/workflows/generate-docs.yml
```diff
@@ -75,14 +75,15 @@ jobs:
           source /opt/intel/oneapi/setvars.sh
           wget https://github.com/vovkos/doxyrest/releases/download/doxyrest-2.1.2/doxyrest-2.1.2-linux-amd64.tar.xz
           tar xf doxyrest-2.1.2-linux-amd64.tar.xz
-          python setup.py develop -G Ninja --build-type=Release \
+          python setup.py build_ext --inplace --generator=Ninja --build-type=Release \
```
antonwolfy: I wonder why we are not using `scripts/gen_docs.py` in this workflow.

.github/workflows/os-llvm-sycl-build.yml
```diff
@@ -146,7 +146,8 @@ jobs:
         shell: bash -l {0}
         run: |
           source set_allvars.sh
-          CC=clang CXX=clang++ python setup.py develop -G Ninja
+          CC=clang CXX=clang++ python setup.py build_ext --inplace --generator=Ninja
```
antonwolfy: Would it make sense to use `python scripts/build_locally.py` here instead?

.github/workflows/conda-package.yml
antonwolfy: and `docs/_legacy/docfiles/user_guides/QuickStart.rst` and `docs/doc_sources/contributor_guides/building.rst`

antonwolfy: some README.md in the examples folder as well

scripts/build_locally.py
```diff
@@ -14,225 +14,205 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import argparse
 import os
-import subprocess
 import sys
 
+# add scripts dir to Python path so we can import _build_helper
+sys.path.insert(0, os.path.abspath("scripts"))
```
antonwolfy: I wonder for what use case we need that? Simple `python scripts/build_locally.py` works fine without that.

scripts/gen_coverage.py
```diff
@@ -14,80 +14,218 @@
...
+
+    p.add_argument(
+        "--gtest-config",
+        help="Path to GTestConfig.cmake file for a custom GTest installation",
+    )
+    p.add_argument(
+        "--bin-llvm",
+        help="Path to folder where llvm-cov/llvm-profdata can be found",
+    )
+    p.add_argument("--skip-pytest", dest="run_pytest", action="store_false")
```
antonwolfy: Based on PR's description it seems it has to be `--no-pytest`, or the description of the PR needs to be updated.

antonwolfy: Missing description of the argument

scripts/gen_docs.py
```diff
@@ -14,175 +14,188 @@
...
+
+    env = os.environ.copy()
+
+    if "CMAKE_ARGS" in env and env["CMAKE_ARGS"].strip():
+        warn("Ignoring pre-existing CMAKE_ARGS in environment", "gen_docs")
+        del env["CMAKE_ARGS"]
+
+    env["CMAKE_ARGS"] = cmake_args
+
+    print(f"[gen_docs] Using CMake args:\n {env['CMAKE_ARGS']}")
```
antonwolfy: Would it make sense to move that to a helper function, looks quite the same use across the files?

scripts/build_locally.py
```diff
@@ -14,225 +14,205 @@
...
+        clean_build_dir(setup_dir)
+
+    if args.no_level_zero and args.target_level_zero:
+        err(
+            "Cannot combine --no-level-zero and --target-level-zero",
+            "build_locally",
+        )
+
+    # Level Zero state (on unless explicitly disabled)
+    if args.no_level_zero:
```
antonwolfy: I wonder, if `--no-level-zero` is deprecated, should we base the logic mostly on `--target-level-zero` option then?
Should we consider `level_zero_enabled = False` if `--target-level-zero` is not passed?

scripts/build_locally.py
```diff
@@ -14,225 +14,205 @@
...
         action="store_true",
+        help="Enable verbose makefile output",
+    )
+
+    p.add_argument(
+        "--no-level-zero",
+        dest="no_level_zero",
+        action="store_true",
+        default=False,
+        help="Disable Level Zero backend (deprecated: use --target-level-zero "
```
antonwolfy: `--target-level-zero OFF` does not work

scripts/build_locally.py
```diff
@@ -14,225 +14,205 @@
...
+        err(
+            "Cannot combine --no-level-zero and --target-level-zero",
+            "build_locally",
+        )
+
+    # Level Zero state (on unless explicitly disabled)
+    if args.no_level_zero:
+        level_zero_enabled = False
+    elif args.target_level_zero:
+        level_zero_enabled = True
```
antonwolfy: This is the same in all three scripts and can be moved to the helper as a function.

scripts/gen_docs.py
ndgrigorian: I plan to remove this, I was mistakenly under the impression that we needed to pass via `CMAKE_ARGS` when using `python setup.py build_ext --inplace`, but I can get it working without it

scripts/gen_coverage.py
ndgrigorian: I'll fix the description

scripts/build_locally.py
ndgrigorian: It didn't work for me locally without this, essentially importing from `_build_helper` wouldn't work without this or an `__init__.py`

.github/workflows/os-llvm-sycl-build.yml
ndgrigorian: It's probably a better approach, maybe this and generate-docs can be updated in a follow-up PR

.github/workflows/generate-docs.yml
ndgrigorian: See below comment, I think it's a good idea for a follow-up PR

scripts/build_locally.py
antonwolfy: Probably it would be better to add `__init__.py` then. It looks clearer.

scripts/build_locally.py
ndgrigorian: I think we should default `--target-level-zero` to `TRUE`. The reason for the change is to later introduce targeting Level Zero architectures (like with CUDA and HIP). We have the ability to target architectures but it's unused at the moment.

`TRUE` would default to current behavior, `<ARCH>` would only build targeting a specific architecture, and `OFF` would imitate `--no-level-zero`.

ndgrigorian: You're right about basing the logic on `--target-level-zero` though. We can keep the option and have `--no-level-zero` override `--target-level-zero=TRUE` while deprecated

scripts/gen_docs.py
ndgrigorian: removed it

scripts/_build_helper.py
```diff
@@ -0,0 +1,157 @@
...
+    level_zero: bool = True,
+    glog: bool = False,
+    verbose: bool = False,
+    other_opts: str = None,
+):
+    args = [
+        f"-DCMAKE_C_COMPILER:PATH={c_compiler}" if c_compiler else "",
+        f"-DCMAKE_CXX_COMPILER:PATH={cxx_compiler}" if cxx_compiler else "",
+        f"-DDPCTL_ENABLE_L0_PROGRAM_CREATION={'ON' if level_zero else 'OFF'}",
+        f"-DDPTL_ENABLE_GLOG:BOOL={'ON' if glog else 'OFF'}",
```
vlad-perevezentsev:
```suggestion
        f"-DDPCTL_ENABLE_GLOG:BOOL={'ON' if glog else 'OFF'}",
```

scripts/_build_helper.py
```diff
@@ -0,0 +1,157 @@
...
+    ]
+
+    if verbose:
+        args.append("-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON")
+    if other_opts:
+        args.extend(other_opts.split())
+
+    return args
```
vlad-perevezentsev, line 109: Is it safe to pass a list containing empty ("") arguments to CMake?

scripts/build_locally.py
```diff
@@ -14,225 +14,206 @@
...
-        target_cuda=args.target_cuda,
-        target_hip=args.target_hip,
+        other_opts=args.cmake_opts,
     )
+
+    # handle architecture conflicts
+    if args.target_hip is not None and not args.target_hip.strip():
+        err("--target-hip requires an explicit architecture", "build_locally")
+
+    # CUDA/HIP targets
```
vlad-perevezentsev, line 165: The previous implementation included a safeguard to prevent using `--target-cuda` together with -`DDPCTL_TARGET_CUDA=...`  (HIP as well) in `--cmake-opts`
```
if target_cuda is not None:
        if not target_cuda.strip():
            raise ValueError(
                "--target-cuda can not be an empty string. "
                "Use --target-cuda=<arch> or --target-cuda"
            )
        if any(opt.startswith("-DDPCTL_TARGET_CUDA=") for opt in cmake_args):
            raise ValueError(
                "Both --target-cuda and -DDPCTL_TARGET_CUDA in --cmake-opts "
                "were specified. Please use only one method "
                "to avoid ambiguity"
            )
        cmake_args += [
            f"-DDPCTL_TARGET_CUDA={target_cuda}",
        ]

```
Why did you decide to remove this check?

scripts/_build_helper.py
```diff
@@ -0,0 +1,157 @@
...
+    print("+", " ".join(cmd))
+    return (
+        subprocess.check_output(cmd, cwd=cwd or os.getcwd())
+        .decode("utf-8")
+        .strip("\n")
+    )
+
+
+def err(msg: str, script: str):
+    print(f"[{script}][error] {msg}", file=sys.stderr)
```
vlad-perevezentsev: This function only prints an error message.
Should we also terminate build_locally script when `err()` is called?

scripts/build_locally.py
```diff
@@ -14,225 +14,206 @@
...
+        help="Enable HIP backend. Architecture required to be specified.",
     )
-    args = parser.parse_args()
 
-    args_to_validate = [
-        "c_compiler",
-        "cxx_compiler",
-        "compiler_root",
-    ]
+    p.add_argument(
```
vlad-perevezentsev, line 121: It does not work
```
$ python scripts/build_locally.py --clean
$ ls | grep _skbuild
> _skbuild

$ ls _skbuild/linux-x86_64-3.13/
> cmake-build  cmake-install  setuptools
```

scripts/_build_helper.py
ndgrigorian: good catch

scripts/build_locally.py
ndgrigorian: It was a deliberate choice, I decided that if users decide to make this mistake, we shouldn't account for it

If we do it for these arguments, the question becomes why we don't use it for all arguments, for one. Next, these CMake args are for power users anyway, who should know better than to pass both. And finally, I decided that the scripts should veer away from encouraging using CMake args in general, and should focus on simplifying the process through the CLI args.

So in general, it was part of a general decision to make the scripts the simplest way to build dpctl

scripts/build_locally.py
ndgrigorian: after a bit more thinking, I will drop the `--target-level-zero` argument for now until introducing it with architecture later

it seems that most recommendations are to use a `--no-arg` if we plan to default one way or another. We can simply make the two mutually exclusive arguments, and if `--target-level-zero` is specified with `--no-target-level-zero` we can fail it

scripts/build_locally.py
ndgrigorian: it seemed to work for me locally, though it didn't actually print the message that it was cleaning, strangely. If I don't activate oneAPI in the environment but pass `--clean`, the folder is gone. It's just remade quickly afterward with the compiler active.

Can you try this approach?

ndgrigorian: tested again and it did seem to print and clean the directory

scripts/_build_helper.py
ndgrigorian: yeah, seems I left out the `RuntimeError` I originally put in here

scripts/_build_helper.py
ndgrigorian: if `--cmake-opts` is `""`, then `args.extend(other_opts.split())` will result in the same `args` as previously, as `extend` with an empty list does not change the extended list, so it should be safe. I don't think it's possible for `split` to produce a list such as `["", "", ""]`

.github/workflows/conda-package.yml
ndgrigorian: updated all except for `docs/_legacy`, since we no longer use the legacy docs and they are kept only as a formality

scripts/build_locally.py
ndgrigorian: removed and it worked, probably also because I switched to absolute instead of relative import

.github/workflows/generate-docs.yml
```diff
@@ -75,14 +75,8 @@ jobs:
...
           tar xf doxyrest-2.1.2-linux-amd64.tar.xz
-          python setup.py develop -G Ninja --build-type=Release \
-              --                                              \
-              -DCMAKE_C_COMPILER:PATH=$(which icx)            \
-              -DCMAKE_CXX_COMPILER:PATH=$(which icpx)         \
-              -DDPCTL_GENERATE_DOCS=ON                        \
-              -DDPCTL_ENABLE_DOXYREST=ON                      \
-              -DDoxyrest_DIR=`pwd`/doxyrest-2.1.2-linux-amd64 \
-              -DCMAKE_VERBOSE_MAKEFILE=ON
+          python scripts/gen_docs.py --c-compiler=$(which icx) --cxx-compiler=$(which icpx) \
```
antonwolfy: Why do we need to pass `--c-compiler` and `--cxx-compiler` options here?
Will the script succeed with the unset values, because there was `source /opt/intel/oneapi/setvars.sh` above?

scripts/build_locally.py
```diff
@@ -14,225 +14,187 @@
...
         action="store_true",
+        help="Enable verbose makefile output",
+    )
+
+    p.add_argument(
+        "--no-level-zero",
+        dest="no_level_zero",
+        action="store_true",
+        default=False,
+        help="Disable Level Zero backend (deprecated: use --target-level-zero "
```
antonwolfy: Do we need to remove the deprecation note than, since `--target-level-zero` is not available as for now?

scripts/build_locally.py
```diff
@@ -14,225 +14,187 @@
...
+    if args.target_hip is not None and not args.target_hip.strip():
+        err("--target-hip requires an explicit architecture", "build_locally")
+
+    # CUDA/HIP targets
+    if args.target_cuda:
+        cmake_args += [f"-DDPCTL_TARGET_CUDA={args.target_cuda}"]
+    if args.target_hip:
+        cmake_args += [f"-DDPCTL_TARGET_HIP={args.target_hip}"]
+
+    cmake_args += [
```
antonwolfy: Isn't that a part of `make_cmake_args` and has been already done at that point?

PR #2173, opened 2025-10-15
Add `3.14` python

ekomarova:
This PR adds 3.14 to `pyproject.toml`

### DIFF
--- a/.github/workflows/run-tests-from-dppy-bits.yaml
+++ b/.github/workflows/run-tests-from-dppy-bits.yaml
@@ -27,7 +27,7 @@ jobs:
 
     strategy:
       matrix:
-        python: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
         experimental: [false]
         runner: [ubuntu-22.04, ubuntu-24.04]
     continue-on-error: ${{ matrix.experimental }}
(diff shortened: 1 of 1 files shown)


ndgrigorian:
@ekomarova 
we can add it to the Python version matrices for the public CI as well

.github/workflows/conda-package.yml
```diff
@@ -423,7 +423,7 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        python: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
```
antonwolfy, line 426: I guess we need to update `.github/workflows/run-tests-from-dppy-bits.yaml` also

ekomarova: I'm not sure, intel channel is used there, I don't know if it will work as expected

antonwolfy: I believe it should work, because the workflow is intended to install the latest packages from `dppy/label/dev` and to run tests. It is similar with what is doing in `.github/workflows/conda-package.yml`.
Btw there is a way to check. If you update `.github/workflows/run-tests-from-dppy-bits.yaml` I will run the workflow manually and post the result here.

ekomarova: Added

antonwolfy: > I will run the workflow manually and post the result here.

I was wrong, this PR has to merged first.

ndgrigorian (approved):

PR #2180, opened 2025-10-22
Drop support for Python 3.9

ndgrigorian:
This PR proposes dropping Python 3.9 support from the project

NumPy and most of the ecosystem has already done the same for some time

includes a bump to `find_package(Python 3.9...)` to 3.10 in `dpctl-config.cmake`

### DIFF
--- a/cmake/dpctl-config.cmake
+++ b/cmake/dpctl-config.cmake
@@ -20,7 +20,7 @@
 #
 
 if(NOT Dpctl_FOUND)
-  find_package(Python 3.9 REQUIRED
+  find_package(Python 3.10 REQUIRED
     COMPONENTS Interpreter Development.Module)
 
   if(Python_EXECUTABLE)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,6 @@ classifiers = [
   "Programming Language :: C",
   "Programming Language :: Python",
   "Programming Language :: Python :: 3",
-  "Programming Language :: Python :: 3.9",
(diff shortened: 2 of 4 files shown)


cmake/dpctl-config.cmake
```diff
@@ -20,7 +20,7 @@
 #
 
 if(NOT Dpctl_FOUND)
-  find_package(Python 3.9 REQUIRED
+  find_package(Python 3.10 REQUIRED
```
antonwolfy, line 23: Should we be more explicit here?
```suggestion
  find_package(Python 3.10...<3.15 REQUIRED
```

pyproject.toml
```diff
@@ -58,7 +57,7 @@ keywords = [
 license = "Apache-2.0"
 name = "dpctl"
 readme = {file = "README.md", content-type = "text/markdown"}
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
antonwolfy, line 60: Do we need to update the example in `examples/cython/sycl_buffer/README.md`?
It has been run with python 3.9.

cmake/dpctl-config.cmake
ndgrigorian: I'm fine with leaving it as only a minimum, we already set max version in `pyproject.toml`

pyproject.toml
ndgrigorian: We should, but examples in general need updating, which we can save for future PRs

PR #2191, opened 2025-11-19
Add deprecation warning for the the dpctl.tensor submodule

ndgrigorian:
As plans have now evolved to move the tensor submodule to dpnp, the tensor submodule will be deprecated in a coming minor version release

### DIFF
--- a/dpctl/tensor/__init__.py
+++ b/dpctl/tensor/__init__.py
@@ -214,7 +214,9 @@ from ._type_utils import can_cast, finfo, iinfo, isdtype, result_type
 # deprecation warning for the dpctl.tensor module
 _warnings.warn(
     "dpctl.tensor is deprecated since dpctl 0.21.1 and will be removed in a "
-    "future release. Install dpnp and use 'import dpnp.tensor' instead.",
+    "future release. The functionality will be moved to separate package, dpnp "
+    "(see: https://github.com/IntelPython/dpnp). After that, use "
+    "'import dpnp.tensor' instead.",
     DeprecationWarning,
     stacklevel=2,
 )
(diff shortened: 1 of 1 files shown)


dpctl/tensor/__init__.py
```diff
@@ -208,6 +211,14 @@
 from ._testing import allclose
 from ._type_utils import can_cast, finfo, iinfo, isdtype, result_type
 
+# deprecation warning for the dpctl.tensor module
+_warnings.warn(
+    "dpctl.tensor is deprecated since dpctl 0.21.1 and will be removed in a "
+    "future release. Install dpnp and use 'import dpnp.tensor' instead.",
```
antonwolfy: Can we propose that, if there is no `dpnp.tensor` and it will not be a part of upcoming minor dpnp release?

dpctl/tensor/__init__.py
```diff
@@ -397,3 +408,17 @@
...
+
+
+def __getattr__(name: str):  # pragma: no cover
+    # per-attribute access deprecation notices per PEP 562
+    if name in __all__:
+        _warnings.warn(
+            f"dpctl.tensor.{name} is deprecated; dpctl.tensor is deprecated "
+            "since dpctl 0.21.1 and will be removed in a future release. "
+            "Install dpnp and use 'import dpnp.tensor' instead.",
+            DeprecationWarning,
```
antonwolfy, line 424: Do we need to add tests per both deprecation warnings?

dpctl/tensor/__init__.py
ndgrigorian: good point, the text is misleading, I guess it should be rephrased to instead indicate it as more of a future change, similar to NumPy deprecation warnings

ndgrigorian: I've updated the text, it's a bit clunky but I think it's good enough

dpctl/tensor/__init__.py
ndgrigorian: I don't think there's much of a need for tests, coverage remains the same anyway and it seems to work for me locally

PR #2192, opened 2025-11-19
Bump minimum NumPy version in `pyproject.toml`

ndgrigorian:
This PR proposes bumping the minimum NumPy version to align with dpnp


ndgrigorian:
@antonwolfy @vlad-perevezentsev @xaleryb 
I've updated this PR to bump only to 1.26.0, to align with dpnp

pyproject.toml
```diff
@@ -9,7 +9,7 @@ requires = [
   "ninja>=1.11.1; platform_system!='Windows'",
   "cmake>=3.29.0",
   "cython>=3.0.10",
-  "numpy >=1.23",
+  "numpy >=1.26.0",
```
vlad-perevezentsev, line 12: I think numpy should also be bumped to version `1.26` in `conda_build_config.yaml` [file](https://github.com/IntelPython/dpctl/blob/master/conda-recipe/conda_build_config.yaml#L2) so that the conda package is built with the same minimum numpy version

ndgrigorian: I made the change, though I will note that conda package builds end up not testing with 1.26 anyway since we pass with --numpy anyway

It would maybe make sense to test with the oldest version too

vlad-perevezentsev (dismissed):
LGTM
Thank you @ndgrigorian

vlad-perevezentsev (approved):

PR #2193, opened 2025-11-19
Update dlpack to v1.2

ndgrigorian:
This PR proposes updating the vendored DLPack header `dlpack.h` to v1.2 and adds data type code and device type enum definitions where necessary

It does not work towards supporting the new C functions to optimize speed of exchange, which could be done in a follow-up PR

### DIFF
--- a/dpctl/tensor/_dlpack.pyx
+++ b/dpctl/tensor/_dlpack.pyx
@@ -36,7 +36,7 @@ from .._backend cimport (
     DPCTLSyclDeviceRef,
     DPCTLSyclUSMRef,
 )
-from ._usmarray cimport USM_ARRAY_C_CONTIGUOUS, USM_ARRAY_WRITABLE, usm_ndarray
+from ._usmarray cimport USM_ARRAY_WRITABLE, usm_ndarray
 
 import ctypes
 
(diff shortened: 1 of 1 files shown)


dpctl/tensor/_dlpack.pyx
```diff
@@ -76,6 +76,7 @@ cdef extern from "dlpack/dlpack.h" nogil:
         kDLWebGPU
         kDLHexagon
         kDLMAIA
+        kDLTrn
```
antonwolfy, line 79: It seems we need to update `numpy_to_dlpack_versioned_capsule` at [line](https://github.com/IntelPython/dpctl/blob/68c2167d44329f4d3a3205490b31c342dd22d59c/dpctl/tensor/_dlpack.pyx#L557) following dlpack changes added in v1.2
NumPy did that [here](https://github.com/numpy/numpy/pull/29872).

ndgrigorian: good point, I didn't think we would need to make a change for that part, but I'll adjust it

ndgrigorian: all logic should now be updated

shape and strides are both set to `NULL` for `ndim == 0` only, and conversely, when converting from dlpack, shape and strides are simplified a bit if `NULL` is provided

antonwolfy: Do we need to test that?

ndgrigorian: I don't think it's necessary, we already have tests which hit these paths in `_dlpack.pyx`. The new lines that are uncovered are all not possible to test (cases where OOM means shape and stride ptrs can't be allocated)

PR #2211, opened 2025-12-11
Fix CMake error in OS compiler workflow

ndgrigorian:
This PR fixes the failing OS compiler workflow, which has been throwing a CMake error

OS compiler versioning output changed, so the Clang version is preceded by another string giving the date and specifying the compiler is the SYCL nightly compiler

This PR modifies dpctl CMake to check if the "Intel SYCL Compiler Nightly" string is in the version string. If so, the next member of the list, containing the Clang version, is used instead

### DIFF
--- a/libsyclinterface/cmake/modules/FindIntelSyclCompiler.cmake
+++ b/libsyclinterface/cmake/modules/FindIntelSyclCompiler.cmake
@@ -102,13 +102,13 @@ if(${clangxx_result} MATCHES "0")
             # Match 'clang version xx.x.x'
             string(REGEX MATCH "^.*clang version [0-9]+\\.[0-9]+\\.[0-9]+.*$" _clang_match "${X}")
             if(_clang_match)
-                string(REGEX REPLACE "^.*clang version ([0-9]+\\.[0-9]+\\.[0-9]+).*$" "\\1" IntelSyclCompiler_VERSION "${X}")
+                string(REGEX REPLACE "^.*clang version ([0-9]+\\.[0-9]+\\.[0-9]+).*$" "\\1" IntelSyclCompiler_VERSION "${_clang_match}")
             endif()
 
             # Match 'Intel(R) oneAPI DPC++/C++ Compiler xxxx.x.x (...)'
             string(REGEX MATCH "^.*Intel\\(R\\) oneAPI DPC\\+\\+\\/C\\+\\+ Compiler [0-9]+\\.[0-9]+\\.[0-9]+.*$" _oneapi_match "${X}")
             if(_oneapi_match)
-                string(REGEX REPLACE "^.*Intel\\(R\\) oneAPI DPC\\+\\+\\/C\\+\\+ Compiler ([0-9]+\\.[0-9]+\\.[0-9]+).*$" "\\1" IntelSyclCompiler_VERSION "${X}")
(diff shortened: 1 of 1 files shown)


libsyclinterface/cmake/modules/FindIntelSyclCompiler.cmake
```diff
@@ -100,6 +100,10 @@ if(${clangxx_result} MATCHES "0")
         MATH(EXPR IDX "${IDX}+1")
     endforeach()
     list(GET IntelSyclCompiler_VERSION_LIST 0 VERSION_STRING)
+    if("${VERSION_STRING}" MATCHES "Intel SYCL compiler Nightly")
```
antonwolfy: Would it make sense to update the `libsyclinterface/cmake/modules/FindIntelSyclCompiler.cmake` with the latest content from the recent 2025.3 release instead?

ndgrigorian: I see what you're saying here now. This file isn't a vendored CMake module, this is a custom CMake module for finding the compiler that consolidated old logic for finding the compiler.

That is, 2025.3 would contain no equivalent file to update with, if that's what you're saying

ndgrigorian: I will say that the solution here is a bit hacky, a more elegant solution may be to check each index instead, and get the index that's either the DPC++ version or the clang version.

This would future-proof any other, future changes to how the versioning is given

antonwolfy: > This file isn't a vendored CMake module, this is a custom CMake module for finding the compiler that consolidated old logic for finding the compiler.

I see, it seems my previous understanding wasn't correct.
I thought it is exactly the same as supplied by DPC++ compiler conda package but not available by expected due to some reason or other deployment.

antonwolfy (approved):
No more comments from me. LGTM!

ndgrigorian:
@antonwolfy I've made a change to make this check more robust with regex, so each index is checked for either until a match is found, and that's used as the version. I added a commit to show the version, and as you can see, it matches what it was before.
DPC++:
Screenshot 2025-12-17 091039
OS compiler:
Screenshot 2025-12-17 091105

libsyclinterface/cmake/modules/FindIntelSyclCompiler.cmake
```diff
@@ -95,20 +95,26 @@ execute_process(
...
-    # Get the dpcpp version
-    string(REGEX MATCH
-        "[0-9]+\.[0-9]+\.[0-9]+"
-        IntelSyclCompiler_VERSION
-        ${VERSION_STRING}
-    )
+            # Match 'Intel(R) oneAPI DPC++/C++ Compiler xxxx.x.x (...)'
+            string(REGEX MATCH "^.*Intel\\(R\\) oneAPI DPC\\+\\+\\/C\\+\\+ Compiler [0-9]+\\.[0-9]+\\.[0-9]+.*$" _oneapi_match "${X}")
+            if(_oneapi_match)
+                string(REGEX REPLACE "^.*Intel\\(R\\) oneAPI DPC\\+\\+\\/C\\+\\+ Compiler ([0-9]+\\.[0-9]+\\.[0-9]+).*$" "\\1" IntelSyclCompiler_VERSION "${X}")
```
antonwolfy: Shouldn't we use the found match?
```suggestion
                string(REGEX REPLACE "^.*Intel\\(R\\) oneAPI DPC\\+\\+\\/C\\+\\+ Compiler ([0-9]+\\.[0-9]+\\.[0-9]+).*$" "\\1" IntelSyclCompiler_VERSION "${_oneapi_match}")
```

antonwolfy (approved):

PR #2225, opened 2026-01-06
Allow Python scalars as the search values in `dpt.searchsorted`

ndgrigorian:
This PR proposes permitting the search values of `dpt.searchsorted` to be Python scalars, a change coming to the 2026 array API spec

Resolves IntelPython/dpnp#2883 

### DIFF
--- a/dpctl/tensor/_searchsorted.py
+++ b/dpctl/tensor/_searchsorted.py
@@ -139,9 +139,10 @@ def searchsorted(
     dt1, dt2 = _resolve_weak_types_all_py_ints(x1_dt, x2_dt, sycl_dev)
     dt = _to_device_supported_dtype(dpt.result_type(dt1, dt2), sycl_dev)
 
+    # get submitted events again in case some were added by sorter handling
+    dep_evs = _manager.submitted_events
     if x1_dt != dt:
         x1_buf = _empty_like_orderK(x1, dt)
-        dep_evs = _manager.submitted_events
         ht_ev, ev = ti_copy(src=x1, dst=x1_buf, sycl_queue=q, depends=dep_evs)
         _manager.add_event_pair(ht_ev, ev)
         x1 = x1_buf
@@ -150,7 +151,6 @@ def searchsorted(
         x2 = dpt.asarray(x2, dtype=dt2, usm_type=res_usm_type, sycl_queue=q)
     if x2.dtype != dt:
         x2_buf = _empty_like_orderK(x2, dt)
-        dep_evs = _manager.submitted_events
         ht_ev, ev = ti_copy(src=x2, dst=x2_buf, sycl_queue=q, depends=dep_evs)
         _manager.add_event_pair(ht_ev, ev)
         x2 = x2_buf
--- a/dpctl/tensor/_searchsorted.py
+++ b/dpctl/tensor/_searchsorted.py
@@ -1,23 +1,29 @@
 from typing import Literal, Union
 
 import dpctl
+import dpctl.tensor as dpt
 import dpctl.utils as du
 
 from ._copy_utils import _empty_like_orderK
 from ._ctors import empty
+from ._scalar_utils import _get_dtype, _get_queue_usm_type, _validate_dtype
 from ._tensor_impl import _copy_usm_ndarray_into_usm_ndarray as ti_copy
 from ._tensor_impl import _take as ti_take
 from ._tensor_impl import (
     default_device_index_type as ti_default_device_index_type,
 )
 from ._tensor_sorting_impl import _searchsorted_left, _searchsorted_right
-from ._type_utils import isdtype, result_type
+from ._type_utils import (
+    _resolve_weak_types_all_py_ints,
+    _to_device_supported_dtype,
+    isdtype,
+)
 from ._usmarray import usm_ndarray
 
 
 def searchsorted(
     x1: usm_ndarray,
-    x2: usm_ndarray,
+    x2: Union[usm_ndarray, int, float, complex, bool],
     /,
     *,
     side: Literal["left", "right"] = "left",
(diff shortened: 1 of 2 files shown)


dpctl/tensor/_searchsorted.py
```diff
@@ -116,29 +136,28 @@ def searchsorted(
...
+    if x1_dt != dt:
+        x1_buf = _empty_like_orderK(x1, dt)
+        dep_evs = _manager.submitted_events
+        ht_ev, ev = ti_copy(src=x1, dst=x1_buf, sycl_queue=q, depends=dep_evs)
+        _manager.add_event_pair(ht_ev, ev)
+        x1 = x1_buf
+
+    if not isinstance(x2, usm_ndarray):
+        x2 = dpt.asarray(x2, dtype=dt2, usm_type=res_usm_type, sycl_queue=q)
+    if x2.dtype != dt:
```
antonwolfy, line 152:
```suggestion
    elif x2.dtype != dt:
```

dpctl/tensor/_searchsorted.py
```diff
@@ -116,29 +136,28 @@ def searchsorted(
...
+        x2_buf = _empty_like_orderK(x2, dt)
+        dep_evs = _manager.submitted_events
+        ht_ev, ev = ti_copy(src=x2, dst=x2_buf, sycl_queue=q, depends=dep_evs)
```
antonwolfy, line 154: Copying `x1` and `x2` to the buffers might be done in parallel.
While currently it assumes a sequence execution order: `x1` must be copied first and then either `x2` is cased to usm_ndarray or `x2` is copying to the buffer (but always once the `x1` copy kernel is completed).

dpctl/tests/test_usm_ndarray_searchsorted.py
```diff
@@ -375,3 +393,23 @@ def test_searchsorted_strided_scalar_needle():
...
     _check(hay_stack, needles, needles_np)
+
+
+@pytest.mark.parametrize("dt", _all_dtypes)
+def test_searchsorted_py_scalars(dt):
+    q = get_queue_or_skip()
+    skip_if_dtype_not_supported(dt, q)
+
+    x = dpt.zeros(10, dtype=dt, sycl_queue=q)
+    py_zeros = (
```
antonwolfy: Would it make sense to parametrize `py_zeros` with below values instead of iterating over the loop `for sc in py_zeros`?

dpctl/tensor/_searchsorted.py
ndgrigorian: this is not meant to be an `elif`. `x2` is taken to `dt2` first if not `usm_ndarray` then is cast to `dt` which is the resolved dtype. This way, it passes through `dt2` under all circumstances.

dpctl/tensor/_searchsorted.py
ndgrigorian: good catch, this has been changed

dpctl/tests/test_usm_ndarray_searchsorted.py
ndgrigorian: applied this change

dpctl/tensor/_searchsorted.py
```diff
@@ -100,7 +120,7 @@ def searchsorted(
                 "Sorter array must be one-dimension with the same "
                 "shape as the first argument array"
             )
-        res = empty(x1.shape, dtype=x1_dt, usm_type=x1.usm_type, sycl_queue=q)
+        res = empty(x1.shape, dtype=x1_dt, usm_type=x1_usm_type, sycl_queue=q)
```
antonwolfy, line 123: Isn't that exactly the same as below?
```suggestion
        res = empty_like(x1)
```

dpctl/tensor/_searchsorted.py
antonwolfy: Then it's unclear for me, why do we need to run 2 kernels in that case?
1.  to convert a scalar to a temporary usm_ndarray and to copy back to the result
2. to cast the array to `dt`

What is a drawback to pass `dt` directly to `asarray` call?

antonwolfy: Also, in the sequence:
1. cast `x1` to `dt`
2. cast `x2` to usm_ndarray if any
3. cast `x2` to `dt`

We have to make step 2 as the 1st or as the last in the sequence, but not in the middle, because currently step 2 depends on events from step 1, but shouldn't.
And from other side, step 2 produces depending events which we must pass to step 3, but we didn't (`asarray` must be completed before we can start casting `x2` to `dt`).

ndgrigorian: > Then it's unclear for me, why do we need to run 2 kernels in that case?
> 
> 1. to convert a scalar to a temporary usm_ndarray and to copy back to the result
> 2. to cast the array to `dt`
> 
> What is a drawback to pass `dt` directly to `asarray` call?

with these Python scalars, it's what we've been doing in general (see: `BinaryElementwiseFunc` for example) and on top of this, it avoids possible undefined behavior. I can't immediately come up with examples, but it can be the case that
```
a = dpt.asarray(sc, dtype=dt)
```
is not necessarily the same as
```
tmp = dpt.asarray(sc, dtype=o_dt)
a = dpt.astype(tmp, dt)
```
but practically speaking we can reassess if any of these cases come up now that we've made some changes (like with unsigned ints)

ndgrigorian: > Also, in the sequence:
> 
> 1. cast `x1` to `dt`
> 2. cast `x2` to usm_ndarray if any
> 3. cast `x2` to `dt`
> 
> We have to make step 2 as the 1st or as the last in the sequence, but not in the middle, because currently step 2 depends on events from step 1, but shouldn't. And from other side, step 2 produces depending events which we must pass to step 3, but we didn't (`asarray` must be completed before we can start casting `x2` to `dt`).

great point, I'll adjust the dependency logic

ndgrigorian:
closing as this is no longer relevant with tensor migration

this work is being migrated to `dpnp`

PR #2229, opened 2026-01-14
Fix `dpt.repeat()` regression for 1D repeats array

vlad-perevezentsev:
After #2223 `dpt.repeat()` started raising an error when `repeats` is as 1D `usm_ndarray` 
```
import dpctl.tensor as dpt

a = dpt.arange(8)
dpt.repeat(a, ind)

# repeats = int(repeats)
# TypeError: only 0-dimensional arrays can be converted to Python scalars

```
Previously this worked because `int(repeats)` was implicitly allowed for 1D usm_ndarrays

This PR updates `dpt.repeat()` to explicitly extract the single element from size-1 1D usm_ndarray repeats before converting it to a Python scalar and extends tests to cover both 0D and 1D repeats cases

### DIFF
--- a/dpctl/tensor/_manipulation_functions.py
+++ b/dpctl/tensor/_manipulation_functions.py
@@ -829,7 +829,12 @@ def repeat(x, repeats, /, *, axis=None):
         if repeats.size == 1:
             scalar = True
             # bring the single element to the host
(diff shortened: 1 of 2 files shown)


ndgrigorian (approved):
LGTM!

PR #2250, opened 2026-02-17
Remove `python-gil` requirement and enable free-threaded Python

ndgrigorian:
This PR removes the required `python-gil` dependency from the dpctl conda package workflow and enables free-threaded Python in extension modules

Adjustments are made to the `SequentialOrderManager` class such that the class is safe in free-threaded, including mutexes in the C++ class as a fall-back in case of (not recommended) simultaneous access to its members and methods

SequentialOrderManager now maintains thread-local storage for individual queue-to-manager-maps, such that each thread has its own manager per queue.

Queue and device caching, meanwhile, are global, to create a concept of default queues and devices that allows operations in extensions like dpnp (which rely on queues being the same for compute follows data) to operate on data passed between threads without copy.

In the futue, per-thread-queues and devices may prove more efficient, in which case, extensions will be asked to be made more robust (checking that context and device are the same, not using queue as a shortcut).

This PR builds on top of work already done removing the tensor submodule, which is pending migration to dpnp

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -28,7 +28,6 @@ requirements:
         - {{ compiler('dpcpp') }} >={{ required_compiler_version }}
     host:
         - python
-        - python-gil     # [py>=314]
         - pip >=24.0
         - level-zero-devel >=1.16
         - pybind11 >=2.12
@@ -52,7 +51,6 @@ requirements:
         - tomli # [py<311]
     run:
         - python
-        - python-gil     # [py>=314]
         - {{ pin_compatible('intel-sycl-rt', min_pin='x.x', max_pin='x') }}
         - {{ pin_compatible('intel-cmplr-lib-rt', min_pin='x.x', max_pin='x') }}
         - numpy
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -63,7 +63,6 @@ test:
         - cython
         - setuptools
         - pytest
-        - pytest-cov
 
 about:
     home: https://github.com/IntelPython/dpctl.git
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -30,7 +30,7 @@ requirements:
         - python
         - pip >=24.0
         - level-zero-devel >=1.16
-        - pybind11 >=2.12
+        - pybind11 >=2.13
         - {{ pin_compatible('intel-sycl-rt', min_pin='x.x', max_pin='x') }}
         - {{ pin_compatible('intel-cmplr-lib-rt', min_pin='x.x', max_pin='x') }}
         # Ensure we are using latest version of setuptools, since we don't need
--- a/dpctl/_sycl_device_factory.pyx
+++ b/dpctl/_sycl_device_factory.pyx
@@ -17,6 +17,7 @@
 # distutils: language = c++
 # cython: language_level=3
 # cython: linetrace=True
+# cython: freethreading_compatible = True
 
 """ This module implements several device creation helper functions:
 
--- a/dpctl/_sycl_device_factory.pyx
+++ b/dpctl/_sycl_device_factory.pyx
@@ -19,7 +19,8 @@
 # cython: linetrace=True
 # cython: freethreading_compatible = True
 
-""" This module implements several device creation helper functions:
+"""
+This module implements several device creation helper functions:
 
   - wrapper functions to create a SyclDevice from the standard SYCL
     device selector classes.
@@ -47,7 +48,7 @@ from ._backend cimport (  # noqa: E211
     _device_type,
 )
 
-from contextvars import ContextVar
+import threading
 
 from ._sycl_device import SyclDeviceCreationError
 from .enum_types import backend_type
@@ -289,7 +290,8 @@ cpdef int get_num_devices(
 
 
 cpdef cpp_bool has_cpu_devices():
-    """ A helper function to check if there are any SYCL CPU devices available.
+    """
+    A helper function to check if there are any SYCL CPU devices available.
 
     Returns:
         bool:
@@ -301,7 +303,8 @@ cpdef cpp_bool has_cpu_devices():
 
 
 cpdef cpp_bool has_gpu_devices():
-    """ A helper function to check if there are any SYCL GPU devices available.
+    """
+    A helper function to check if there are any SYCL GPU devices available.
 
     Returns:
         bool:
@@ -313,7 +316,8 @@ cpdef cpp_bool has_gpu_devices():
 
 
 cpdef cpp_bool has_accelerator_devices():
-    """ A helper function to check if there are any SYCL Accelerator devices
+    """
+    A helper function to check if there are any SYCL Accelerator devices
     available.
 
     Returns:
@@ -328,7 +332,8 @@ cpdef cpp_bool has_accelerator_devices():
 
 
 cpdef SyclDevice select_accelerator_device():
-    """A wrapper for ``sycl::device{sycl::accelerator_selector_v}`` constructor.
+    """
+    A wrapper for ``sycl::device{sycl::accelerator_selector_v}`` constructor.
 
     Returns:
         dpctl.SyclDevice:
@@ -350,7 +355,8 @@ cpdef SyclDevice select_accelerator_device():
 
 
 cpdef SyclDevice select_cpu_device():
-    """A wrapper for ``sycl::device{sycl::cpu_selector_v}`` constructor.
+    """
+    A wrapper for ``sycl::device{sycl::cpu_selector_v}`` constructor.
 
     Returns:
         dpctl.SyclDevice:
@@ -372,7 +378,8 @@ cpdef SyclDevice select_cpu_device():
 
 
 cpdef SyclDevice select_default_device():
-    """A wrapper for ``sycl::device{sycl::default_selector_v}`` constructor.
+    """
+    A wrapper for ``sycl::device{sycl::default_selector_v}`` constructor.
 
     Returns:
         dpctl.SyclDevice:
@@ -394,7 +401,8 @@ cpdef SyclDevice select_default_device():
 
 
 cpdef SyclDevice select_gpu_device():
-    """A wrapper for ``sycl::device{sycl::gpu_selector_v}`` constructor.
+    """
+    A wrapper for ``sycl::device{sycl::gpu_selector_v}`` constructor.
 
     Returns:
         dpctl.SyclDevice:
@@ -417,21 +425,23 @@ cpdef SyclDevice select_gpu_device():
 
 cdef class _DefaultDeviceCache:
     cdef dict __device_map__
+    cdef object _cache_lock
 
     def __cinit__(self):
         self.__device_map__ = {}
+        self._cache_lock = threading.Lock()
 
-    cdef get_or_create(self):
-        """Return instance of SyclDevice and indicator if cache
-        has been modified"""
-        key = 0
-        if key in self.__device_map__:
-            return self.__device_map__[key], False
-        dev = select_default_device()
-        self.__device_map__[key] = dev
-        return dev, True
+    def get_or_create(self):
+        """Return cached default SyclDevice, creating it if needed."""
+        with self._cache_lock:
+            key = 0
+            if key in self.__device_map__:
+                return self.__device_map__[key]
+            dev = select_default_device()
+            self.__device_map__[key] = dev
+            return dev
 
-    cdef _update_map(self, dev_map):
+    def _update_map(self, dev_map):
         self.__device_map__.update(dev_map)
 
     def __copy__(self):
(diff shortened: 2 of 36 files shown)


dpctl/utils/src/sequential_order_keeper.hpp
```diff
@@ -176,8 +186,20 @@ class SequentialOrder
 
     void wait()
     {
-        sycl::event::wait(submitted_events);
-        sycl::event::wait(host_task_events);
-        prune_complete();
+        // snapeshot events outside of mutex to avoid
```
vlad-perevezentsev:
```suggestion
        // snapshot events outside of mutex to avoid
```

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -142,8 +145,33 @@ class dpctl_capi
...
+        dpctl_capi *capi_ptr = global_capi.load(std::memory_order_acquire);
+
+        // fast path
+        if (capi_ptr) {
+            return *capi_ptr;
+        }
+
+        static std::mutex init_mtx;
+
+        // release gil while holding lock
```
vlad-perevezentsev: If I understand correctly you release GIL before acquiring the mutex then reacquire it inside the lock to prevent a deadlock between threads?
Could you add a brief comment explaining this?

dpctl/_sycl_queue.pxd
```diff
@@ -43,6 +43,7 @@ cdef public api class _SyclQueue [
     cdef DPCTLSyclQueueRef _queue_ref
     cdef SyclContext _context
     cdef SyclDevice _device
+    cdef object __weakref__
```
vlad-perevezentsev: What is the reason for adding this?

ndgrigorian: left over from previous implementation of the updated cache, I'll drop it

dpctl/apis/include/dpctl4pybind11.hpp
ndgrigorian: yep that's more or less correct, I'll adjust the comment.

dpctl/_sycl_queue_manager.pyx
```diff
@@ -56,32 +58,34 @@ cdef class _DeviceDefaultQueueCache:
...
             raise TypeError
-        if ctx_dev in self.__device_queue_map__:
-            return self.__device_queue_map__[ctx_dev], False
-        if q is None:
-            q = SyclQueue(*ctx_dev)
-        self.__device_queue_map__[ctx_dev] = q
-        return q, True
-
-    cdef _update_map(self, dev_queue_map):
+        with self._cache_lock:
```
vlad-perevezentsev, line 61: I am wondering if it is safe to create `q = SyclQueue(key)` above outside `self._cache_lock`?
What happens if multiple threads will use  `get_or_create` for one device using `SyclDevice` or `srt` as the key?

dpctl/tests/test_sycl_queue_manager.py
```diff
@@ -24,19 +24,20 @@
...
     try:
         d = dpctl.SyclDevice()
     except dpctl.SyclDeviceCreationError:
         pytest.skip("Could not create default device")
 
     q1 = get_device_cached_queue(d)
-    cache_copy = copy.copy(cache.get())
-    q2, changed = cache_copy.get_or_create(d)
+    cache_copy = copy.copy(_global_device_queue_cache)
+    q2 = cache_copy.get_or_create(d)
```
vlad-perevezentsev, line 39: it possible to add tests to check for multithreading?

ndgrigorian: There are the new examples, which I am using for tests for now. I think we can add tests in a follow-up PR, but as this is a beta version of free-threaded support, I think it should be ok to follow up.

dpctl/_sycl_queue_manager.pyx
ndgrigorian: it shouldn't be a problem, multiple queues would be created, but only one will be cached, the others will essentially be discarded.

We could go back and refactor it a bit later to improve efficiency somehow (would need to do some work designing the mechanism for it) but there should be nothing unsafe

vlad-perevezentsev (dismissed):
No more comments from my side
LGTM
Thank you @ndgrigorian

pyproject.toml
vchamarthi: We may need to bump cython minimum version to 3.1.0 since free threaded [directives added in 3.1.0](https://github.com/cython/cython/releases/tag/3.1.0)

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -128,8 +131,35 @@ class dpctl_capi
...
+        std::lock_guard<std::mutex> lock(init_mtx);
+
+        // double check after acquiring lock
+        capi_ptr = global_capi.load(std::memory_order_relaxed);
+
+        if (!capi_ptr) {
+            // acquire gil to safely call into Python C API
+            py::gil_scoped_acquire acquire;
+
+            capi_ptr = new dpctl_capi();
```
antonwolfy, line 159: Does switching from a function-local static to a heap pointer means `~dpctl_capi()` and the whole `Deleter` finalization guard (lines 170-186) are now dead code — the held Python objects leak?
I guess leaking a process singleton is a defensible no-GIL trade-off, but it bypasses machinery that was deliberately written for interpreter finalization.

Do we need to add an explicit comment stating it's intentional?

dpctl/utils/_order_manager.py
```diff
@@ -16,7 +16,7 @@ class _SequentialOrderManager:
     def __init__(self):
         self._state = _OrderManager(16)
 
-    def __dealloc__(self):
+    def __del__(self):
```
antonwolfy, line 20: `__dealloc__` was dead on a pure-Python class; renaming to `__del__` makes it actually call `SyclEvent.wait_for(...)`.

Under the free-threading, finalizers run on arbitrary threads and possibly during interpreter shutdown — combined with the existing `weakref.finalize` path, two finalizers now wait on events.

Confirm this is intended and that it's guarded against shutdown.

examples/python/using_order_manager.py
```diff
@@ -0,0 +1,199 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Demonstrates using the SequentialOrderManager with USM and the Python threading
+module.
+"""
+
+import concurrent.futures
```
antonwolfy, line 20: `SequentialOrderManager`'s map is `threading.local()`, so each worker thread gets a fresh, thread-private order manager.

The `independent_threads`/`fork_join` cases pass only because each thread does its own blocking wait() — the order manager is effectively a no-op for cross-thread coordination. Only `explicit_event_passing` does genuine cross-thread handoff (via explicit `add_event_pair`).

It seems we need at least to reframe the docstrings to say each thread keeps its own sequential order and cross-thread deps must be passed explicitly.

examples/python/using_order_manager.py
```diff
@@ -0,0 +1,199 @@
+# Copyright 2020-2025 Intel Corporation
```
antonwolfy: Seems we haven't to use the range:
```suggestion
# Copyright 2026 Intel Corporation
```

pyproject.toml
```diff
@@ -60,7 +61,7 @@ readme = {file = "README.md", content-type = "text/markdown"}
 requires-python = ">=3.10"
 
 [project.optional-dependencies]
-coverage = ["Cython", "pytest", "pytest-cov", "coverage", "tomli"]
+coverage = ["Cython", "pytest", "coverage", "tomli"]
```
antonwolfy: What is about the `--cov-report term-missing` below?
Actually `ini.options` has to be renamed to `ini_options` and it's only the reason why pytest is not failing with that change (removed pytest-cov).

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -128,8 +131,35 @@ class dpctl_capi
...
+        if (capi_ptr) {
+            return *capi_ptr;
+        }
+
+        static std::mutex init_mtx;
+
+        // initialization requires calls into Python C-API, so initializing
+        // thread must hold the GIL, while other threads must not block on the
+        // mutex while holding the GIL, as this creates a deadlock
+        py::gil_scoped_release release;
```
antonwolfy, line 147: On the slow path it unconditionally constructs `py::gil_scoped_release`. The above comment assumes the initializing thread holds the GIL, but nothing enforces it.
If `get()` is ever first reached from a thread that doesn't hold the GIL (a C++/SYCL worker, or after a `gil_scoped_release` — note `SequentialOrder::wait` is bound with `call_guard<gil_scoped_release>()`), releasing an unheld GIL is a fatal abort. The old `static dpctl_capi api{}` had no such precondition.

pyproject.toml
antonwolfy: Should we also limit to use `cython>=3.1.0` for optional deps?

dpctl/utils/src/device_queries.cpp
```diff
@@ -132,7 +132,7 @@ std::uint32_t py_intel_memory_bus_width(const sycl::device &d)
 
 }; // namespace
 
-PYBIND11_MODULE(_device_queries, m)
+PYBIND11_MODULE(_device_queries, m, py::mod_gil_not_used())
```
antonwolfy, line 135: Should we update the example in `docs/doc_sources/api_reference/dpctl_pybind11.rst` with the same `py::mod_gil_not_used()`?

dpctl/utils/src/sequential_order_keeper.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <mutex>
```
antonwolfy, line 6: Missing `<utility>` include

dpctl/_sycl_device_factory.pyx
```diff
@@ -414,46 +423,45 @@ cpdef SyclDevice select_gpu_device():
...
+        """Return cached default SyclDevice, creating it if needed."""
+        with self._cache_lock:
+            key = 0
+            if key in self.__device_map__:
+                return self.__device_map__[key]
+            dev = select_default_device()
+            self.__device_map__[key] = dev
+            return dev
+
+    def _update_map(self, dev_map):
```
antonwolfy, line 444: `_update_map` is public + unlocked — safe today (only called on the not-yet-shared `_copy`), but a foot-gun if ever called on the live global. We probably might need to consider locking it internally or clear documenting.

dpctl/_sycl_device_factory.pyx
```diff
@@ -414,46 +423,45 @@ cpdef SyclDevice select_gpu_device():
...
         self.__device_map__.update(dev_map)
 
     def __copy__(self):
```
antonwolfy, line 447: Seems not covered by any test

dpctl/_sycl_queue_manager.pyx
```diff
@@ -56,32 +58,34 @@ cdef class _DeviceDefaultQueueCache:
...
+            if ctx_dev in self.__device_queue_map__:
+                return self.__device_queue_map__[ctx_dev]
+            if q is None:
+                q = SyclQueue(*ctx_dev)
+            self.__device_queue_map__[ctx_dev] = q
+            return q
+
+    def _update_map(self, dev_queue_map):
```
antonwolfy, line 69: Same comment to `_update_map`

dpctl/_sycl_device_factory.pyx
antonwolfy: Or alternatively to revert back to `cdef` at least to make it inaccessible from the python code, i.e. to have more control when the method is called.

pyproject.toml
ndgrigorian: in theory it doesn't matter, because the tests it's used in don't use free-threading, but we could pin it regardless

examples/python/using_order_manager.py
ndgrigorian: We can add this in a follow-up as documentation on how to interact with free-threading, a dedicated page even.

For now, this was the purpose of the cross-thread example, to actually showcase how per-thread order manager works more clearly than in a docstring.

dpctl/apis/include/dpctl4pybind11.hpp
ndgrigorian: I believe that the deleter is still called at shutdown by pybind11. I'll add a comment.

ndgrigorian: noting that the deleter actually isn't called, but regardless, adding comment

dpctl/apis/include/dpctl4pybind11.hpp
ndgrigorian: it should not be possible to reach without the GIL. It would only be possible if `dpctl_capi::get()` was called by a thread not owned by Python, which I don't believe to be an intended use-case, as this is only meant to be used for initializing pybind11 casting

dpctl/_sycl_device_factory.pyx
ndgrigorian: I don't think there's much reason to make such a change, the method is not meant to be called by users and is an internal method of an internal class.

dpctl/_sycl_device_factory.pyx
ndgrigorian: it is just call to dict.update, I don't think it needs a test

pyproject.toml
```diff
@@ -125,16 +126,16 @@ multi_line_output = 3
...
 addopts = [
   "--junitxml=junit.xml",
-  "--ignore setup.py",
-  "--ignore run_test.py",
-  "--cov-report term-missing",
-  "--tb native",
+  "--ignore=setup.py",
+  "--ignore=run_test.py",
+  "--tb=native",
   "--strict",
```
antonwolfy: `--strict` is a deprecated alias for `--strict-markers`:
```suggestion
  "--strict-markers",
```

conda-recipe/meta.yaml
```diff
@@ -28,7 +28,6 @@ requirements:
         - {{ compiler('dpcpp') }} >={{ required_compiler_version }}
     host:
         - python
-        - python-gil     # [py>=314]
         - pip >=24.0
         - level-zero-devel >=1.16
         - pybind11 >=2.12
```
antonwolfy: `mod_gil_not_used()` requires 2.13 at least

antonwolfy (dismissed):
Thank you @ndgrigorian, LGTM!
I left a couple minor nits more:

antonwolfy (approved):

PR #2273, opened 2026-04-02
task: syclqueue.copy() method

jharlow-intel:
This PR adds the python interface to the sycl runtimes copy methods and tests. I tried to follow the format of memcpy best I could

I used an LLM primarily to generate this, and then reviewed it and tweaked it by hand. My first time contributing a new method to DPCTL so apologies if I overlooked some obvious mistakes in it. I'm here to learn!


dpctl/_sycl_queue.pyx
```diff
@@ -1426,6 +1426,76 @@ cdef class SyclQueue(_SyclQueue):
...
 
+    cpdef copy(self, dest, src, size_t count):
+        """Copy ``count`` bytes from ``src`` to ``dest`` and wait.
+
+        This is a synchronizing variant corresponding to
+        :meth:`dpctl.SyclQueue.copy_async`.
+        """
+        cdef DPCTLSyclEventRef ERef = NULL
+
+        ERef = _memcpy_impl(<SyclQueue>self, dest, src, count, NULL, 0)
```
ndgrigorian: Since this calls `_memcpy_impl` it is actually wrapping `memcpy`, not `copy`, i.e., https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#api:queue-copy

The goal here should be to wrap the `copy` call from `sycl::queue`, which may seem redundant but, may have use (there have been cases in the past where `memcpy` breaks, for example). Users may want access to this specific API for some specific reason.

This means that the `copy` implementations will need to be added to the C API, probably `libsyclinterface/source/dpctl_sycl_queue_interface.cpp` and the respective `libsyclinterface/include/syclinterface/dpctl_sycl_queue_interface.h`.

jharlow-intel: Yeah, Codex really dropping the ball and editing functions and comment it shouldn't, I'm leaving this PR a draft for now as I pick through it more finely, don't bother reviewing this until I know it's un-slopified and remove the draft tag and in-progress label.

Only opened the PR this early because I wanted the internal CI GPU testing

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -41,6 +41,7 @@
 #include <stdexcept>
 #include <sycl/sycl.hpp> /* SYCL headers   */
 #include <utility>
+#include <vector>
```
jharlow-intel, line 44: seems like the LLM picked up that this should have been included in the first place for this file

jharlow-intel:
@antonwolfy @ndgrigorian @vlad-perevezentsev  ready for review

ndgrigorian:
@jharlow-intel 
please merge in the master branch, or rebase onto it, it will make it much easier to test and iterate on this branch, with tensor removed (build time is ~5 minutes compared to over an hour)

jharlow-intel:
@ndgrigorian merged master into this branch, let me know what else you might need to help push this review along

jharlow-intel:
@ndgrigorian @antonwolfy ping (:

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
ndgrigorian: looks like it was right

dpctl/_sycl_queue.pyx
```diff
@@ -531,6 +533,82 @@ cdef DPCTLSyclEventRef _memcpy_impl(
     return ERef
 
 
+cdef DPCTLSyclEventRef _copy_impl(
```
ndgrigorian: this and `_memcpy_impl` are effectively the same thing, but one calls C-API for `sycl_queue::copy` and the other `sycl_queue::memcpy`. It may be worth considering if we can refactor even more, passing the function pointers to an impl func

dpctl/_sycl_queue.pyx
```diff
@@ -1421,6 +1499,82 @@ cdef class SyclQueue(_SyclQueue):
 
         return SyclEvent._create(ERef)
 
+    cpdef copy(self, dest, src, size_t count):
```
ndgrigorian: one of the advantages of copy vs memcpy is that copy permits typed pointers instead of only `void`.

We could have an optional arg that passes a numpy style dtype str (see WorkGroupMemory implementation) and checks that both pointers are appropriately sized within the type itself, before passing the the void ptr copies

If not in this PR, a TODO for it would be nice, as somewhere down the line I believe we need `dpctl.dtype`/`dpctl.types` to handle our various typing issues on Cython/Python level

dpctl/_sycl_queue.pyx
jharlow-intel: Done! Refactored this into more consolidated implementations

dpctl/_sycl_queue.pyx
jharlow-intel: Did my best to implement this, with size checking and tests to accompany that.

dpctl/tests/test_sycl_queue_copy.py
```diff
@@ -0,0 +1,282 @@
...
+    try:
+        q = dpctl.SyclQueue()
+    except dpctl.SyclQueueCreationError:
+        pytest.skip("Default constructor for SyclQueue failed")
+
+    num_elements = 16
+    nbytes = num_elements * element_size
+
+    src = dpctl.memory.MemoryUSMShared(nbytes, queue=q)
+    dst = dpctl.memory.MemoryUSMShared(nbytes, queue=q)
```
ndgrigorian: can we also parametrize this test to use `MemoryUSMHost` and `MemoryUSMDevice`? Could be done with a simple lambda for each USM type

dpctl/_sycl_queue.pyx
```diff
@@ -475,6 +511,15 @@ cdef DPCTLSyclEventRef _memcpy_impl(
...
     cdef int ret_code = 0
+    cdef size_t element_size = 0
+
+    if dtype is not None:
+        element_size = _get_dtype_size(dtype)
+        if byte_count % element_size != 0:
+            raise ValueError(
+                f"byte_count ({byte_count}) must be a multiple of dtype "
+                f"element size ({element_size} bytes for '{dtype}')"
+            )
```
ndgrigorian: we could move this into _copy_impl and drop `dtype` as a parameter in `_copy_memcpy_impl` given that it isn't used for anything but validation

dpctl/tests/test_sycl_queue_copy.py
```diff
@@ -0,0 +1,305 @@
...
+import dpctl.memory
+
+
+def _create_memory(q):
+    nbytes = 1024
+    mobj = dpctl.memory.MemoryUSMShared(nbytes, queue=q)
+    return mobj
+
+
+def test_copy_copy_host_to_host():
```
vlad-perevezentsev: It should be `test_copy_host_to_host`, right?

dpctl/_sycl_queue.pyx
```diff
@@ -1421,6 +1495,101 @@ cdef class SyclQueue(_SyclQueue):
...
+                for idx, de in enumerate(dEvents):
+                    if isinstance(de, SyclEvent):
+                        depEvents[idx] = (<SyclEvent>de).get_event_ref()
+                    else:
+                        free(depEvents)
+                        raise TypeError(
+                            "A sequence of dpctl.SyclEvent is expected"
+                        )
+            ERef = _copy_impl(self, dest, src, count, depEvents, nDE, dtype)
+            free(depEvents)
```
vlad-perevezentsev: if  `_copy_impl` returns `ValueError` on line above , this line never executes and  `depEvents` leaks
I think it's better to use `try/finally` blocks for this

ndgrigorian: this is true in a few other places as well and needs to be resolved, it would be good to fix it in `memcpy` and `copy` here

jharlow-intel: Super good catch team, will fix it here and double check to make sure memory leaks won't occur elsewhere in the code I've added

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -693,6 +694,70 @@ DPCTLQueue_MemcpyWithEvents(__dpctl_keep const DPCTLSyclQueueRef QRef,
...
+            if (DepEvents) {
+                dep_events.reserve(DepEventsCount);
+                for (size_t i = 0; i < DepEventsCount; ++i) {
+                    event *ei = unwrap<event>(DepEvents[i]);
+                    if (ei)
+                        dep_events.push_back(*ei);
+                }
+            }
+
+            // Bind queue::copy with uint8_t so Count is interpreted as bytes.
```
ndgrigorian: nit of all nits:
this comment is accurate but made me do a doubletake on re-reading the code, because the grammar is a bit unclear: it's saying that we will use `sycl_queue::copy` with `uint8_t` so that count is the number of bytes. Saying `Count is interpreted as bytes` can easily be thought of as instead reinterpreting Count into bytes

just a simple rephrasing would do for both comments

dpctl/_sycl_queue.pyx
```diff
@@ -531,6 +566,45 @@ cdef DPCTLSyclEventRef _memcpy_impl(
...
+        q, dst, src, byte_count, dep_events, dep_events_count,
+        DPCTLQueue_Memcpy, DPCTLQueue_MemcpyWithEvents
+    )
+
+
+cdef DPCTLSyclEventRef _copy_impl(
+     SyclQueue q,
+     object dst,
+     object src,
+     size_t byte_count,
```
ndgrigorian: the `queue::copy` is equivalent to `handler::copy`, which says in [the table](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.members.handler.copy):

`Copies count elements of type T from the pointer src to the pointer dest.`

this means that `count` is expected to be an element count, not a byte count. It's only a byte count for byte-sized data.

Perhaps the best way to approach this: do not allow dtype to be None (including in the copy function itself), and just default dtype to "u1" for uint8_t, making it clear for users that in the default case, you are copying bytes

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
jharlow-intel: Good point, rephrased it to remove the ambiguity

dpctl/_sycl_queue.pyx
jharlow-intel: Done, dtype now defaults to u1.

dpctl/_sycl_queue.pyx
vlad-perevezentsev, line 546: While reviewing this function I noticed a possible issue
If `src` is a buffer object and `dst` fails the type check then `src_buf_view` is not released

This is not related to the current changes but it would be good to fix.
@ndgrigorian  could you confirm if this is correct?

ndgrigorian: yep that seems accurate, @jharlow-intel we should add to the type check logic to release the `src` buffer if it's a buffer, in this case.

jharlow-intel: I gotchu, will fix this

jharlow-intel:
3.14 test errors are not related to this PR

ndgrigorian (approved):
LGTM, thank you @jharlow-intel

PR #2291, opened 2026-04-23
Rename `SyclProgram` to `SyclKernelBundle`

ndgrigorian:
This PR renames `SyclProgram` to `SyclKernelBundle` and deprecates the `SyclProgram` class

`sycl::program` hasn't been used in SYCL or libsyclinterface in a very long time, it is sensible to now move away from the `program` terminology entirely, as it is no longer a part of the SYCL spec.

The submodule will likely be moved/renamed in a future PR.

### DIFF
--- a/dpctl/program/_program.pyx
+++ b/dpctl/program/_program.pyx
@@ -18,14 +18,17 @@
 # cython: language_level=3
 # cython: linetrace=True
 
-"""Implements a Python interface for SYCL's program and kernel runtime classes.
+"""Implements a Python interface for SYCL's kernel bundle and kernel runtime
+classes.
 
-The module also provides functions to create a SYCL program from either
-a OpenCL source string or a SPIR-V binary file.
+The module also provides functions to create a SYCL kernel bundle from either
+an OpenCL source string or a SPIR-V binary file.
 
 """
 
 from libc.stdint cimport uint32_t
+
+import warnings
 
 from dpctl._backend cimport (  # noqa: E211, E402;
     DPCTLKernel_Copy,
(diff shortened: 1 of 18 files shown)


ndgrigorian:
@tdeng5

dpctl/program/_program.pyx
```diff
@@ -315,25 +340,36 @@ cpdef create_program_from_spirv(SyclQueue q, const unsigned char[:] IL,
...
-    return SyclProgram._create(KBref)
+    return SyclKernelBundle._create(KBref)
 
 
-cdef api DPCTLSyclKernelBundleRef SyclProgram_GetKernelBundleRef(
-    SyclProgram pro
-):
-    """ C-API function to access opaque kernel bundle reference from
-    Python object of type :class:`dpctl.program.SyclKernel`.
+cpdef create_program_from_source(SyclQueue q, str src, str copts=""):
```
vlad-perevezentsev, line 348: Does it make sense to implement the same logic for `SyclProgramCompilationError`?
Currently it is just removed without any alias

ndgrigorian: it's a bit of a strange situation, because we do export it in `__init__`, so it is public facing, but no function actually raises it after these changes, and anything designed to `try...except` would break if we re-export it as an alias anyway, like with `SyclProgram`.

I think re-exporting like `SyclProgram` is fine though, I'll add it.

ndgrigorian:
@antonwolfy @vlad-perevezentsev 
ping

vlad-perevezentsev (approved):
No more comments from my side
Overall LGTM
Thank you @ndgrigorian

PR #2292, opened 2026-04-23
Restructure Python examples

ndgrigorian:
This PR proposes a restructuring of the Python examples, specifically the USM-related examples

These examples were each very simple scripts, all relating to various operations with USM: allocation, host-access, and copying between host and device. By consolidating them, they can be run using the same `_runner` construct as the other examples.

As a result of this, the Conda package workflow's run of examples is updated to use the `runner` construct. This also means that all of the examples are being run, instead of just the USM examples.


examples/python/usm_operations.py
```diff
@@ -0,0 +1,140 @@
...
+    mhv[:] = msv[78 : 78 + len(mh)]
+
+    print("Byte-values of the USM-host buffer")
+    print(list(mhv))
+
+    # USM-device buffer is not host accessible
+    md = dpmem.MemoryUSMDevice(16)
+    try:
+        memoryview(md)
+    except Exception as e:
```
vlad-perevezentsev: In `subdevices.py` you narrowed `except Exception` to `except dpctl.SyclSubDeviceCreationError`
Should this similarly also `except ValueError` according to[ 278 line](https://github.com/IntelPython/dpctl/blob/master/dpctl/memory/_memory.pyx#L278) in _memory.pyx?

examples/python/usm_operations.py
```diff
@@ -0,0 +1,140 @@
+#                      Data Parallel Control (dpctl)
+#
+# Copyright 2020-2025 Intel Corporation
```
vlad-perevezentsev, line 3:
```suggestion
# Copyright 2020-2026 Intel Corporation
```

vlad-perevezentsev (dismissed):
LGTM
Thank you @ndgrigorian 

Just a few minor comments below:

examples/python/usm_operations.py
ndgrigorian: good point!

examples/python/usm_operations.py
ndgrigorian: I will make a follow-up soon to change all license headers to 2020

vlad-perevezentsev (approved):
LGTM
Thanks @ndgrigorian

Issue #2293, opened 2026-04-23
(Question) Does `SequentialOrderManager` track externally-submitted work on a shared `sycl::queue`


abagusetty:
@ndgrigorian @vlad-perevezentsev @antonwolfy @alvarovm

Just wanted to reach out related to a setup we have been working with an application using cupy and dpnp (porting).

**App**: Heavily uses cupy and hand-written CUDA kernels. These were ported to DPNP and SYCL kernels for Intel PVC (on Aurora). 

**Intent**: The app relies on the use of `stream = cupy.cuda.get_current_stream()` , `stream.synchronize()`, and passing the stream obtained by the above to pass it down to user CUDA kernels. A pipeline methodology where a given instance/process is tied to a single stream through the upstream cupy to downstream user CUDA kernels. 

With the port to DPNP & SYCL (currently), we have relied on using a single queue written for user kernels and dpnp/dpctl ops wasnt aware of this queue. Since they were all created with the same default context (using `ZE_SERIALIZE=2, OverrideImmediateCmdListSynchronousMode=1`) helped with the initial correctness testing. Since everything got synchronized and default-context shared by any number of queues wasnt a problem.

**Goal**: The intent now is to move from the above design and explore if upstream `dpctl` provides us a single master (in-order) queue per-device `q = dpctl.SyclQueue(devs[device_id], property="in_order")` such that we can use it to the downstream user SYCL kernels as well. Since everything is in-order we expect similar FIFO behavior as `cudaStream`

Few questions if you can provide any pointers:
0. Would the above mentioned design be feasible. 
1. Since there is no such equivalent as `stream = cupy.cuda.get_current_stream()`, we have tried something like `from ._sycl_queue_manager import get_device_cached_queue` but the obtained queue from `get_device_cached_queue` is inherently an OoO queue which doesnt really suite our broader needs in the user-app. Since we are trying to mimic the infrastructure based on `cudaStream` FIFO. So we had to start using `q = dpctl.SyclQueue(devs[device_id], property="in_order")`. 
2. Does `SequentialOrderManager` track externally-submitted work on a dpctl obtained `sycl::queue`

More details:
Nothing in SequentialOrder observes the queue itself. So if a user extension does:
```bash
pythonq = dpctl.SyclQueue()
a = dpt.arange(..., sycl_queue=q)          # dpctl registers events
my_ext.launch_kernel(q, a.usm_data, ...)   # user submits directly, NOT registered
[...]

PR #2299, opened 2026-04-26
Use a no-op order manager for in-order queues

ndgrigorian:
As pointed out in #2293, the `SequentialOrderManager` construct may not bring any benefit with in-order queues, and, with an in-order queue, an extension/function implementation has to insert events into the `SequentialOrderManager` if using library calls which also use `SequentialOrderManager` unnecessarily.

To remedy this, this PR proposes adding a `_NoOpOrderManager` implementation class that `SequentialOrderManager` can use when passed an in-order queue.

The returned object isn't cached, as it's mostly a no-op. Notably, the `_NoOpOrderManager` stores the in-order queue that it was given, so that `wait()` can call `q.wait()` so that the method is still meaningful.


ndgrigorian:
@antonwolfy @vlad-perevezentsev 
this is ready for review, since this impacts order manager, it will impact all of tensor and dpnp. On my tests so far, I have seen no issues, but I will keep trying.

Please experiment with it as well

dpctl/utils/_order_manager.py
```diff
@@ -74,11 +108,16 @@ def __init__(self):
             default=defaultdict(_SequentialOrderManager),
         )
 
-    def __getitem__(self, q: SyclQueue) -> _SequentialOrderManager:
+    def __getitem__(
+        self, q: SyclQueue
+    ) -> Union[_SequentialOrderManager, _NoOpOrderManager]:
```
antonwolfy, line 113: Would it make sense to implement a protocol instead of union to have a pure type-checking construct?
```python
from typing import Protocol

class OrderManagerProtocol(Protocol):
    """Protocol defining order manager interface"""
    
    def add_event_pair(self, host_task_ev, comp_ev) -> None: ...
    
    @property
    def num_host_task_events(self) -> int: ...
    
    # ... etc

def __getitem__(self, q: SyclQueue) -> OrderManagerProtocol:
    # type checkers understand this is an interface
```

dpctl/utils/_order_manager.py
```diff
@@ -74,11 +108,16 @@ def __init__(self):
...
         """Get order manager for given SyclQueue"""
-        _local = self._map.get()
         if not isinstance(q, SyclQueue):
             raise TypeError(f"Expected `dpctl.SyclQueue`, got {type(q)}")
+        if q.is_in_order:
+            # we don't need to cache the NoOpOrderManager since it's stateless
+            return _NoOpOrderManager(q)
```
antonwolfy, line 119: Is it expected there will be no `wait()` called for that queue on `clear` method below?

antonwolfy: I wonder about finalization of tasks from in-order queue at exit

dpctl/utils/_order_manager.py
ndgrigorian: we could, though also, we could just use `|` instead of the `Union` since we restrict to >=3.10 anyway.

I'll try the protocol approach and see if it saves some redundancy

dpctl/utils/_order_manager.py
ndgrigorian: > Is it expected there will be no `wait()` called for that queue on `clear` method below?

I think it makes sense in the sense that, no events were present in it to begin with. The original `SequentialOrderManager` calls the wait on its own events, but this order manager, owns no events. But I see your point about the finalization though, it's fair: unlike the normal order manager, this one wouldn't clear.

We will need to track some queues, I think.

PR #2300, opened 2026-04-27
Use work-around for Coverage workflow

ndgrigorian:
Coverage currently segfaults due to segfault when building with oneAPI 2026.0

As a work-around, use 2025.3 until the issue is resolved. Slips in a change bumping coverage workflow to Python 3.14 and making `--cmake-opts` actually work in `gen_coverage` and `gen_docs` scripts, an issue which was observed during reproduction of the crash.

### DIFF
--- a/.github/workflows/generate-coverage.yaml
+++ b/.github/workflows/generate-coverage.yaml
@@ -34,7 +34,8 @@ jobs:
 
       - name: Install latest Intel OneAPI
         run: |
-          sudo apt install intel-oneapi-compiler-dpcpp-cpp
+          # TODO: drop 2025.3 requirement when coverage build segfault is resolved
+          sudo apt install intel-oneapi-compiler-dpcpp-cpp-2025.3
           sudo apt install intel-oneapi-tbb
           sudo apt install intel-oneapi-umf
           sudo apt install hwloc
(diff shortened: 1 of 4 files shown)


.github/workflows/generate-coverage.yaml
```diff
@@ -34,7 +34,8 @@ jobs:
 
       - name: Install latest Intel OneAPI
         run: |
-          sudo apt install intel-oneapi-compiler-dpcpp-cpp
+          # TODO: drop 2025.3 requirement when coverage build segfault is resolved
```
vlad-perevezentsev, line 37: Could you also add the Jira ticket number?

vlad-perevezentsev (approved):
In general , LGTM
Thank you @ndgrigorian

.github/workflows/generate-coverage.yaml
ndgrigorian: I was thinking about this, and in general, I would prefer not to reference our internal JIRA tickets in the code base at all if possible, going forward, as they aren't useful to external users at all.

We can track the issues in JIRA and then create new tickets for ourselves when they're resolved or implemented.

If an issue is expansive (and therefore, some code could be forgotten) we could make a GH issue and flag it in the comments instead.

PR #2303, opened 2026-04-27
Remove work-around in `dpctl.SyclPlatform.get_devices` for `custom` and `automatic` device types

ndgrigorian:
This PR removes a work-around added to `dpctl.SyclPlatform.get_devices` due to a bug in the implementation, where the device types "automatic" and "custom" would return "all"

To work-around, these were treated as "unknown" and returned an empty list instead of a full list

The work-around is no longer necessary


vlad-perevezentsev (approved):
LGTM
Thank you @ndgrigorian

PR #2304, opened 2026-04-28
Add support for specialization constants

ndgrigorian:
This PR introduces support for specialization constants in dpctl, including both a Cython class `SpecializationConstant` for construction and passing of the constructed class to `create_kernel_bundle_from_spirv` via a new `specializations` keyword argument.

The `SpecializationConstant` class supports multiple constructors, including from Python buffers, a dtype string and a Python buffer (casting to the dtype via NumPy), and a number of bytes and a pointer as integers.

Also introduces `dpctl.program.utils` with `parse_spirv_specializations` utility function, allowing the user to query a SPIR-V directly from Python.


dpctl/program/utils/__init__.py
```diff
@@ -0,0 +1,25 @@
+#                      Data Parallel Control (dpctl)
+#
+# Copyright 2020-2025 Intel Corporation
```
antonwolfy: It seems we need to update the copyright year everywhere in dpctl

dpctl/program/utils/_utils.py
```diff
@@ -0,0 +1,192 @@
...
+    name: str
+    itemsize: int
+    default_value: int | float | bool | None
+
+
+def parse_spirv_specializations(
+    spv_bytes: bytes | bytearray | memoryview,
+) -> tuple[SpecializationConstantInfo]:
+    """
+    Parses SPIR-V byte stream to extract information about specializations,
```
antonwolfy, line 58: Needs to add rendering documentation for the new functions

dpctl/program/utils/_utils.py
```diff
@@ -0,0 +1,192 @@
...
+"""Implements various utilities for the dpctl.program module."""
+
+from dataclasses import dataclass
+from enum import IntEnum
+
+import numpy as np
+
+
+class SpirvOpCode(IntEnum):
+    OpName = 5
```
antonwolfy, line 28: I wonder if it makes sense to import the interesting constants directly from the [SPIRV-Headers](https://github.com/KhronosGroup/SPIRV-Headers)?
Or probably to add just a ref comment with a link to the code from where the constants come?

libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp
```diff
@@ -257,6 +275,22 @@ _CreateKernelBundleWithIL_ocl_impl(const context &ctx,
...
+    if (SpecConsts != nullptr && NumSpecConsts > 0) {
+        auto clSetProgramSpecConstF = get_clSetProgramSpecializationConstant();
+        if (clSetProgramSpecConstF) {
+            for (size_t i = 0; i < NumSpecConsts; ++i) {
+                clSetProgramSpecConstF(clProgram, SpecConsts[i].id,
+                                       SpecConsts[i].size, SpecConsts[i].value);
+            }
+        }
+        else {
+            error_handler("clSetProgramSpecializationConstant is not available "
```
antonwolfy, line 328: Missing to release `clProgram` on error

dpctl/program/utils/_utils.py
```diff
@@ -0,0 +1,192 @@
...
+
+    i = 5  # skip 5 word header
+    while i < len(words):
+        word = words[i]
+        opcode = word & 0xFFFF
+        word_count = word >> 16
+
+        if word_count == 0:
+            raise ValueError(f"Invalid SPIR-V instruction at word index {i}")
+
```
antonwolfy, line 114: Missing boundary check:
```python
if i + word_count > len(words):
    raise ValueError(f"Invalid SPIR-V: instruction at offset {i} extends beyond buffer")
```

dpctl/program/_program.pyx
```diff
@@ -252,6 +269,160 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+            if (
+                isinstance(args[0], numbers.Integral) and
+                isinstance(args[1], numbers.Integral)
+            ):
+                target_obj = PyBytes_FromStringAndSize(
+                    <const char *><size_t>args[1], <Py_ssize_t>args[0]
+                )
+            elif isinstance(args[0], str):
+                target_obj = np.ascontiguousarray(args[1], dtype=args[0])
+
```
antonwolfy, line 366: Do we need an explicit error handling here in case of `else`?

dpctl/program/_program.pyx
```diff
@@ -252,6 +269,160 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+        elif len(args) == 1:
+            target_obj = args[0]
+            if not PyObject_CheckBuffer(target_obj):
+                # attempt to coerce to a numpy array
+                target_obj = np.ascontiguousarray(target_obj)
+        else:
+            raise TypeError(
```
antonwolfy: Probably it'd better to move `raise TypeError(...)` here from line 335, otherwise that branch is unreachable.

libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp
```diff
@@ -444,8 +480,22 @@ _CreateKernelBundleWithIL_ze_impl(const context &SyclCtx,
     ZeDevice = get_native<ze_be>(SyclDev);
 
     // Specialization constants are not supported by DPCTL at the moment
```
antonwolfy: Is the comment obsolete now?

dpctl/program/utils/_utils.py
```diff
@@ -0,0 +1,192 @@
...
+        if spec_id in unique_ids:
+            continue
+        unique_ids.add(spec_id)
+        type_id = constants.get(target_id)
+        type_info = types.get(type_id, {"dtype": "unknown_type", "itemsize": 0})
+        name = names.get(target_id, f"unnamed_spec_const_{spec_id}")
+
+        dtype_str = type_info["dtype"]
+        raw_default = defaults.get(target_id)
+        default_value = None
```
antonwolfy, line 209: It seems when `raw_default` is True or False (from OpSpecConstantTrue/OpSpecConstantFalse at lines 141, 146), the default value is not set in the result. Is that intended?

dpctl/program/_program.pyx
```diff
@@ -252,6 +269,160 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+    The data type specified by the first argument must be a supported data
+    type (bool, integral, or floating point).
+
+    - ``SpecializationConstant(spec_id, nbytes, raw_ptr)``
+      If the constructor is invoked with two variadic arguments where both are
+      integers, the first argument is interpreted as the number of bytes and
+      the second argument is interpreted as a pointer to the data.
+
+    Note that when constructing from a buffer, the
+    :class:`.SpecializationConstant`, shares memory with the original object.
```
antonwolfy: We probably need to warn also about object lifetime: if the source object is deleted, the SpecializationConstant holds a dangling pointer.

dpctl/program/utils/_utils.py
ndgrigorian: there's no wheel package on PyPI for SPIRV-headers, which is why this wasn't done. In theory there is already `spirv-dis` which disassembles SPIRV and can be used to get some of the info, but for that same reason this wasn't done

A link makes sense in the ref though

libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp
ndgrigorian: yep, good catch

dpctl/program/utils/__init__.py
ndgrigorian: I'm inclined to just use 2020 everywhere, as was discussed at some point

dpctl/program/utils/_utils.py
ndgrigorian: good catch

dpctl/program/_program.pyx
ndgrigorian: I decided to resolve this by instead just copying. I don't think there's much benefit or much to gain from permitting modification of the intermediate buffers

dpctl/program/utils/__init__.py
vlad-perevezentsev:
```suggestion
# Copyright 2026 Intel Corporation
```

dpctl/program/utils/_utils.py
```diff
@@ -0,0 +1,201 @@
+#                      Data Parallel Control (dpctl)
+#
+# Copyright 2020-2025 Intel Corporation
```
vlad-perevezentsev:
```suggestion
# Copyright 2026 Intel Corporation
```

dpctl/program/_program.pyx
```diff
@@ -252,6 +269,172 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+        return f"SpecializationConstant({self._spec_const.id})"
+
+    def __eq__(self, other):
+        if not isinstance(other, SpecializationConstant):
+            return False
+        cdef SpecializationConstant _other = <SpecializationConstant>other
+        if (
+            self._spec_const.id != _other._spec_const.id or
+            self._spec_const.size != _other._spec_const.size or
+            self._spec_const.value != _other._spec_const.value
```
vlad-perevezentsev: Since `memcmp()` operates on buffer pointers, `.value` appears to be a pointer so this check compares addresses rather than contents and returns `True` then __eq__ returns `False` for two distinct objects with identical values
```
In [1]: import dpctl.program as dpctl_prog

In [2]: a = dpctl_prog.SpecializationConstant(0, "i4", 42)

In [3]: b = dpctl_prog.SpecializationConstant(0, "i4", 42)

In [4]: a == b
Out[4]: False
```

libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp
```diff
@@ -257,6 +293,28 @@ _CreateKernelBundleWithIL_ocl_impl(const context &ctx,
         return nullptr;
     }
 
+    if (SpecConsts != nullptr && NumSpecConsts > 0) {
+        auto clSetProgramSpecConstF = get_clSetProgramSpecializationConstant();
+        if (clSetProgramSpecConstF) {
+            for (size_t i = 0; i < NumSpecConsts; ++i) {
+                clSetProgramSpecConstF(clProgram, SpecConsts[i].id,
```
vlad-perevezentsev: What happens if `clSetProgramSpecConstF` returns an error other than `CL_SUCCESS`?

docs/doc_sources/api_reference/dpctl/program.rst
```diff
@@ -37,3 +37,15 @@ execution via :py:meth:`dpctl.SyclQueue.submit`.
...
+:py:mod:`dpctl.program.utils`
+-----------------------------
+
+.. py:module:: dpctl.program.utils
+
+.. currentmodule:: dpctl.program.utils
+
+.. autofunction:: parse_spirv_specializations
+
+.. autoclass:: SpecializationConstantInfo
```
vlad-perevezentsev, line 51: I don not see this in the generated documentation
https://intelpython.github.io/dpctl/pulls/2304/api_reference/dpctl/program.html#module-dpctl.program.utils

vlad-perevezentsev: It looks like this may be missing in utils/__init__.py

docs/doc_sources/api_reference/dpctl/program.rst
```diff
@@ -37,3 +37,15 @@ execution via :py:meth:`dpctl.SyclQueue.submit`.
     :nosignatures:
 
     SyclKernelBundleCompilationError
```
vlad-perevezentsev, line 40: `SpecializationConstant` is missing

dpctl/program/_program.pyx
```diff
@@ -252,6 +269,172 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+      or is one, it must have a supported data type (bool, integral, or
+      floating point). The specialization constant will be constructed from the
+      data in the buffer
+
+    - ``SpecializationConstant(spec_id, dtype, obj)``
+    If the constructor is invoked with two variadic arguments, and the first
+    argument is a string, it is interpreted as a NumPy ``dtype`` string and the
+    second argument will be coerced to a NumPy array with that data type.
+    The data type specified by the first argument must be a supported data
+    type (bool, integral, or floating point).
```
vlad-perevezentsev: The indents are incorrect

dpctl/program/_program.pyx
```diff
@@ -252,6 +269,172 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+
+        if not isinstance(spec_id, numbers.Integral):
+            raise TypeError(
+                "Specialization constant ID must be of type `int`, got "
+                f"{type(spec_id)}"
+            )
+
+        if len(args) == 0 or len(args) > 2:
+            raise TypeError(
+                f"Constructor takes 2 or 3 arguments, got {len(args)}."
```
vlad-perevezentsev: The text of the error is not clear
`len(args)` returns 0 
```
dpctl_prog.SpecializationConstant(0)
# TypeError: Constructor takes 2 or 3 arguments, got 0.
```

dpctl/program/_program.pyx
```diff
@@ -252,6 +269,172 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+        return self._spec_const.id
+
+    @property
+    def size(self):
+        """
+        Returns the size in bytes of the data for this specialization constant.
+        """
+        return self._spec_const.size
+
+    cdef size_t addressof(self):
```
vlad-perevezentsev, line 459: What is the reason for adding this?
It is not used

ndgrigorian: initially, I was considering if we needed to permit SpecConst kernel args, but it's a good point, we don't need them for that, it's only for JIT, so it can be removed

dpctl/program/_program.pyx
ndgrigorian: fixed this

libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp
ndgrigorian: I've added a check for this

dpctl/program/_program.pyx
ndgrigorian: I thought about it and instead have added a test using it similar to how it's used for Kernel and KernelBundle, and addressof_ref from Python

dpctl/program/_program.pyx
ndgrigorian: good point, I've added 1 to it (for the positional argument)

dpctl/program/_program.pyx
```diff
@@ -337,11 +538,44 @@ cpdef create_kernel_bundle_from_spirv(
...
+        num_spconsts = len(specializations)
+        spconsts = <_spec_const *>(
+            malloc(num_spconsts * sizeof(_spec_const))
+        )
+        if spconsts == NULL:
+            raise MemoryError(
+                "Failed to allocate memory for specialization constants."
+            )
+        for i, spconst in enumerate(specializations):
+            if not isinstance(spconst, SpecializationConstant):
```
antonwolfy: The loop variable is declared `cdef SpecializationConstant spconst`, so for `i, spconst in enumerate(specializations)` performs the type coercion at the loop assignment.
A non-SpecializationConstant element raises Cython's own `TypeError` before reaching the `isinstance` guard at line 555 — so the custom error message and the `free(spconsts)` cleanup never run, leaking the `spconsts` allocation.

dpctl/program/__init__.py
```diff
@@ -41,6 +42,12 @@
     "SyclKernelBundleCompilationError",
     "SyclProgram",
     "SyclProgramCompilationError",
+    "SpecializationConstant",
+]
+
+# add submodules
+__all__ += [
+    "utils",
```
antonwolfy, line 51: `"utils"` in `__all__` is never imported

antonwolfy: Tests only pass because they do from `dpctl.program.utils import ...` explicitly. Needs to add `from . import utils`

dpctl/program/utils/_utils.py
```diff
@@ -0,0 +1,201 @@
...
+
+    i = 5  # skip 5 word header
+    while i < len(words):
+        word = words[i]
+        opcode = word & 0xFFFF
+        word_count = word >> 16
+
+        if word_count == 0:
+            raise ValueError(f"Invalid SPIR-V instruction at word index {i}")
+        if i + word_count > len(words):
```
antonwolfy, line 109: SPIR-V parser raises uncaught `IndexError` / `UnicodeDecodeError` on malformed input.

The only bounds check is that the declared `word_count` fits the buffer; per-opcode operand reads (`OpTypeInt` needs `i+3`, `OpDecorate` needs `i+3`, `OpName` decodes UTF-8) are unchecked. A truncated instruction throws raw `IndexError`, and invalid UTF-8 in `OpName` throws `UnicodeDecodeError` — both contradicting the documented `ValueError("Invalid SPIR-V binary")` contract for a function whose explicit job is parsing untrusted binaries. Neither path is tested.

libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp
```diff
@@ -443,9 +520,22 @@ _CreateKernelBundleWithIL_ze_impl(const context &SyclCtx,
...
+        spec_ids.reserve(NumSpecConsts);
+        spec_values.reserve(NumSpecConsts);
+        for (size_t i = 0; i < NumSpecConsts; ++i) {
+            spec_ids.push_back(SpecConsts[i].id);
+            spec_values.push_back(SpecConsts[i].value);
+        }
+    }
     ze_module_constants_t ZeSpecConstants = {};
-    ZeSpecConstants.numConstants = 0;
+    ZeSpecConstants.numConstants = static_cast<std::uint32_t>(NumSpecConsts);
```
antonwolfy: `numConstants`  set even when arrays are empty.

`ZeSpecConstants.numConstants` is set to `NumSpecConsts` unconditionally, while `pConstantIds`/`pConstantValues` become `nullptr` when the vectors are empty. If a caller passes `NumSpecConsts > 0` with `SpecConsts == nullptr`, L0 gets a nonzero count with null arrays → UB. The OCL path silently ignores the same input, so the two backends diverge. (While the Cython layer never produces this, so it's a hardening/API-consistency issue.)

dpctl/program/_program.pyx
```diff
@@ -253,6 +270,185 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+                f"{type(spec_id)}"
+            )
+
+        if len(args) == 0 or len(args) > 2:
+            raise TypeError(
+                "Constructor takes 2 or 3 arguments, got "
+                f"{len(args) + 1}."
+            )
+
+        self._spec_const.id = <uint32_t>spec_id
```
antonwolfy, line 345: It wraps silently for out-of-range/negative IDs. SPIR-V `SpecId` is a 32-bit decoration; an out-of-range value should raise, not wrap.

dpctl/program/_program.pyx
```diff
@@ -253,6 +270,185 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+        ret_code = PyObject_GetBuffer(
+            target_obj, &(_local_buffer), PyBUF_SIMPLE | PyBUF_ANY_CONTIGUOUS
+        )
+        if ret_code != 0:
+            raise ValueError(
+                "Failed to get buffer view for the provided object."
+            )
+
+        self._spec_const.size = <size_t>_local_buffer.len
+        copied_data = malloc(self._spec_const.size)
```
antonwolfy, line 400: What if it is zero-sized constant, but we will raise `MemoryError` then below.

dpctl/program/_program.pyx
```diff
@@ -253,6 +270,185 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+        PyBuffer_Release(&(_local_buffer))
+
+    def __dealloc__(self):
+        if self._spec_const.value != NULL:
+            free(<void*>self._spec_const.value)
+
+    def __repr__(self):
+        return f"SpecializationConstant({self._spec_const.id})"
+
+    def __eq__(self, other):
```
antonwolfy, line 420: Do we need to add definition for `__hash__`, since defining `__eq__` might drop hashability?

dpctl/program/_program.pyx
```diff
@@ -253,6 +270,185 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+        if not isinstance(other, SpecializationConstant):
+            return False
```
antonwolfy: To unblock reflected comparison:
```suggestion
            return NotImplemented
```

dpctl/program/_program.pyx
```diff
@@ -253,6 +270,185 @@ cdef api SyclKernelBundle SyclKernelBundle_Make(DPCTLSyclKernelBundleRef KBRef):
...
+            )
+
+        self._spec_const.id = <uint32_t>spec_id
+
+        if len(args) == 2:
+            if (
+                isinstance(args[0], numbers.Integral) and
+                isinstance(args[1], numbers.Integral)
+            ):
+                target_obj = PyBytes_FromStringAndSize(
```
antonwolfy, line 357: No `nbytes >= 0` check.

A negative `Py_ssize_t` into `PyBytes_FromStringAndSize` is undefined; the unsafe-pointer path is inherently risky but should at least range-check.

libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp
```diff
@@ -611,12 +703,14 @@ DPCTLKernelBundle_CreateFromSpirv(__dpctl_keep const DPCTLSyclContextRef CtxRef,
     switch (BE) {
     case backend::opencl:
         KBRef = _CreateKernelBundleWithIL_ocl_impl(*SyclCtx, *SyclDev, IL,
-                                                   length, CompileOpts);
+                                                   length, CompileOpts,
```
antonwolfy: No try/catch across the C ABI boundary in the OpenCL/top-level path.
The spec-constant build failures enlarge the throwing input space.

dpctl/program/utils/_utils.py
antonwolfy: Should that be covered in the tests?

libsyclinterface/include/syclinterface/dpctl_sycl_kernel_bundle_interface.h
```diff
@@ -61,7 +70,9 @@ DPCTLKernelBundle_CreateFromSpirv(__dpctl_keep const DPCTLSyclContextRef Ctx,
                                   __dpctl_keep const DPCTLSyclDeviceRef Dev,
                                   __dpctl_keep const void *IL,
                                   size_t Length,
-                                  const char *CompileOpts);
+                                  const char *CompileOpts,
+                                  size_t NumSpecConsts,
+                                  const DPCTLSpecConst *SpecConsts);
```
antonwolfy: Do we need to add `__dpctl_keep`?

dpctl/program/utils/_utils.py
```diff
@@ -0,0 +1,230 @@
...
+            if word_count < 3:
+                raise ValueError(f"Truncated OpTypeFloat at word index {i}")
+            result_id = int(words[i + 1])
+            width = int(words[i + 2])
+            types[result_id] = {
+                "dtype": f"f{width // 8}",
+                "itemsize": width // 8,
+            }
+        elif opcode == SpirvOpCode.OpSpecConstant:
+            if word_count < 3:
```
antonwolfy: Seems a typo:
```suggestion
            if word_count < 4:
```

dpctl/program/_program.pyx
```diff
@@ -337,11 +564,45 @@ cpdef create_kernel_bundle_from_spirv(
...
+    cdef size_t num_spconsts
+    cdef _spec_const *spconsts
+    cdef SpecializationConstant spconst
+
+    if specializations is not None:
+        num_spconsts = len(specializations)
+        spconsts = <_spec_const *>(
+            malloc(num_spconsts * sizeof(_spec_const))
+        )
+        if spconsts == NULL:
```
antonwolfy, line 576: The check will be True when passed `specializations=[]` and so it will cause raising of MemoryError

libsyclinterface/source/dpctl_sycl_kernel_bundle_interface.cpp
```diff
@@ -257,6 +293,51 @@ _CreateKernelBundleWithIL_ocl_impl(const context &ctx,
         return nullptr;
     }
 
+    if (NumSpecConsts > 0 && SpecConsts == nullptr) {
+        error_handler("NumSpecConsts > 0 but SpecConsts is NULL.", __FILE__,
+                      __func__, __LINE__);
+        return nullptr;
```
antonwolfy: Do we need to call `get_clReleaseProgram` here on negative exit? Or probably to move the check o input arguments before `get_clCreateProgramWithIL()` call,

antonwolfy (approved):
Thank @ndgrigorian , LGTM

PR #2306, opened 2026-04-29
Avoid warnings from Unified Runtime Level Zero adapter

ndgrigorian:
This PR proposes adding a backend check in the C API for `SyclPlatform.get_devices`, to avoid a warning from the unified runtime

The warning is raised when `device_type::custom` is requested. The UR LZ adapter has no concept of custom, causing it to warn that the device type is unknown

### DIFF
--- a/libsyclinterface/source/dpctl_sycl_platform_interface.cpp
+++ b/libsyclinterface/source/dpctl_sycl_platform_interface.cpp
@@ -304,6 +304,15 @@ DPCTLPlatform_GetDevices(__dpctl_keep const DPCTLSyclPlatformRef PRef,
         return wrap<vecTy>(DevicesVectorPtr);
     }
 
+    DPCTLSyclBackendType BTy =
+        DPCTL_SyclBackendToDPCTLBackendType(P->get_backend());
+    if (DTy == DPCTLSyclDeviceType::DPCTL_CUSTOM &&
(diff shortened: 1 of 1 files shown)


libsyclinterface/source/dpctl_sycl_platform_interface.cpp
```diff
@@ -304,6 +304,15 @@ DPCTLPlatform_GetDevices(__dpctl_keep const DPCTLSyclPlatformRef PRef,
         return wrap<vecTy>(DevicesVectorPtr);
     }
 
+    DPCTLSyclBackendType BTy =
+        DPCTL_SyclBackendToDPCTLBackendType(P->get_backend());
+    if (DTy == DPCTLSyclDeviceType::DPCTL_CUSTOM &&
+        BTy == DPCTLSyclBackendType::DPCTL_LEVEL_ZERO)
+    {
+        // avoid ugly warnings from unified runtime
+        return wrap<vecTy>(DevicesVectorPtr);
```
antonwolfy, line 313: Any way to cover that in the test?

ndgrigorian: it isn't obvious how to capture the UR warnings. It might be possible by some gymnastics capturing all stdout, but, then it only applies for Level Zero. I can test a few things first though

vlad-perevezentsev (approved):
LGTM
Thank you @ndgrigorian

PR #2312, opened 2026-05-07
Update allowed docs versions and add backfill workflow

ndgrigorian:
This PR updates the allowed versions of dpctl docs in the dropdown, and adds a workflow to backfill old documentation versions.

This workflow is designed to allow passing a tag to the workflow and manually executing, creating and populating the folder for the tag


.github/workflows/backfill-docs.yml
```diff
@@ -0,0 +1,124 @@
...
+              -DCMAKE_VERBOSE_MAKEFILE=ON
+          python -m pip install -e . --no-build-isolation --no-deps
+          python -c "import dpctl; print(dpctl.__version__)" || exit 1
+          pushd "$(find _skbuild -name cmake-build)" || exit 1
+          cmake --build . --target Sphinx || exit 1
+          mv ../cmake-install/docs/docs ~/docs
+          git clean -dfx
+          popd
+          git reset --hard
+      - name: Publish docs
```
antonwolfy, line 103: We use `peaceiris/actions-gh-pages` action in dpnp for that (`.github/workflows/build-sphinx.yml` in case you will find that helpful.

.github/workflows/backfill-docs.yml
```diff
@@ -0,0 +1,124 @@
...
+env:
+  CHECK_CONFIG_SCRIPT: "import sys; from packaging.version import parse; print('true' if parse('0.17.0') <= parse(sys.argv[1]) < parse('0.22.0') else 'false')"
+
+jobs:
+  build-and-backfill:
+    name: Build and Backfill Documentation
+    runs-on: ubuntu-latest
+    timeout-minutes: 240
+    permissions:
+      contents: write
```
antonwolfy, line 21: Do we need `actions: write` to cancel any previous runs that are not completed for a given workflow?

.github/workflows/backfill-docs.yml
```diff
@@ -0,0 +1,124 @@
...
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@d07a454dad7609a92316b57b23c9ccfd4f59af66 # 0.13.1
+        with:
+          access_token: ${{ github.token }}
+      - name: Add Intel repository
+        if: ${{ !github.event.pull_request || github.event.action != 'closed' }}
```
antonwolfy: Why do we need to check if it's triggered for the PR if `workflow_dispatch` means only manual trigger, isn't it?

ndgrigorian: good point

.github/workflows/backfill-docs.yml
ndgrigorian: I'll take a look, may be good to reuse in the other docs workflow too

ndgrigorian: I'll attempt to apply this to both in a follow up

.github/workflows/backfill-docs.yml
```diff
@@ -0,0 +1,119 @@
...
+        run: |
+          sudo apt-get install liblua5.2-dev
+      - name: Install Doxygen
+        run: |
+          sudo apt-get install doxygen
+      - name: Install Ninja
+        run: |
+          sudo apt-get install ninja-build
+      - name: Setup Python
+        if: ${{ !github.event.pull_request || github.event.action != 'closed' }}
```
antonwolfy: Seems missed to remove

ndgrigorian: good catch, fixed it

antonwolfy (approved):
No more comments from me

PR #2322, opened 2026-06-08
Return tuple instead of list from device and platform collection APIs

vlad-perevezentsev:
Initially the goal was to change `dpctl.get_devices()` to return a tuple instead of a list to align with the Python Array API specification where `info.devices` is defined as a tuple

While implementing this change it became clear that the same behavior should be applied consistently across other APIs that return collections of devices or platforms.
This PR also updates `get_platforms, get_composite_devices, SyclContext.get_devices, SyclPlatform.get_devices, SyclPlatform.get_composite_devices, SyclDevice.create_sub_devices, SyclDevice.component_devices, and SyclDevice.sub_group_sizes` to return tuples instead of lists 

### DIFF
--- a/dpctl/_sycl_platform.pyx
+++ b/dpctl/_sycl_platform.pyx
@@ -391,8 +391,8 @@ cdef class SyclPlatform(_SyclPlatform):
                 Default: ``dpctl.device_type.all``.
 
         Returns:
-            tuple:
-                A :obj:`tuple` of :class:`dpctl.SyclDevice` objects
+            Tuple[:class:`dpctl.SyclDevice`]:
+                A tuple of :class:`dpctl.SyclDevice` objects
                 that belong to this platform.
 
         Raises:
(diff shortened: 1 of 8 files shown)


dpctl/_sycl_platform.pyx
```diff
@@ -596,13 +596,13 @@ def lsplatform(verbosity=0):
...
+cpdef tuple get_platforms():
     """
-    Returns a list of all available SYCL platforms on the system.
+    Returns a tuple of all available SYCL platforms on the system.
 
     Returns:
-        List[:class:`.SyclPlatform`]:
-            A list of SYCL platforms on the system.
+        Tuple[:class:`.SyclPlatform`]:
+            A tuple of SYCL platforms on the system.
```
ndgrigorian, line 605: while we're at it, can we adjust the other docstrings to use the `Tuple[:class:]` return format? It's a bit more readable

ndgrigorian (dismissed):
other than that small adjustment lgtm

dpctl/_sycl_platform.pyx
vlad-perevezentsev: That is a good point
I have updated it
thanks

ndgrigorian (approved):
LGTM, thank you @vlad-perevezentsev

PR #2324, opened 2026-06-12
Add missing device info queries

ndgrigorian:
This PR adds the missing SYCL 2020 spec (i.e., non-extension) device info queries to `dpctl.SyclDevice`

- [X] Have you added your changes to the changelog?


dpctl/__init__.py
```diff
@@ -118,6 +123,11 @@
     "backend_type",
     "event_status_type",
     "global_mem_cache_type",
+    "local_mem_type",
```
antonwolfy, line 126: Missed to update `docs/doc_sources/api_reference/dpctl/constants.rst`

dpctl/_sycl_device.pyx
```diff
@@ -2110,6 +2144,446 @@ cdef class SyclDevice(_SyclDevice):
...
+        cdef size_t i
+        cdef list res
+
+        arr = DPCTLDevice_GetPartitionProperties(
+            self._device_ref, &arr_len
+        )
+        if arr is not NULL and arr_len > 0:
+            res = []
+            for i in range(arr_len):
+                res.append(partition_property(arr[i] + 1))
```
antonwolfy: Do we need to guard against possible `-1` value which `dpctl_partition_property_to_int` can return?

libsyclinterface/tests/test_sycl_device_interface.cpp
```diff
@@ -551,6 +551,136 @@ TEST_P(TestDPCTLSyclDeviceInterface, ChkGetCompositeDevice)
...
+
+TEST_P(TestDPCTLSyclDeviceInterface, ChkGetSingleFPConfig)
+{
+    int *arr = nullptr;
+    size_t len = 0;
+    EXPECT_NO_FATAL_FAILURE(arr = DPCTLDevice_GetSingleFPConfig(DRef, &len));
+    EXPECT_TRUE(len > 0);
+    EXPECT_TRUE(arr != nullptr);
+    if (arr)
+        delete[] arr;
```
antonwolfy: Do we need to use `DPCTLInt_Array_Delete(arr)` instead?

libsyclinterface/include/syclinterface/dpctl_sycl_device_interface.h
```diff
@@ -828,4 +828,271 @@ DPCTL_API
...
+DPCTLPartitionAffinityDomainType DPCTLDevice_GetPartitionTypeAffinityDomain(
+    __dpctl_keep const DPCTLSyclDeviceRef DRef);
+
+/*!
+ * @brief Wrapper over device.get_info<info::device::half_fp_config>().
+ *
+ * @param    DRef           Opaque pointer to a ``sycl::device``
+ * @param    res_len        Populated with size of the returned array
+ * @return   Returns an array of DPCTLFPConfigType values. Caller must free
+ * with DPCTLSize_t_Array_Delete.
```
antonwolfy:
```suggestion
 * with DPCTLInt_Array_Delete.
```

antonwolfy: We have a memory release instruction here, but not for DPCTLDevice_GetSingleFPConfig. Is that intentional?
And probably this is applicable to some other docstrings as well.

libsyclinterface/tests/test_sycl_device_interface.cpp
```diff
@@ -551,6 +551,136 @@ TEST_P(TestDPCTLSyclDeviceInterface, ChkGetCompositeDevice)
...
+                ptp == DPCTL_PARTITION_EQUALLY ||
+                ptp == DPCTL_PARTITION_BY_COUNTS ||
+                ptp == DPCTL_PARTITION_BY_AFFINITY_DOMAIN);
+}
+
+TEST_P(TestDPCTLSyclDeviceInterface, ChkGetSingleFPConfig)
+{
+    int *arr = nullptr;
+    size_t len = 0;
+    EXPECT_NO_FATAL_FAILURE(arr = DPCTLDevice_GetSingleFPConfig(DRef, &len));
```
antonwolfy, line 664: There are 7 new functions which are not testing:
- GetImageMaxBufferSize
- GetHalfFPConfig
- GetDoubleFPConfig
- GetAtomicFenceOrderCapabilities
- GetAtomicFenceScopeCapabilities
- GetPartitionAffinityDomains
- GetPartitionTypeAffinityDomain

antonwolfy: Also none of 22 new API has null-DRe tests, like above ChkCopy

dpctl/_sycl_device.pyx
ndgrigorian: added exceptions, which is what we do in the other enums

libsyclinterface/tests/test_sycl_device_interface.cpp
ndgrigorian: yes, fixed

libsyclinterface/include/syclinterface/dpctl_sycl_device_interface.h
ndgrigorian: I'm going to remove the deletion instruction for consistency, as it is never used elsewhere

libsyclinterface/tests/test_sycl_device_interface.cpp
ndgrigorian: added tests

libsyclinterface/include/syclinterface/dpctl_sycl_device_interface.h
```diff
@@ -828,4 +828,270 @@ DPCTL_API
...
+/*!
+ * @brief Wrapper over device.get_info<info::device::half_fp_config>().
+ *
+ * @param    DRef           Opaque pointer to a ``sycl::device``
+ * @param    res_len        Populated with size of the returned array
+ * @return   Returns an array of DPCTLFPConfigType values.
+ * @ingroup DeviceInterface
+ */
+DPCTL_API
+__dpctl_keep int *
```
antonwolfy: Should it be `__dpctl_give` on the array-returning functions?

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -982,3 +982,417 @@ void DPCTLDevice_DisablePeerAccess(__dpctl_keep const DPCTLSyclDeviceRef DRef,
...
+            error_handler(e, __FILE__, __func__, __LINE__);
+        }
+    }
+    return cstr_version;
+}
+
+DPCTLLocalMemType
+DPCTLDevice_GetLocalMemType(__dpctl_keep const DPCTLSyclDeviceRef DRef)
+{
+    if (DRef) {
```
antonwolfy, line 1137: In `DPCTLDevice_GetGlobalMemCacheType` we use `error_handler("Argument DRef is null", __FILE__, __func__, __LINE__);` on else-branch.
What is the common approach?

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -982,3 +982,417 @@ void DPCTLDevice_DisablePeerAccess(__dpctl_keep const DPCTLSyclDeviceRef DRef,
...
+        try {
+            auto mem_type = D->get_info<info::device::local_mem_type>();
+            switch (mem_type) {
+            case info::local_mem_type::none:
+                return DPCTL_LOCAL_MEM_TYPE_NONE;
+            case info::local_mem_type::local:
+                return DPCTL_LOCAL_MEM_TYPE_LOCAL;
+            case info::local_mem_type::global:
+                return DPCTL_LOCAL_MEM_TYPE_GLOBAL;
+            }
```
antonwolfy: In DPCTLDevice_GetGlobalMemCacheType in case of default we use:
```cpp
        assert(false);
        return DPCTL_MEM_CACHE_TYPE_INDETERMINATE;
```
Should we do something similar here?

libsyclinterface/tests/test_sycl_device_interface.cpp
```diff
@@ -551,6 +561,242 @@ TEST_P(TestDPCTLSyclDeviceInterface, ChkGetCompositeDevice)
...
+    uint32_t ab = 0;
+    EXPECT_NO_FATAL_FAILURE(ab = DPCTLDevice_GetAddressBits(DRef));
+    EXPECT_TRUE(ab == 32 || ab == 64);
+}
+
+TEST_P(TestDPCTLSyclDeviceInterface, ChkGetMaxSamplers)
+{
+    uint32_t ms = 0;
+    EXPECT_NO_FATAL_FAILURE(ms = DPCTLDevice_GetMaxSamplers(DRef));
+    EXPECT_TRUE(ms >= 0);
```
antonwolfy: Should we expect returning 0 as a negative test result?
```suggestion
    EXPECT_TRUE(ms > 0);
```

dpctl/tests/_device_attributes_checks.py
```diff
@@ -659,6 +659,146 @@ def check_global_mem_cache_line_size(device):
...
+def check_image_max_buffer_size(device):
+    sz = device.image_max_buffer_size
+    assert isinstance(sz, int)
+    assert sz >= 0
+
+
+def check_max_samplers(device):
+    ms = device.max_samplers
+    assert isinstance(ms, int)
+    assert ms >= 0
```
antonwolfy, line 683: Same question here, if `ms == 0` is a valid result

libsyclinterface/tests/test_sycl_device_interface.cpp
antonwolfy: The point is that `ms >= 0` is always true for unsigned integer.

dpctl/_sycl_device.pyx
```diff
@@ -2123,6 +2179,409 @@ cdef class SyclDevice(_SyclDevice):
...
+            bool:
+                Whether the device is currently available.
+        """
+        return DPCTLDevice_IsAvailable(self._device_ref)
+
+    @property
+    def version(self):
+        """ Returns a backend-defined device version string.
+
+        Returns:
```
antonwolfy, line 2278: Missing `Raises:` clause

dpctl/_sycl_device.pyx
```diff
@@ -2123,6 +2179,409 @@ cdef class SyclDevice(_SyclDevice):
...
+            raise RuntimeError("Descriptor 'version' not available")
+        cdef str ver_str = ver.decode()
+        DPCTLCString_Delete(ver)
+        return ver_str
+
+    @property
+    def backend_version(self):
+        """ Returns a backend-defined driver version string.
+
+        Returns:
```
antonwolfy, line 2297: Missing `Raises:` clause

dpctl/_sycl_device.pyx
```diff
@@ -2123,6 +2179,409 @@ cdef class SyclDevice(_SyclDevice):
...
+        cdef const char *ver = DPCTLDevice_GetVersion(self._device_ref)
+        if ver is NULL:
+            raise RuntimeError("Descriptor 'version' not available")
+        cdef str ver_str = ver.decode()
+        DPCTLCString_Delete(ver)
+        return ver_str
+
+    @property
+    def backend_version(self):
+        """ Returns a backend-defined driver version string.
```
antonwolfy: That duplicates docstring of `driver_version` property and misaligned with `DPCTLDevice_GetBackendVersion` which states:
> @return   Returns a C string with the backend version.

libsyclinterface/include/syclinterface/dpctl_sycl_device_interface.h
ndgrigorian: fixed and fixed it in a few other places

libsyclinterface/source/dpctl_sycl_device_interface.cpp
ndgrigorian: I've cleaned it up in `DPCTLDevice_GetGlobalMemCacheType`, which was the odd one out

libsyclinterface/source/dpctl_sycl_device_interface.cpp
ndgrigorian: I removed the assert and instead we just fall through by default. The assert wasn't doing anything outside debug builds.

dpctl/tests/_device_attributes_checks.py
ndgrigorian: It is, I've removed the check and have done so in other places as well, it's a good point

ndgrigorian: Oops, meant for above, left the check on Python test

libsyclinterface/tests/test_sycl_device_interface.cpp
ndgrigorian: It is, I've removed the check and have done so in other places as well, it's a good point

dpctl/tests/_device_attributes_checks.py
ndgrigorian: And yes `ms == 0` is valid

libsyclinterface/source/dpctl_sycl_device_interface.cpp
```diff
@@ -982,3 +982,417 @@ void DPCTLDevice_DisablePeerAccess(__dpctl_keep const DPCTLSyclDeviceRef DRef,
...
+        return DPCTL_PARTITION_BY_COUNTS;
+    case info::partition_property::partition_by_affinity_domain:
+        return DPCTL_PARTITION_BY_AFFINITY_DOMAIN;
+    default:
+        // TODO: investigate ext_intel_partition_by_cslice extension
+        return -1;
+    }
+}
+
+int dpctl_partition_affinity_domain_to_int(info::partition_affinity_domain pad)
```
antonwolfy: Seems fully duplicated with `DPCTL_SyclPartitionAffinityDomainToDPCTLType`, or intentional?

antonwolfy (dismissed):
LGTM with only small question below

libsyclinterface/source/dpctl_sycl_device_interface.cpp
ndgrigorian: I agree, after reevaluating, I decided that it's more consistent to instead add values for unknown/invalid enum entries and have them be part of the array, rather than skipping them, which removed need for these duplicate helpers.

libsyclinterface/include/syclinterface/dpctl_sycl_enum_types.h
```diff
@@ -143,6 +143,7 @@ typedef enum
  */
 typedef enum
 {
+    DPCTL_PARTITION_AFFINITY_DOMAIN_UNKNOWN = 0,
```
antonwolfy, line 146: Is not that a is a numeric-ABI change to already available API? We should at least state that explicitly.

ndgrigorian: it is though this PR is not yet part of any release, and would only impact bleeding-edge builds

where do you feel it should be stated, changelog?

ndgrigorian: added it to the changelog

antonwolfy (approved):

PR #2331, opened 2026-06-30
Add Python bindings for SYCL IPC memory via dpctl

zxue2:
Add inter-process communication (IPC) support for SYCL USM memory, enabling zero-copy GPU memory sharing across processes. It wraps sycl::ext::oneapi::experimental::ipc::memory (get/open/close).

C API (libsyclinterface):
  - dpctl_sycl_ipc_memory_interface.h: declares DPCTLIPCMem_GetHandle, DPCTLIPCMem_OpenHandle, DPCTLIPCMem_CloseHandle, DPCTLIPCMem_FreeHandleData
  - dpctl_sycl_ipc_memory_interface.cpp: implements the C API by calling the SYCL experimental ipc::memory functions; auto-discovered by CMake.

Cython declarations:
  - _backend.pxd: extern block for the 3 new C functions

Python subpackage (dpctl.ipc):
  - IPCMemoryHandle: Cython extension class wrapping IPC memory export/import __init__(usm_memory): calls get() + put(), stores handle as bytes to_bytes(): serializable payload for cross-process transport open(handle_bytes, device, nbytes): returns MemoryUSMDevice close_mapping(usm_memory): explicitly closes an IPC mapping

**Test In Docker (oneAPI 2026.0 installed):**
```
python -c "
import dpctl
from dpctl.memory import MemoryUSMDevice
from dpctl.ipc import IPCMemoryHandle
d = dpctl.SyclDevice('level_zero:gpu')
q = dpctl.SyclQueue(d)
mem = MemoryUSMDevice(4096, queue=q)
h = IPCMemoryHandle(mem)
raw = h.to_bytes()
print(f'IPC handle: {len(raw)} bytes')
mem2 = IPCMemoryHandle.open(raw, d, nbytes=4096)
print(f'Opened: {mem2.nbytes} bytes on {mem2.sycl_device.name}')
IPCMemoryHandle.close_mapping(mem2)
print('close OK')
"

IPC handle: 120 bytes
Opened: 4096 bytes on Intel(R) Arc(TM) Pro B60 Graphics
close OK
```


dpctl/ipc/_ipc_memory.pyx
```diff
@@ -0,0 +1,253 @@
...
+import dpctl
+
+
+__all__ = ["IPCMemoryHandle"]
+
+
+cdef class IPCMemoryHandle:
+    """Wrapper around a SYCL IPC memory handle.
+
+    Instances are created by passing a :class:`dpctl.memory.MemoryUSMDevice`
```
ndgrigorian: if this docstring is accurate and the extension only works with `MemoryUSMDevice`, then shouldn't we add `isinstance` check below to guarantee it's device memory?

Is it a requirement that it be a USM device allocation, or can host and/or shared work?

dpctl/ipc/_ipc_memory.pyx
```diff
@@ -0,0 +1,253 @@
...
+        cdef DPCTLSyclContextRef ctx_ref = context.get_context_ref()
+        cdef char *data_out = NULL
+        cdef size_t size_out = 0
+
+        cdef int rc = DPCTLIPCMem_GetHandle(ptr, ctx_ref, &data_out, &size_out)
+        if rc != 0:
+            raise RuntimeError(
+                "DPCTLIPCMem_GetHandle failed — device may not support "
+                "aspect::ext_oneapi_ipc_memory"
+            )
```
ndgrigorian: can we also add this aspect to `dpctl.SyclDevice` as part of this PR? In general to users, it would be helpful to be able to check, and then we can perform a check in here as well

dpctl/ipc/_ipc_memory.pyx
```diff
@@ -0,0 +1,253 @@
...
+
+        cdef DPCTLSyclContextRef ctx_ref = context.get_context_ref()
+        DPCTLIPCMem_CloseHandle(ptr, ctx_ref)
+
+        # Prevent the _Memory destructor from calling sycl::free on the
+        # now-unmapped pointer. Setting _opaque_ptr to NULL makes
+        # __dealloc__ skip OpaqueSmartPtr_Delete.
+        usm_memory._opaque_ptr = NULL
+        usm_memory._memory_ptr = NULL
+        usm_memory.nbytes = 0
```
ndgrigorian: this seems a bit dangerous for users

I think it would make the most sense to instead, since `_opaque_ptr` is `std::shared_ptr`, share the `shared_ptr` with the memory object shared pointer if possible

dpctl/ipc/_ipc_memory.pyx
```diff
@@ -0,0 +1,253 @@
...
+            If the handle cannot be opened.
+        """
+        cdef const char *raw = PyBytes_AS_STRING(handle_bytes)
+        cdef size_t raw_size = <size_t>len(handle_bytes)
+
+        if context is None:
+            context = device.sycl_platform.default_context
+
+        cdef DPCTLSyclContextRef ctx_ref = context.get_context_ref()
+        cdef DPCTLSyclDeviceRef dev_ref = device.get_device_ref()
```
ndgrigorian: if we add the aspect check, we can add a check here for the device

dpctl/ipc/_ipc_memory.pxd
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# distutils: language = c++
+# cython: language_level=3
+
+"""Declarations for the IPCMemoryHandle Cython extension type."""
```
ndgrigorian: I'm conflicted on whether this belongs in a separate namespace or not. So far, we've been free with incorporating extensions into the main dpctl namespace, as long as we have fall-backs so that they don't crash, etc. when the extension is unavailable. This will enable us to use AdaptiveCpp in the future.

Having an entire submodule not be built is a bit different. Perhaps instead we can move this into `dpctl.memory` and implement fallbacks in the C-API and an `is_available` free function like we have for `WorkGroupMemory`, etc. to check for it

dpctl/ipc/_ipc_memory.pyx
zxue2: urIPCGetMemHandleExp without USM type restriction but urIPCOpenMemHandleExp is hardcoded to UR_USM_TYPE_DEVICE.  
So urIPCOpenMemHandleExp may keep it safe. Please kindly double confirm it from UR.

btw,  zeMemGetIpcHandle with device memory allocation  and zeMemGetIpcHandleWithProperties with device or host memory allocation.  I suppose there may be some extension in UR in future.

dpctl/ipc/_ipc_memory.pyx
zxue2: pls refer "[Add aspect::ext_oneapi_ipc_memory for SyclDevice](https://github.com/IntelPython/dpctl/pull/2331/changes/352d2f84b8a9ea51192c223c4b94b0f665fcdfc9)"

dpctl/ipc/_ipc_memory.pyx
zxue2: Could you pls make it as follow-up improvement with separate PR given there should be no obvious risk for current use case: deterministic ipc open/get/close, explicit-lifetime IPC resource?

dpctl/ipc/_ipc_memory.pyx
zxue2: sure. updated

dpctl/ipc/_ipc_memory.pxd
zxue2: I would expect dpctl.ipc namespace to hold both memory ipc and event ipc .  What about revisiting it post event ipc coming?

ndgrigorian: I understand it would hold both functionality, but I think it is still more sensible to incorporate it with existing modules as opposed to adding this new very specific submodule.

dpctl/ipc/_ipc_memory.pyx
ndgrigorian: if this is the case, then let's add the check for `isinstance(memory, USMDeviceMemory)`

dpctl/ipc/_ipc_memory.pxd
ndgrigorian: that is to say, if we move current IPC implementation into dpctl.memory and add fallbacks for if it isn't present, it would be fine as is

dpctl/ipc/_ipc_memory.pyx
ndgrigorian: I will instead open a PR into this branch which fixes this issue, I would prefer it be fixed before we merge anything

dpctl/ipc/_ipc_memory.pxd
zxue2: moved to dpctl.mem

dpctl/ipc/_ipc_memory.pyx
zxue2: remove the class now

dpctl/ipc/_ipc_memory.pyx
ndgrigorian: I've added MemoryIPCDevice and shared pointer gymnastics to resolve concerns

dpctl/memory/_memory.pyx
```diff
@@ -961,3 +1024,306 @@ cdef api object Memory_Make(
...
+    returns an opaque ``bytes`` payload suitable for inter-process
+    transport (e.g. via pickle, ZMQ, shared-memory).
+
+    On the receiving side, call :meth:`IPCMemoryHandle.open` with the
+    payload and a target device to obtain a :class:`MemoryUSMDevice`
+    backed by the IPC-mapped memory.
+
+    Parameters
+    ----------
+    usm_memory : MemoryUSMDevice
```
zxue2, line 1070: Is it better to use the new MemoryIPCDevice based on  MemoryUSMDevice?

One more thing,  I am fine with current implementation which can satisfy current demand.  But restricting MemoryIPCDevice to base MemoryUSMDevice may lack extension for zeMemGetIpcHandleWithProperties that is with device or host memory allocation in future.  May have a separate jira there for improvement in future.

ndgrigorian: I figured we can extend to add additional classes later for host/shared allocations, so yes

dpctl/memory/_memory.pyx
```diff
@@ -961,3 +1024,306 @@ cdef api object Memory_Make(
...
+    RuntimeError
+        If IPC memory is not supported or the handle export fails.
+
+    Examples
+    --------
+    Sender process::
+
+        from dpctl.memory import MemoryUSMDevice, IPCMemoryHandle
+        mem = MemoryUSMDevice(4096)
+        handle = IPCMemoryHandle(mem)
```
antonwolfy, line 1086: Is the class meant to be public? Then it seems missed in `dpctl/memory/__init__.py` and `__all__` above/

dpctl/memory/_memory.pyx
```diff
@@ -892,6 +901,60 @@ cdef class MemoryUSMDevice(_Memory):
                     )
 
 
+cdef class MemoryIPCDevice(MemoryUSMDevice):
+    """
+    Class representing emory object backed by an IPC-mapped USM device pointer.
```
antonwolfy:
```suggestion
    Class representing memory object backed by an IPC-mapped USM device pointer.
```

libsyclinterface/include/syclinterface/dpctl_sycl_ipc_memory_interface.h
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * @defgroup IPCMemoryInterface IPC Memory Interface
+ */
+
+/*!
+ * @brief Get an IPC memory handle for a USM device pointer.
+ *
+ * Wraps ``sycl::ext::oneapi::experimental::ipc::memory::get()``.
+ * The returned handle bytes are copied out and the driver-side handle
+ * resource is released (via ``put``) before returning.
```
antonwolfy: The comment seems inaccurate, there is no `put` call inside `DPCTLIPCMem_GetHandle`

dpctl/memory/_memory.pyx
```diff
@@ -892,6 +901,60 @@ cdef class MemoryUSMDevice(_Memory):
...
+            raise ValueError("Referenced queue could not be copied.")
+        try:
+            base.queue = SyclQueue._create(QRef_copy)
+        except dpctl.SyclQueueCreationError as sqce:
+            raise ValueError(
+                "SyclQueue object could not be created from "
+                "copy of referenced queue"
+            ) from sqce
+
+        return MemoryIPCDevice(<object>base)
```
antonwolfy: If the construction fails due to some reason, there will be `__dealloc__` called, which will do `DPCTLIPCMem_CloseHandle(...)`, because we already did: `base._memory_ptr = USMRef`.

From other side. since the construction fails at that point, there will be exception raised, and so there will be the 2nd call of `DPCTLIPCMem_CloseHandle(...)` at line 1214, during the exception handling.

zxue2: pls kindly check "fix: prevent double-close of IPC memory handle in MemoryIPCDevice"

dpctl/memory/_memory.pyx
zxue2: IPCMemoryHandle is intentionally not in __all__ — users interact with IPC via the free functions (SyclIPCGetMemHandle, SyclIPCOpenMemHandle, SyclIPCCloseMemHandle)

dpctl/memory/_memory.pyx
zxue2: fixed

libsyclinterface/include/syclinterface/dpctl_sycl_ipc_memory_interface.h
zxue2: yes, it gets removed

zxue2:
also  add fix about "Remove unnecessary context parameters"

dpctl/memory/_memory.pyx
ndgrigorian: I think it would still be good to expose it, it can be constructed directly without any problems

ndgrigorian: also will allow us to document the class

ndgrigorian:
@zxue2
can we also add the IPC functions, IPCMemoryHandle, and MemoryIPCDevice to the docs? There is an RST for the memory, it should be straightfoward, if formatting is bad we can follow up

ndgrigorian (dismissed):
aside from above changes LGTM

zxue2:
> @zxue2 can we also add the IPC functions, IPCMemoryHandle, and MemoryIPCDevice to the docs? There is an RST for the memory, it should be straightfoward, if formatting is bad we can follow up

Please kindly check "public IPC class and update the docs"

dpctl/memory/_memory.pyx
zxue2: sure , updated

ndgrigorian (dismissed):
LGTM if no more comments from @antonwolfy

dpctl/memory/_memory.pyx
```diff
@@ -892,6 +902,70 @@ cdef class MemoryUSMDevice(_Memory):
...
+            base._memory_ptr = NULL  # caller retains ownership of USMRef
+            raise ValueError(
+                "SyclQueue object could not be created from "
+                "copy of referenced queue"
+            ) from sqce
+
+        try:
+            result = MemoryIPCDevice(<object>base)
+        except Exception:
+            # If MemoryIPCDevice.__cinit__ failed after _cinit_other copied
```
antonwolfy: There still might be an issue when the runtime classifies the mapped pointer as something other than "device", e.g. "unknown", and in that case there again will be double close:
```
open(raw_handle_bytes, device, nbytes=4096)                    _memory.pyx:1205
│
├─ mapped_ptr = DPCTLIPCMem_OpenHandle(...)   ✅ non-NULL      :1205
│     mapped_ptr = 0xABC   (a single live IPC mapping)
│
├─ q = SyclQueue(context, device)             ✅ ok           :1213
│
└─ try:                                                         :1219
   └─ create_ipc_from_usm_pointer_size_qref(mapped_ptr=0xABC)  :1221
      │
      ├─ base = _Memory(...);  base._memory_ptr = 0xABC        :940-943
      │
      └─ try:                                                   :957
         └─ MemoryIPCDevice(<object>base)                       :958
            │
            │  ── MemoryUSMDevice.__cinit__(base) ──            :884
            │   (1) self._cinit_other(base)                     :888
            │         self._memory_ptr = 0xABC   ← COPIED
            │         self._opaque_ptr = NULL
            │   (2) get_usm_type() != "device"  → TRUE          :889
            │   (3) copy=False → raise ValueError               :896  ⚠️
            │
            │  __cinit__ raised → CPython destroys partial obj
            │  → MemoryIPCDevice.__dealloc__:                   :917
            │       _memory_ptr (0xABC) is not NULL → TRUE
            │       ═══► DPCTLIPCMem_CloseHandle(0xABC)   ◄═══ CLOSE #1 🔴  :920
            │
            └─ except Exception:  base._memory_ptr = NULL       :959-964
               raise            ← re-raises the SAME ValueError
      │
      │  ValueError propagates out of create_ipc...
      │
      └─ except ValueError:                                     :1223
            # comment: "Failed before construction — we still own it"
            ═══► DPCTLIPCMem_CloseHandle(0xABC)          ◄═══ CLOSE #2 🔴  :1226
            raise

```

antonwolfy: Probably, at the moment it is not so much possible scenario. but it might be better to design `create_ipc_from_usm_pointer_size_qref` a way to take the ownership of the IPC mapping (ref by `USMRef`), since we already have the `DPCTLIPCMem_CloseHandle` call in `__dealloc__` method.

Here is what I mean:
```cython
    @staticmethod
    cdef void _close_ipc_mapping(
        DPCTLSyclUSMRef USMRef, DPCTLSyclQueueRef QRef
    ):
        """Close an IPC mapping, deriving its context from the queue ref."""
        cdef DPCTLSyclContextRef ctx_ref = DPCTLQueue_GetContext(QRef)
        DPCTLIPCMem_CloseHandle(USMRef, ctx_ref)
        DPCTLContext_Delete(ctx_ref)

    def __dealloc__(self):
        if self._memory_ptr is not NULL and self.queue is not None:
            MemoryIPCDevice._close_ipc_mapping(
                self._memory_ptr, self.queue.get_queue_ref()
            )
            self._memory_ptr = NULL

    @staticmethod
    cdef object create_ipc_from_usm_pointer_size_qref(
        DPCTLSyclUSMRef USMRef,
        Py_ssize_t nbytes,
        DPCTLSyclQueueRef QRef,
    ):
        """Create a MemoryIPCDevice wrapping an IPC-mapped pointer.

        This method takes ownership of the IPC mapping referenced by
        *USMRef*: on success the returned :class:`MemoryIPCDevice` owns it
        and closes it; on any failure the mapping is closed exactly once
        before the exception propagates.
        """
        cdef DPCTLSyclQueueRef QRef_copy = NULL
        cdef _Memory base

        if QRef is NULL:
            raise TypeError("Argument DPCTLSyclQueueRef is NULL")

        if nbytes <= 0:
            MemoryIPCDevice._close_ipc_mapping(USMRef, QRef)
            raise ValueError("Number of bytes must be positive")

        base = _Memory.__new__(_Memory)
        base._cinit_empty()
        base._memory_ptr = USMRef
        base.nbytes = nbytes

        QRef_copy = DPCTLQueue_Copy(QRef)
        if QRef_copy is NULL:
            base._memory_ptr = NULL
            MemoryIPCDevice._close_ipc_mapping(USMRef, QRef)
            raise ValueError("Referenced queue could not be copied.")

        try:
            base.queue = SyclQueue._create(QRef_copy)
        except dpctl.SyclQueueCreationError as sqce:
            base._memory_ptr = NULL
            MemoryIPCDevice._close_ipc_mapping(USMRef, QRef)
            raise ValueError(
                "SyclQueue object could not be created from "
                "copy of referenced queue"
            ) from sqce

        return MemoryIPCDevice(<object>base)
```
and so no try/except block or `DPCTLIPCMem_CloseHandle` while calling `create_ipc_from_usm_pointer_size_qref`:
```cython
        return MemoryIPCDevice.create_ipc_from_usm_pointer_size_qref(
            mapped_ptr, nbytes, q.get_queue_ref())
```

dpctl/tests/test_sycl_ipc_memory.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+def _get_ipc_device():
+    """Return a device with IPC memory support, or skip the test."""
+    try:
+        dev = dpctl.SyclDevice()
+    except dpctl.SyclDeviceCreationError:
+        pytest.skip("Default device not available")
+    if not dev.has_aspect_ext_oneapi_ipc_memory:
+        pytest.skip("Device does not support IPC memory")
```
antonwolfy, line 48: Do we need to check and to skip if `SyclIPCMemAvailable()` returns False?

dpctl/memory/_memory.pyx
```diff
@@ -961,3 +1035,304 @@ cdef api object Memory_Make(
...
+
+        from dpctl.memory import MemoryUSMDevice, IPCMemoryHandle
+        mem = MemoryUSMDevice(4096)
+        handle = IPCMemoryHandle(mem)
+        raw = handle.to_bytes()  # send to another process
+
+    Receiver process::
+
+        from dpctl.memory import IPCMemoryHandle
+        mapped_mem = IPCMemoryHandle.open(raw, device, nbytes=4096)
```
antonwolfy, line 1092: There is undefined device in the example

dpctl/memory/_memory.pyx
```diff
@@ -961,3 +1035,304 @@ cdef api object Memory_Make(
...
+
+
+def SyclIPCCloseMemHandle(_Memory usm_memory not None):
+    """Close an IPC memory mapping.
+
+    After this call, *usm_memory* is invalidated.
+
+    Parameters
+    ----------
+    usm_memory : _Memory
```
antonwolfy: `_Memory` is the internal type and isn't a part of public API and documentation
```suggestion
    usm_memory : MemoryIPCDevice
```

zxue2: Please refer commit "replace _Memory with MemoryUSMDevice for IPC functions"

dpctl/tests/test_sycl_ipc_memory.py
zxue2: Pls refer commit "add IPCMemoryHandle.is_available to test_sycl_ipc_memory.py"

dpctl/memory/_memory.pyx
zxue2: Please refer commit "Make create_ipc_from_usm_pointer_size_qref sole ownership of IPC mapping"

dpctl/memory/_memory.pyx
zxue2: It seems fine as an example

antonwolfy (approved):
Thank you @zxue2, no more comments from me

ndgrigorian (approved):

PR #2353, opened 2026-08-06
Remove `sub_group_independent_forward_progress` device info query

ndgrigorian:
This PR proposes the removal of the `sycl::info::device::sub_group_independent_forward_progress` query, which is now deprecated in SYCL 2020 and will be deprecated in 2026.2.0 compiler

### DIFF
--- a/dpctl/_sycl_device.pyx
+++ b/dpctl/_sycl_device.pyx
@@ -99,7 +99,6 @@ from ._backend cimport (  # noqa: E211
     DPCTLDevice_GetPreferredVectorWidthShort,
     DPCTLDevice_GetProfilingTimerResolution,
     DPCTLDevice_GetSingleFPConfig,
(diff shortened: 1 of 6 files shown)


dpctl/_sycl_device.pyx
```diff
@@ -1246,20 +1245,6 @@ cdef class SyclDevice(_SyclDevice):
         )
         return max_num_sub_groups
 
-    @property
-    def sub_group_independent_forward_progress(self):
```
antonwolfy, line 1250: I wonder if it would it be safe to remove without any release with deprecation notice?

ndgrigorian: I feel it should be ok, historically we have removed deprecated things (like the entire host aspect previously) without such deprecation notices. I don't think this aspect is in any common use.

antonwolfy (approved):
No more comments from me

PR #2354, opened 2026-08-06
Add `sycl::info::context` queries to `dpctl.SyclContext`

ndgrigorian:
This PR proposes exposing a number of previously-unexposed queries to `dpctl.SyclContext`

### DIFF
--- a/dpctl/_sycl_context.pyx
+++ b/dpctl/_sycl_context.pyx
@@ -34,7 +34,12 @@ from ._backend cimport (  # noqa: E211
     DPCTLContext_CreateFromDevices,
     DPCTLContext_Delete,
     DPCTLContext_DeviceCount,
+    DPCTLContext_GetAtomicFenceOrderCapabilities,
+    DPCTLContext_GetAtomicFenceScopeCapabilities,
+    DPCTLContext_GetAtomicMemoryOrderCapabilities,
+    DPCTLContext_GetAtomicMemoryScopeCapabilities,
     DPCTLContext_GetDevices,
+    DPCTLContext_GetPlatform,
     DPCTLContext_Hash,
     DPCTLDeviceMgr_GetCachedContext,
     DPCTLDeviceVector_CreateFromArray,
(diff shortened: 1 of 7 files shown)


libsyclinterface/tests/test_sycl_context_interface.cpp
```diff
@@ -181,6 +184,100 @@ TEST_P(TestDPCTLContextInterface, ChkGetBackend)
...
+{
+    DPCTLSyclContextRef CRef = nullptr;
+    int *arr = nullptr;
+    size_t len = 0;
+
+    EXPECT_NO_FATAL_FAILURE(CRef = DPCTLContext_Create(DRef, nullptr, 0));
+    ASSERT_TRUE(CRef);
+    EXPECT_NO_FATAL_FAILURE(
+        arr = DPCTLContext_GetAtomicMemoryOrderCapabilities(CRef, &len));
+    EXPECT_TRUE(len > 0);
```
antonwolfy: It's probably better to use `ASSERT_TRUE`, not `EXPECT_TRUE`, before the dereference loop.

In the 4 capability happy-path tests, `EXPECT_TRUE(arr != nullptr)` is non-fatal, then the code unconditionally loops `arr[i]`. If a query ever returned `len > 0` with `arr == nullptr`, it segfaults past the failed expectation.

It should be better to change to `ASSERT_TRUE(arr != nullptr)` (and ideally `ASSERT_TRUE(len > 0)`).

dpctl/_sycl_context.pyx
```diff
@@ -84,6 +94,33 @@ cdef void _init_helper(_SyclContext context, DPCTLSyclContextRef CRef):
     context._ctxt_ref = CRef
 
 
+cdef tuple _to_enum_tuple(
```
antonwolfy, line 97: Is that intended, that it's exactly the same as `_to_enum_tuple` in `dpctl/_sycl_device.pyx`?

antonwolfy (dismissed):
LGTM with minor nits below

dpctl/_sycl_context.pyx
ndgrigorian: It is, unfortunately it's a bit inconvenient to share such methods between Cython files without introducing some new common file. It can potentially be refactored later.

antonwolfy (approved):
No more comments from me

PR #2355, opened 2026-08-06
remove work-around from Windows build script

ndgrigorian:
This PR proposes alignment with https://github.com/IntelPython/dpnp/pull/3013 in removing an obsolete work-around from the Windows build script

Also bumps the clang version hint (`lib/clang/22/` is present in 2026.1.1)


antonwolfy:
The w/a can only be removed when installing the compiler from Intel channel.
But still required when installing the compiler from conda-forge channel, it looks as it was intentional the repack design decision.

PR #2359, opened 2026-08-10
Remove `host_task` object management

ndgrigorian:
Inspired by comment in: https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1915

we move away from use of `host_task`, which enables compatibility with AdaptiveCpp, and fixes long-standing undefined behavior of calling `sycl::free` from within a `host_task`. The `host_task` decrementing the refcount would cause the `USMDeleter` to be called in its scope, which is undefined behavior under the SYCL spec

Instead, we use a polling thread which checks for event completion, then decreases references


dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -823,46 +824,26 @@ sycl::event keep_args_alive(sycl::queue &q,
...
+                        shp_arr[i]->dec_ref();
+                    }
                 }
             });
-        });
     }
 
-    return host_task_ev;
+    // return dummy event for API compatibility
+    return sycl::event{};
```
vlad-perevezentsev, line 877: Should we document this or add a CHANGELOG note that now `keep_args_alive` in `dpctl4pybind11.,hpp` returns an empty (already-complete) event instead of `host_task` event?

dpctl/apis/include/dpctl4pybind11.hpp
vlad-perevezentsev, line 828: q is now unused
Should we add  `(void)q` like `async_dec_ref_event` does for `QRef`?

dpctl/apis/include/detail/keep_alive_pool.hpp
```diff
@@ -0,0 +1,125 @@
...
+namespace detail
+{
+
+class KeepAlivePool
+{
+public:
+    /*!
+     * @brief Number of waiter threads.
+     */
+    static constexpr std::size_t num_threads = 4;
```
vlad-perevezentsev: Why was 4 threads chosen? What happens if there are more than 4 concurrent tasks?

dpctl/utils/_order_manager.py
```diff
@@ -25,6 +26,14 @@ def __del__(self):
         SyclEvent.wait_for(_local.get_host_task_events())
```
vlad-perevezentsev, line 27: Could you add a comment here noting that this line should be removed once the deprecated `add_event_pair` is gone? 
Because with the new `add_event` API `get_host_task_events()` is always empty

dpctl/apis/include/dpctl4pybind11.hpp
ndgrigorian: I can document it, hard to say where, maybe in the new docstring for the free function

dpctl/apis/include/dpctl4pybind11.hpp
ndgrigorian: Good idea

dpctl/apis/include/detail/keep_alive_pool.hpp
ndgrigorian: The compiler chose 4, and the compiler thread pool implementation is what this is based off of.

With more than 4, it will just mean that the releasing of some references will be held up by the releasing of other references, which is dependent on the work being done. So in theory, a large computational task could hold up the freeing of several small tasks, but I'm not sure this is a problem in practice.

dpctl/utils/_order_manager.py
ndgrigorian: Good idea

docs/doc_sources/api_reference/dpctl/utils.rst
```diff
@@ -15,3 +15,13 @@
 
     Thread-local object mapping each :class:`dpctl.SyclQueue` to an order
     manager, used to ensure sequential ordering of offloaded tasks.
+
+    Record submitted tasks with ``add_event`` and use ``submitted_events``
+    as the dependency list of subsequent submissions. To keep Python objects
+    referenced by a task alive until it completes, use
+    :func:`dpctl.keep_args_alive`.
+
+    .. deprecated::
```
antonwolfy: It seems the deprecation notice rendered wrongly:
> Deprecated since version ``add_event_pair``,: host_task_events...

dpctl/apis/include/detail/keep_alive_pool.hpp
antonwolfy: Should we make that configurable through the build-time option? Like the compiler which exposes `SYCL_QUEUE_THREAD_POOL_SIZE`.

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -823,46 +827,26 @@ sycl::event keep_args_alive(sycl::queue &q,
...
-                cgh.depends_on(depends);
-                use_depends = false;
-            }
-            else {
-                cgh.depends_on(host_task_ev);
-            }
-            cgh.host_task([n_objects_held, shp_arr = std::move(shp_arr)]() {
-                py::gil_scoped_acquire acquire;
+                if (n_objects_held > 0) {
+                    py::gil_scoped_acquire acquire;
```
antonwolfy, line 867: Do we need to prevent unguarded GIL acquisition during finalization (similarly to `async_dec_ref`):
```hpp
#if PY_VERSION_HEX < 0x30d0000
    const bool finalizing = _Py_IsFinalizing();
#else
    const bool finalizing = Py_IsFinalizing();
#endif
    if (Py_IsInitialized() && !finalizing) { ... }
```

dpctl/apis/include/detail/keep_alive_pool.hpp
```diff
@@ -0,0 +1,125 @@
...
+                std::unique_lock<std::mutex> lock(queue_mutex_);
+                condition_.wait(lock, [this] { return !tasks_.empty(); });
+                item = std::move(tasks_.front());
+                tasks_.pop();
+            }
+
+            try {
+                sycl::event::wait(item.first);
+            } catch (const std::exception &) {
+                // run the task anyway: an async error must not strand the
```
antonwolfy, line 115: It might be good to add any verbose logic here later, in the follow-up for debugging purpose.

ndgrigorian: Good idea, we will need to figure out how to handle the errors as well

dpctl/_sycl_queue.pyx
```diff
@@ -1236,7 +1257,7 @@ cdef class SyclQueue(_SyclQueue):
             with nogil:
                 DPCTLEvent_Wait(htERef)
             DPCTLEvent_Delete(htERef)
-            raise RuntimeError("Could not submit keep_args_alive host_task")
+            raise RuntimeError("Could not schedule keep_args_alive")
```
antonwolfy, line 1433: Missing `Py_DECREF(args)` leading to reference leak, while it is handled properly in `keep_args_alive`.

dpctl/apis/include/detail/keep_alive_pool.hpp
```diff
@@ -0,0 +1,135 @@
...
+                // task or else it may leak
+            }
+
+            try {
+                item.second();
+            } catch (const std::exception &) {
```
antonwolfy, line 121: `item.second()` is an arbitrary `std::function<void()>` handed in through the public submit() API. If a caller's task throws anything not derived from `std::exception` (a bare `int`, a custom type, etc.), it escapes `run()`, escapes the `std::thread` entry function, and the C++ runtime calls `std::terminate` — killing the whole process. 
```suggestion
            } catch (...) {
```

antonwolfy: And probably the same at line#114

dpctl/apis/include/dpctl4pybind11.hpp
antonwolfy, line 828: @vlad-perevezentsev, is that intended in dpnp we are using the local copy of that function `dpnp::utils::keep_args_alive` instead of dpctl vendored function?

Btw, due to that there is no real impact on dpnp due to that dpctl migration change, except new deprecation wanrnings.

docs/doc_sources/beginners_guides/installation.rst
```diff
@@ -247,6 +247,24 @@ devices at the same time:
...
+To keep Python objects used by an offloaded task alive until tasks complete,
+:py:mod:`dpctl` maintains a pool of threads that wait on the task's events.
+The pool uses four threads by default. The size is fixed when ``dpctl`` is
+compiled and can be changed with the ``--keep-alive-pool-size`` argument:
+
+.. code-block:: bash
+
+    python scripts/build_locally.py --verbose --keep-alive-pool-size=8
+
+Alternatively, you use the ``DPCTL_KEEP_ALIVE_POOL_SIZE`` CMake option:
```
antonwolfy, line 262:
```suggestion
Alternatively, you can use the ``DPCTL_KEEP_ALIVE_POOL_SIZE`` CMake option:
```

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -802,6 +829,9 @@ sycl::event keep_args_alive(sycl::queue &q,
                             const py::object (&py_objs)[num],
                             const std::vector<sycl::event> &depends = {})
 {
+    // q is only retained for API compatibility
+    (void)q;
```
antonwolfy, line 833: There might be potential loss of implicit in-order-queue serialization.

The below example is pure hypothetical. The below pattern was safe before and is a use-after-free now:
```cpp
sycl::queue q{ctx, dev, sycl::property::queue::in_order{}};   // in-order

// kernel writes into USM buffer owned by python object `arr`
q.submit([&](sycl::handler &cgh) {
    cgh.parallel_for(range, [=](id<1> i){ data[i] = ...; });   // event NOT captured
});

// keep `arr` alive until the queue's work is done.
// Caller relies on q being in-order, so passes empty depends:
keep_args_alive(q, {arr}, /*depends=*/{});
```

- **Old behavior**: the release `host_task` is enqueued on the in-order `q` after the kernel, so `arr`'s dec_ref (and any USM free) happens only once the kernel completes. Correct.
- **New behavior**: `depends` is empty, so the pool worker calls `sycl::event::wait({})` — waits on nothing — and runs the release immediately, decref'ing/freeing `arr` while the kernel is still reading its buffer → use-after-free / data corruption, silently, with no compile error.

The hazard is real for any caller that passed empty or partial `depends` and leaned on in-order queue semantics, while dpnp is not affected.

dpctl/apis/include/detail/keep_alive_pool.hpp
```diff
@@ -0,0 +1,135 @@
...
+     */
+    static constexpr std::size_t num_threads = DPCTL_KEEP_ALIVE_POOL_SIZE;
+
+    static_assert(num_threads > 0,
+                  "DPCTL_KEEP_ALIVE_POOL_SIZE must be greater than zero");
+
+    /*!
+     * @brief The instance belonging to this shared object.
+     */
+    static KeepAlivePool &local_instance()
```
antonwolfy, line 64: Note, the below is a pure AI detected observation, but sounds fair.

**What the pool does at construction.** `keep_alive_pool.hpp:64-99`: `local_instance()` is a leaked singleton (`new`, `~KeepAlivePool() = delete`), and the constructor spawns N threads and immediately `detach()`es them. `run()` is an infinite `for(;;)` loop, so the workers **never return** — they run until the process dies. There is no stop flag, no join, no drain.

**Why that's a shutdown hazard.** At interpreter/process teardown, a detached worker can be in one of three states, each touching a subsystem that is concurrently being torn down:

1. **Blocked in `sycl::event::wait(item.first)`** (`:113`) on events whose backing SYCL context is being destroyed by the runtime's static destructors → waiting on an event of a dead context is UB (hang/crash).
2. **Inside a USM deleter**, `shp_usm[i].reset()` → `sycl::free(ptr, ctx)` (`:861` of the template). If the SYCL context is already gone, `sycl::free` on it is UB.
3. **Inside `py::gil_scoped_acquire`** (`:867`) during/after `Py_Finalize()`.

State 3 is guarded by `interpreter_is_live()` (checks `!Py_IsFinalizing()`), but only against a *TOCTOU window*: the check can pass, then the main thread begins finalization, then the worker acquires the GIL → hang or crash. The check narrows the window; it cannot close it, because there's no handshake between the main thread's finalization and the detached workers.

**No drain means abandoned work.** Any tasks still in `tasks_` (or in-flight) when the process exits are simply dropped — their dec_ref/free never run. This is **usually benign**: at true process exit the OS reclaims memory and leaked Python refs don't matter. The real risk is a **crash or hang *at* exit** — a segfault in `sycl::free` on a torn-down context, or a GIL-acquire hang — which surfaces as flaky, hard-to-reproduce "crashes on interpreter shutdown" reports rather than any wrong-answer bug during normal use.

**Contrast with the old model:** release work was tied to queue `q`; the SYCL runtime's own teardown ordering (and the order manager's event waits during normal interpreter shutdown) kept the work coupled to live contexts. The pool decouples release onto free-running threads that coordinate with *neither* the Python interpreter shutdown *nor* SYCL static destruction.

**Mitigations worth considering:** a `Py_AtExit`/`atexit` (or static-flag) drain that, with a timeout, waits for `tasks_` to empty and in-flight tasks to finish before finalization proceeds; and/or a stop flag so workers can break the loop. At minimum, document the contract ("all `depends` must complete before interpreter/process exit"), which the docstrings currently do only partially.

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -823,46 +853,28 @@ sycl::event keep_args_alive(sycl::queue &q,
...
-            }
-            cgh.host_task([shp_usm = std::move(shp_usm)]() {
-                // no body, but shared pointers are captured in
-                // the lambda, ensuring that USM allocation is
-                // kept alive
-            });
-        });
-    }
+    if (n_usm_owners_held > 0 || n_objects_held > 0) {
+        dpctl::detail::get_keep_alive_pool().submit(
```
antonwolfy, line 857: There might be a refcount leak if an exception is thrown before/during submit() in the public template.
Needs a scope guard that dec_refs the collected handles unless submit() succeeded.

dpctl/_async_dec_ref.hpp
```diff
@@ -0,0 +1,134 @@
...
+ */
+void async_dec_ref(PyObject **obj_array,
+                   size_t obj_array_size,
+                   DPCTLSyclEventRef *depERefs,
+                   size_t nDepERefs,
+                   int *status)
+{
+    using dpctl::syclinterface::unwrap;
+
+    std::vector<PyObject *> obj_vec(obj_array, obj_array + obj_array_size);
```
antonwolfy, line 73: That might throw `bad_alloc`, but outside try/catch block

dpctl/CMakeLists.txt
```diff
@@ -199,8 +199,16 @@ endforeach()
 set(_cy_file ${CMAKE_CURRENT_SOURCE_DIR}/_sycl_queue.pyx)
 get_filename_component(_trgt ${_cy_file} NAME_WLE)
 build_dpctl_ext(${_trgt} ${_cy_file} "dpctl" SYCL)
-# _sycl_queue include _host_task_util.hpp
-target_include_directories(${_trgt} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+# _sycl_queue includes _async_dec_ref.hpp, which includes
```
antonwolfy, line 202: This is AI agentic comment which seems having fair concern on ODR pitfall with real example.

**The setup, now confirmed:**
- `num_threads` is baked into an inline header: `static constexpr std::size_t num_threads = DPCTL_KEEP_ALIVE_POOL_SIZE;` (`keep_alive_pool.hpp:56`), default 4 (`:41-43`).
- The macro is applied **PRIVATE to the `_sycl_queue` target only** (`dpctl/CMakeLists.txt:208-210`), with the comment *"only it needs to know the pool size."* So `_sycl_queue`'s TU compiles the header with, say, `=16`; **every other TU** that includes `dpctl4pybind11.hpp` → `detail/keep_alive_pool.hpp` (other dpctl extensions, and downstream dpnp / dpctl.tensor / numba-dpex) sees the header **default of 4**.
- The `KeepAlivePool()` constructor (`:94-99`) is an **inline member function** (external linkage) that uses `num_threads`. The ODR requires all TUs to see an *identical* definition — but its definition differs (loop bound 16 vs 4). Same for the inline `local_instance()`. **This is an ODR violation**; the linker silently keeps one arbitrary definition and discards the rest.

**Real example where it bites:**
1. dpctl built with `-DDPCTL_KEEP_ALIVE_POOL_SIZE=16`. Only `_sycl_queue.so`'s TU has the `=16` definition; the installed header still reads 4.
2. A future dpnp adopts dpctl's pool and, in its own `.so`, includes the header (sees 4) and calls `KeepAlivePool::local_instance()` **directly** (it's `public`, so nothing stops it) to submit its own release tasks.
3. The program now links two conflicting definitions of the constructor — one spawning 16 threads, one spawning 4. Because the symbols are inline/external and both TUs ODR-use them, the dynamic linker collapses them to **one arbitrary winner** (depends on link order / symbol interposition). dpctl configured 16 but may get a pool of 4, or the reverse — non-deterministic across platforms and build configs, and formally UB.

**Why it's benign today:** the constructor / `local_instance()` are ODR-used **only inside `_sycl_queue`** (via `_async_dec_ref.hpp`, included by exactly one TU). Everyone else reaches the pool solely through `get_keep_alive_pool()` → `KeepAlivePool_Get()` (`dpctl4pybind11.hpp:309-322`) — a C-API function pointer exported by `_sycl_queue`, which routes to `_sycl_queue`'s own correctly-sized instance. So only one definition is actually emitted and the violation stays latent. It goes **live the moment any other TU ODR-uses the constructor** (a direct `local_instance()` call).

**How to make it robust:** don't bake a macro-varying value into an inline definition. Options: read the size once at runtime (env var) inside a single non-inline function defined only in `_sycl_queue`; or make `local_instance()` private / non-public so downstream *cannot* instantiate the pool and is forced through `KeepAlivePool_Get()`; or at minimum document that the header default is authoritative for any TU other than `_sycl_queue`, so a customized size only takes effect for the canonical instance.

dpctl/_sycl_queue.pyx
```diff
@@ -2024,3 +2049,74 @@ cdef class RawKernelArg:
         as a ``size_t``.
         """
         return <size_t>self._arg_ref
+
+
+def keep_args_alive(args, depends):
```
antonwolfy, line 2484: It might be helpful to add tests exercise the actual pool release path (refcount returns to baseline after gating events complete), the shutdown/finalization guard:

```python
import sys, gc, time, weakref, dpctl

def test_keep_args_alive_releases_after_event(q):
    class S: pass
    obj = S()
    ref = weakref.ref(obj)
    base = sys.getrefcount(obj)
    e = q.submit_barrier()            # a real, completable SyclEvent
    dpctl.keep_args_alive((obj,), [e])
    assert sys.getrefcount(obj) == base + 1   # incref happened synchronously
    e.wait()
    del obj
    # with a drain hook: dpctl.utils._wait_pool_idle(); assert ref() is None
    for _ in range(500):              # fallback: bounded poll
        gc.collect()
        if ref() is None: break
        time.sleep(0.01)
    assert ref() is None              # pool ran the deferred decref

def test_shutdown_with_pending_keepalive():
    script = r"""
import dpctl, faulthandler; faulthandler.enable()
q = dpctl.SyclQueue()
e = q.submit_barrier()          # or a long kernel to leave work in-flight
class S: pass
dpctl.keep_args_alive((S(),), [e])
# intentionally do NOT wait -> exercise release racing interpreter teardown
"""
    p = subprocess.run([sys.executable, "-c", script],
                       timeout=60, capture_output=True)
    assert p.returncode == 0
    assert b"Fatal Python error" not in p.stderr and b"Aborted" not in p.stderr

```

dpctl/apis/include/detail/keep_alive_pool.hpp
antonwolfy: At least we need to document somewhere here explicitly, something similar to note in `dpctl.keep_args_alive` docstrings.

dpctl/_async_dec_ref.hpp
```diff
@@ -0,0 +1,134 @@
+//===--- _async_dec_ref.hpp - Implements async DECREF ---------------------===//
+//
+//                      Data Parallel Control (dpctl)
+//
+// Copyright 2022 Intel Corporation
```
antonwolfy, line 5:
```suggestion
// Copyright 2026 Intel Corporation
```

dpctl/apis/include/detail/keep_alive_pool.hpp
vlad-perevezentsev: `unique_lock`. `condition_.wait`. `tasks_.pop()` are outside both try blocks 
If they throw, it also escapes `run()` and the thread triggering `std::terminate`
Should we wrap them in try as well?

dpctl/apis/include/dpctl4pybind11.hpp
```diff
@@ -174,15 +189,7 @@ class dpctl_capi
     {
         void operator()(py::object *p) const
         {
-            const bool initialized = Py_IsInitialized();
```
vlad-perevezentsev, line 176: Why was `Py_IsInitialized()` check removed in the new `interpreter_is_live()`?
Was that intentional?

ndgrigorian: Nope, probably an error of copying it over into a separate function, good catch

dpctl/_sycl_queue.pyx
ndgrigorian: looks like it was never present...

dpctl/apis/include/detail/keep_alive_pool.hpp
ndgrigorian: Py_AtExit drain would be the wrong solution, it's best to let the OS clean it up, I feel. Other projects don't implement such a thing for their own shutdowns, etc.

But it's possible we can document it

dpctl/apis/include/dpctl4pybind11.hpp
ndgrigorian: Unfortunately this is fair, my hope was to not have to implement an empty kernel submission as a means of maintaining the refcounts, but it seems to be necessary for in-order queues, so I've added it.

dpctl/_sycl_queue.pyx
ndgrigorian: My concern with a test like this is that it's reliant on specific work happening on the queue, with specific timing. Tests like this feel very unreliable and flaky.

It may be more possible now that we need a dummy event again in some cases.

dpctl/apis/include/detail/keep_alive_pool.hpp
ndgrigorian: they shouldn't throw, and according to what I've looked into, `std::unique_lock` should only throw `std::system_error` which is unrecoverable, but the `catch (...)` has been implemented

dpctl/apis/include/detail/keep_alive_pool.hpp
ndgrigorian: I ultimately did end up with an atexit hook

dpctl/_deferred_release_watcher.hpp
```diff
@@ -0,0 +1,358 @@
...
+     * @brief Run the releases that come due, until the list is stopped or the
+     * interpreter begins shutting down.
+     */
+    static void watch()
+    {
+        DeferredReleases &releases = local_deferred_releases();
+        clock::duration poll_interval = min_poll_interval;
+        std::vector<std::function<void()>> due;
+
+        while (releases.wait_for_due(poll_interval)) {
```
vlad-perevezentsev: Should this loop have a `catch()`?

Both `collect_due` and `wait_for_due` seems can throw
If an exception escapes the thread entry function, it will call `std::terminate` and abort the whole process.

I see similar protection is already used in `is_complete` and `run_collected`

dpctl/_sycl_queue.pyx
```diff
@@ -1337,81 +1346,144 @@ cdef class SyclQueue(_SyclQueue):
         """
         return <size_t>self._queue_ref
 
-    cpdef SyclEvent _submit_keep_args_alive(
-        self,
-        object args,
-        list dEvents
-    ):
-        """ SyclQueue._submit_keep_args_alive(args, events)
+    cpdef SyclEvent keep_args_alive(self, object args, object depends=None):
```
vlad-perevezentsev: Why is `depends` typed as object here while `submit_async` and `submit_barrier` use `depends` as list?

With a generic `object` `len(depends)` may not match iteration which can lead to invalid access in the event buffer

PR #2360, opened 2026-08-11
Pin compiler to minimum `2026.0.0`

ndgrigorian:
A number of new features in dpctl require newer compiler versions

This PR updates the minimum required DPC++ compiler version to 2026.0.0, as later versions should be compatible, but this sets a floor for where some features may be lost, and will allow the removal of some macros hiding features from later versions from earlier compilers

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set required_compiler_version = "2024.2.0" %}
+{% set required_compiler_version = "2026.0.0" %}
 
 {% set pyproject = load_file_data('pyproject.toml') %}
 {% set py_build_deps = pyproject.get('build-system', {}).get('requires', []) %}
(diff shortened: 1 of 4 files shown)


conda-recipe/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set required_compiler_version = "2024.2.0" %}
+{% set required_compiler_version = "2026.0.0" %}
```
antonwolfy, line 1: Need to update `docs/doc_sources/contributor_guides/building.rst` which mentions "oneAPI DPC++ 2024.0 or later."

antonwolfy, line 1: Do we need to bump the compile-time floor in `dpctl_config.h.in:31`?
> #define __SYCL_COMPILER_VERSION_REQUIRED 20221201L

ndgrigorian: Done

antonwolfy (approved):
Thank you @ndgrigorian

PR #2361, opened 2026-08-13
Add `dpctl.SyclQueue.memset()` method

vlad-perevezentsev:
This PR proposes adding `dpctl.SyclQueue.memset` and `dpctl.SyclQueue.memset_async` methods  as a Python wrapper over `sycl::queue::memset` backed by the existing `DPCTLQueue_Memset` and a new `DPCTLQueue_MemsetWithEvents` C-API function and adding a new `test_sycl_queue_memset.py`

### DIFF
--- a/dpctl/_backend.pxd
+++ b/dpctl/_backend.pxd
@@ -21,7 +21,7 @@
 types defined by dpctl's C API.
 """
 
-from libc.stdint cimport int64_t, uint32_t, uint64_t
+from libc.stdint cimport int64_t, uint8_t, uint32_t, uint64_t
 from libcpp cimport bool
 
 
@@ -677,12 +677,12 @@ cdef extern from "syclinterface/dpctl_sycl_queue_interface.h":
     cdef DPCTLSyclEventRef DPCTLQueue_Memset(
         const DPCTLSyclQueueRef Q,
         void *Dest,
-        int Val,
+        uint8_t Val,
         size_t Count)
     cdef DPCTLSyclEventRef DPCTLQueue_MemsetWithEvents(
         const DPCTLSyclQueueRef Q,
         void *Dest,
-        int Val,
+        uint8_t Val,
         size_t Count,
         const DPCTLSyclEventRef *depEvents,
         size_t depEventsCount)
--- a/dpctl/_backend.pxd
+++ b/dpctl/_backend.pxd
@@ -679,6 +679,13 @@ cdef extern from "syclinterface/dpctl_sycl_queue_interface.h":
         void *Dest,
         int Val,
         size_t Count)
+    cdef DPCTLSyclEventRef DPCTLQueue_MemsetWithEvents(
+        const DPCTLSyclQueueRef Q,
+        void *Dest,
+        int Val,
+        size_t Count,
+        const DPCTLSyclEventRef *depEvents,
+        size_t depEventsCount)
     cdef DPCTLSyclEventRef DPCTLQueue_Prefetch(
         const DPCTLSyclQueueRef Q,
         const void *Src,
--- a/dpctl/_sycl_queue.pxd
+++ b/dpctl/_sycl_queue.pxd
@@ -107,6 +107,7 @@ cdef public api class SyclQueue (_SyclQueue) [
     cpdef SyclEvent copy_async(
         self, dest, src, size_t count, list dEvents=*, str dtype=*
     )
+    cpdef memset(self, mem, int val, size_t count=*)
     cpdef prefetch(self, ptr, size_t count=*)
     cpdef mem_advise(self, ptr, size_t count, int mem)
     cpdef SyclEvent submit_barrier(self, dependent_events=*)
--- a/dpctl/_sycl_queue.pxd
+++ b/dpctl/_sycl_queue.pxd
@@ -108,6 +108,9 @@ cdef public api class SyclQueue (_SyclQueue) [
         self, dest, src, size_t count, list dEvents=*, str dtype=*
     )
     cpdef memset(self, mem, int val, size_t count=*)
+    cpdef SyclEvent memset_async(
+        self, mem, int val, size_t count=*, list dEvents=*
+    )
     cpdef prefetch(self, ptr, size_t count=*)
     cpdef mem_advise(self, ptr, size_t count, int mem)
     cpdef SyclEvent submit_barrier(self, dependent_events=*)
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -1674,6 +1674,11 @@ cdef class SyclQueue(_SyclQueue):
         Internally, this dispatches ``sycl::queue::memset``. The operation is
         byte-wise: ``count`` bytes are set, each to the same value ``val``.
 
+        Note:
+            The returned event does not keep ``mem`` alive. Keep ``mem``
+            alive until the event completes, otherwise its USM allocation
+            may be freed mid-operation, causing a use-after-free.
+
         Args:
             mem:
                 Destination USM allocation, an instance of
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -1631,6 +1631,9 @@ cdef class SyclQueue(_SyclQueue):
         Internally, this dispatches ``sycl::queue::memset``. The operation is
         byte-wise: ``count`` bytes are set, each to the same value ``val``.
 
+        This is a synchronizing variant corresponding to
+        :meth:`dpctl.SyclQueue.memset_async`.
+
         Args:
             mem:
                 Destination USM allocation, an instance of
@@ -1646,6 +1649,8 @@ cdef class SyclQueue(_SyclQueue):
         Raises:
             TypeError:
                 If ``mem`` is not an instance of :class:`dpctl.memory._Memory`.
+            RuntimeError:
+                If the memset operation encountered an error.
         """
         cdef DPCTLSyclEventRef ERef = NULL
 
@@ -1689,6 +1694,8 @@ cdef class SyclQueue(_SyclQueue):
             TypeError:
                 If ``mem`` is not an instance of :class:`dpctl.memory._Memory`,
                 or ``dEvents`` is not a sequence of :class:`dpctl.SyclEvent`.
+            RuntimeError:
+                If the memset operation encountered an error.
         """
         cdef DPCTLSyclEventRef ERef = NULL
         cdef DPCTLSyclEventRef *depEvents = NULL
@@ -1717,7 +1724,7 @@ cdef class SyclQueue(_SyclQueue):
 
         if (ERef is NULL):
             raise RuntimeError(
-                "SyclQueue.memset operation encountered an error"
+                "SyclQueue.memset_async operation encountered an error"
             )
 
         return SyclEvent._create(ERef)
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -49,6 +49,7 @@ from ._backend cimport (  # noqa: E211
     DPCTLQueue_MemAdvise,
     DPCTLQueue_Memcpy,
     DPCTLQueue_MemcpyWithEvents,
+    DPCTLQueue_Memset,
     DPCTLQueue_Prefetch,
     DPCTLQueue_SubmitBarrierForEvents,
     DPCTLQueue_SubmitNDRange,
@@ -1593,6 +1594,48 @@ cdef class SyclQueue(_SyclQueue):
             )
 
         return SyclEvent._create(ERef)
+
+    cpdef memset(self, mem, int val, size_t count=0):
+        """Fill USM allocation ``mem`` with the byte value ``val`` and wait.
+
+        Internally, this dispatches ``sycl::queue::memset``. The operation is
+        byte-wise: ``count`` bytes are set, each to the same value ``val``.
+
+        Args:
+            mem:
+                Destination USM allocation, an instance of
+                :class:`dpctl.memory._Memory`.
+            val (int):
+                Value to fill ``mem`` with. Following ``sycl::queue::memset``,
+                it is interpreted as an ``unsigned char``, i.e. only the least
+                significant byte is used.
+            count (int, optional):
+                Number of bytes to fill. If ``0`` or greater than the size of
+                ``mem``, the whole allocation is filled. Default: ``0``.
+
+        Raises:
+            TypeError:
+                If ``mem`` is not an instance of :class:`dpctl.memory._Memory`.
+        """
+        cdef void *ptr
+        cdef DPCTLSyclEventRef ERef = NULL
+
+        if isinstance(mem, _Memory):
+            ptr = <void*>(<_Memory>mem).get_data_ptr()
+        else:
+            raise TypeError("Parameter `mem` should have type _Memory")
+
+        if (count <= 0 or count > mem.nbytes):
+            count = mem.nbytes
+
+        ERef = DPCTLQueue_Memset(self._queue_ref, ptr, val, count)
+        if (ERef is NULL):
+            raise RuntimeError(
+                "SyclQueue.memset operation encountered an error"
+            )
+        with nogil:
+            DPCTLEvent_Wait(ERef)
+        DPCTLEvent_Delete(ERef)
 
     cpdef prefetch(self, mem, size_t count=0):
         cdef void *ptr
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -1650,6 +1650,9 @@ cdef class SyclQueue(_SyclQueue):
         Raises:
             TypeError:
                 If ``mem`` is not an instance of :class:`dpctl.memory._Memory`.
+            OverflowError:
+                If ``val`` does not fit in a C ``int`` or ``count`` is
+                negative.
             RuntimeError:
                 If the memset operation encountered an error.
         """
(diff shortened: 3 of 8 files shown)


vlad-perevezentsev:
@ndgrigorian I implemented `cpdef memset` without validating `val`, i.e. matching `sycl::queue::memset` semantics where the value is silently truncated toits least significant byte.
Does it make sense to explicitly validate that `val` is within `[0, 255]`?

dpctl/_sycl_queue.pyx
```diff
@@ -1594,6 +1595,48 @@ cdef class SyclQueue(_SyclQueue):
 
         return SyclEvent._create(ERef)
 
+    cpdef memset(self, mem, int val, size_t count=0):
```
ndgrigorian, line 1629: Can we also add `memset_async` variant that actually returns the event?

vlad-perevezentsev: Yes, sure
But to do that I will need to extend dpctl C-API with `DPCTLQueue_MemsetWithEvents`.
I will implement it

vlad-perevezentsev: Done in [fd541c0](https://github.com/IntelPython/dpctl/pull/2361/commits/fd541c008a91c65e29d2445ee5ba24b265c090a0)

dpctl/_sycl_queue.pyx
vlad-perevezentsev, line 1633:
```
This is a synchronizing variant corresponding to
:meth:`dpctl.SyclQueue.memset_async`.
```

dpctl/_sycl_queue.pyx
```diff
@@ -1594,6 +1625,103 @@ cdef class SyclQueue(_SyclQueue):
...
+                        raise TypeError(
+                            "A sequence of dpctl.SyclEvent is expected"
+                        )
+                ERef = _memset_impl(self, mem, val, count, depEvents, nDE)
+            finally:
+                free(depEvents)
+
+        if (ERef is NULL):
+            raise RuntimeError(
+                "SyclQueue.memset operation encountered an error"
```
ndgrigorian:
```suggestion
                "SyclQueue.memset_async operation encountered an error"
```

dpctl/_sycl_queue.pyx
```diff
@@ -1594,6 +1625,103 @@ cdef class SyclQueue(_SyclQueue):
...
+                Number of bytes to fill. If ``0`` or greater than the size of
+                ``mem``, the whole allocation is filled. Default: ``0``.
+            dEvents (List[dpctl.SyclEvent], optional):
+                Events that this operation depends on.
+
+        Returns:
+            dpctl.SyclEvent:
+                Event associated with the memset operation.
+
+        Raises:
```
ndgrigorian, line 1703: We can mention the `RuntimeError` here, same as in `memset`

dpctl/tests/test_sycl_queue_memset.py
```diff
@@ -0,0 +1,166 @@
...
+
+"""Defines unit test cases for the SyclQueue.memset."""
+
+import pytest
+
+import dpctl
+import dpctl.memory
+
+
+def _create_memory(q, nbytes=1024):
```
ndgrigorian, line 26: I think it would be good to add host and device memory to at least a few tests

dpctl/_sycl_queue.pxd
```diff
@@ -107,6 +107,10 @@ cdef public api class SyclQueue (_SyclQueue) [
     cpdef SyclEvent copy_async(
         self, dest, src, size_t count, list dEvents=*, str dtype=*
     )
+    cpdef memset(self, mem, int val, size_t count=*)
+    cpdef SyclEvent memset_async(
+        self, mem, int val, size_t count=*, list dEvents=*
+    )
```
ndgrigorian, line 113: we declare `int` here when technically it's `uint8_t`

dpctl/_backend.pxd
```diff
@@ -679,6 +679,13 @@ cdef extern from "syclinterface/dpctl_sycl_queue_interface.h":
         void *Dest,
         int Val,
         size_t Count)
+    cdef DPCTLSyclEventRef DPCTLQueue_MemsetWithEvents(
+        const DPCTLSyclQueueRef Q,
+        void *Dest,
+        int Val,
```
ndgrigorian: here and `DPCTLQueue_Memeset` declare` int val`, when it should be `uint8_t` (that's what's declared in `dpctl_sycl_queue_interface.cpp`). Both should be corrected

dpctl/_sycl_queue.pxd
ndgrigorian: this comment was meant for `_backend.pxd`

dpctl/_sycl_queue.pyx
```diff
@@ -602,6 +604,35 @@ cdef DPCTLSyclEventRef _copy_impl(
...
+cdef DPCTLSyclEventRef _memset_impl(
+     SyclQueue q,
+     object mem,
+     int val,
+     size_t count,
+     DPCTLSyclEventRef *dep_events,
+     size_t dep_events_count,
+) except *:
+    cdef void *ptr = NULL
+    cdef DPCTLSyclEventRef ERef = NULL
```
ndgrigorian, line 617:
```suggestion
    cdef DPCTLSyclEventRef ERef = NULL
    cdef uint8_t byte_val = <uint8_t>val
```
I think we should add something like this here

ndgrigorian: then pass it into the dpctl memset calls

dpctl/_sycl_queue.pyx
vlad-perevezentsev: Done
Thank you

dpctl/tests/test_sycl_queue_memset.py
vlad-perevezentsev: It makes sense
I have added new tests, thanks

dpctl/_sycl_queue.pyx
vlad-perevezentsev: Done

libsyclinterface/include/syclinterface/dpctl_sycl_queue_interface.h
```diff
@@ -481,6 +481,29 @@ DPCTLQueue_Memset(__dpctl_keep const DPCTLSyclQueueRef QRef,
                   uint8_t Value,
                   size_t Count);
 
+/*!
+ * @brief C-API wrapper for ``sycl::queue::memset``.
+ *
+ * @param    QRef           An opaque pointer to the ``sycl::queue``.
+ * @param    USMRef         An USM pointer to the memory to fill.
+ * @param    Value          A value to fill, interpreted as an unsigned char.
+ * @param    Count          A number of uint8_t elements to fill.
```
antonwolfy:
```suggestion
 * @param    Count          A number of uint8_t bytes to fill.
```

libsyclinterface/include/syclinterface/dpctl_sycl_queue_interface.h
```diff
@@ -468,10 +468,10 @@ __dpctl_give DPCTLSyclEventRef DPCTLQueue_SubmitBarrierForEvents(
  *
  * @param    QRef           An opaque pointer to the ``sycl::queue``.
  * @param    USMRef         An USM pointer to the memory to fill.
- * @param    Value          A value to fill.
+ * @param    Value          A value to fill, interpreted as an unsigned char.
  * @param    Count          A number of uint8_t elements to fill.
```
antonwolfy:
```suggestion
 * @param    Count          A number of uint8_t bytes to fill.
```

libsyclinterface/include/syclinterface/dpctl_sycl_queue_interface.h
antonwolfy: `memset`'s third arg is a byte count; the wording was copy-pasted from the `fill` docs.

dpctl/tests/test_sycl_queue_memset.py
```diff
@@ -0,0 +1,207 @@
...
+]
+
+
+def _read_back(q, mobj, nbytes):
+    """Copy USM memory to host and return it as bytes (works for device)."""
+    host = bytearray(nbytes)
+    q.memcpy(host, mobj, nbytes)
+    return bytes(host)
+
+
```
antonwolfy, line 43: Every test allocates raw USM and checks a flat byte pattern. There's no test that views the buffer as e.g. int32, fills with byte `0xFF`, and asserts each element is `-1`/`0xFFFFFFFF` — the canonical check that byte-fill semantics are right.

dpctl/tests/test_sycl_queue_memset.py
```diff
@@ -0,0 +1,207 @@
...
+    dpctl.memory.MemoryUSMDevice,
+]
+
+
+def _read_back(q, mobj, nbytes):
+    """Copy USM memory to host and return it as bytes (works for device)."""
+    host = bytearray(nbytes)
+    q.memcpy(host, mobj, nbytes)
+    return bytes(host)
+
```
antonwolfy, line 42: The docstring promises `val` is interpreted as `unsigned char` (LSB only), but nothing verifies e.g. `memset(mem, 0x1FF)` fills `0xFF` or `-1` fills `0xFF`. There is no error case to test here (values truncate, not reject).

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -899,10 +899,46 @@ DPCTLQueue_Memset(__dpctl_keep const DPCTLSyclQueueRef QRef,
         return wrap<event>(new event(std::move(ev)));
     }
     else {
-        error_handler("QRef or USMRef passed to fill8 were NULL.", __FILE__,
+        error_handler("QRef or USMRef passed to memset were NULL.", __FILE__,
+                      __func__, __LINE__);
+        return nullptr;
+    }
+};
```
antonwolfy:
```suggestion
}
```

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -899,10 +899,46 @@ DPCTLQueue_Memset(__dpctl_keep const DPCTLSyclQueueRef QRef,
...
+        }
+    }
+    else {
+        error_handler("QRef or USMRef passed to memset were NULL.", __FILE__,
                       __func__, __LINE__);
         return nullptr;
     }
+
+    return wrap<event>(new event(ev));
 };
```
antonwolfy:
```suggestion
}
```

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -899,10 +899,46 @@ DPCTLQueue_Memset(__dpctl_keep const DPCTLSyclQueueRef QRef,
...
+
+                cgh.memset(USMRef, static_cast<int>(Value), Count);
+            });
+        } catch (const std::exception &ex) {
+            error_handler(ex, __FILE__, __func__, __LINE__);
+            return nullptr;
+        }
+    }
+    else {
+        error_handler("QRef or USMRef passed to memset were NULL.", __FILE__,
```
antonwolfy:
```suggestion
        error_handler("QRef or USMRef passed to memset_async were NULL.", __FILE__,
```

dpctl/_sycl_queue.pyx
```diff
@@ -1594,6 +1626,112 @@ cdef class SyclQueue(_SyclQueue):
...
+                raise MemoryError()
+            try:
+                for idx, de in enumerate(dEvents):
+                    if isinstance(de, SyclEvent):
+                        depEvents[idx] = (<SyclEvent>de).get_event_ref()
+                    else:
+                        raise TypeError(
+                            "A sequence of dpctl.SyclEvent is expected"
+                        )
+                ERef = _memset_impl(self, mem, byte_val, count, depEvents, nDE)
```
antonwolfy:
```suggestion
                ERef = _memset_impl(<SyclQueue>self, mem, byte_val, count, depEvents, nDE)
```

dpctl/_sycl_queue.pyx
```diff
@@ -1594,6 +1626,112 @@ cdef class SyclQueue(_SyclQueue):
...
+        ERef = _memset_impl(<SyclQueue>self, mem, byte_val, count, NULL, 0)
+        if (ERef is NULL):
+            raise RuntimeError(
+                "SyclQueue.memset operation encountered an error"
+            )
+        with nogil:
+            DPCTLEvent_Wait(ERef)
+        DPCTLEvent_Delete(ERef)
+
+    cpdef SyclEvent memset_async(
```
antonwolfy, line 1671: `memset_async` submits the fill and returns immediately with a `SyclEvent`, but the returned event holds no Python reference to `mem`.
If the caller lets `mem` go out of scope / drops its last reference before waiting on the event, the USM allocation can be freed by the garbage collector while the device is still writing to it — a use-after-free on device memory. The fix on the user's side is simply to keep `mem` alive (and/or wait() on the event) before releasing it.

It'd good to add a not to the docstrings, highlighting that hazard for the user.

dpctl/tests/test_sycl_queue_memset.py
vlad-perevezentsev: I have added an additional test for that case 
Thanks

dpctl/tests/test_sycl_queue_memset.py
vlad-perevezentsev: This case is already covered by `test_memset_value_truncated_to_byte` (negatives and overflow)
https://github.com/IntelPython/dpctl/pull/2361/changes#diff-8cf72656d339e0f446c2227869696dee611bdce0bc37434f62419482dc4cd075R192

dpctl/_sycl_queue.pyx
vlad-perevezentsev: It makes sense
I have added a Note 
Thanks

dpctl/tests/test_sycl_queue_memset.py
```diff
@@ -0,0 +1,223 @@
...
+def test_memset_async_with_dependent_events():
+    try:
+        q = dpctl.SyclQueue()
+    except dpctl.SyclQueueCreationError:
+        pytest.skip("Default constructor for SyclQueue failed")
+    nbytes = 64
+    mobj = _create_memory(q, nbytes)
+
+    e1 = q.memset_async(mobj, 0x01)
+    e2 = q.memset_async(mobj, 0x02, nbytes, [e1])
```
antonwolfy: `e2` overwrites the entire buffer with `0x02`, the final state is `0x02` no matter what `e1` did or whether `e2` actually waited for it.
To actually prove the dependency you'd need the two ops to interact where order changes the value — e.g. `e1` fills the whole buffer with `0x01`, `e2` (depending on `e1`) fills only the first half with `0x02`, then assert first half `0x02` + second half `0x01`.

dpctl/tests/test_sycl_queue_memset.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@pytest.mark.parametrize("mem_cls", _MEMORY_CLASSES)
+def test_memset_fills_whole_allocation(mem_cls):
+    try:
+        q = dpctl.SyclQueue()
+    except dpctl.SyclQueueCreationError:
+        pytest.skip("Default constructor for SyclQueue failed")
+    nbytes = 256
+    mobj = mem_cls(nbytes, queue=q)
```
antonwolfy, line 63: Not every SYCL device supports every USM kind. Needs to check the aspect (or catch the allocation failure) and pytest.skip when unsupported.
Applicable to the memcpy tests also.

dpctl/_sycl_queue.pyx
```diff
@@ -1594,6 +1626,119 @@ cdef class SyclQueue(_SyclQueue):
...
+            val (int):
+                Value to fill ``mem`` with. Following ``sycl::queue::memset``,
+                it is interpreted as an ``unsigned char``, i.e. only the least
+                significant byte is used.
+            count (int, optional):
+                Number of bytes to fill. If ``0`` or greater than the size of
+                ``mem``, the whole allocation is filled. Default: ``0``.
+
+        Raises:
+            TypeError:
```
antonwolfy, line 1651: Might raise OverflowError also:
```python
import dpctl
import dpctl.memory as dpm

q = dpctl.SyclQueue()
m = dpm.MemoryUSMShared(16, queue=q)

# val too big for a C int (typically 32-bit here):
q.memset(m, 2**40)
# OverflowError: Python int too large to convert to C long

# negative count into an unsigned size_t:
q.memset(m, 0xAB, -1)
# OverflowError: can't convert negative value to size_t
```

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -899,11 +899,47 @@ DPCTLQueue_Memset(__dpctl_keep const DPCTLSyclQueueRef QRef,
...
+            return nullptr;
+        }
+    }
+    else {
+        error_handler("QRef or USMRef passed to memset_async were NULL.",
+                      __FILE__, __func__, __LINE__);
+        return nullptr;
+    }
+
+    return wrap<event>(new event(ev));
```
antonwolfy:
```suggestion
    return wrap<event>(new event(std::move(ev)));
```

vlad-perevezentsev: Good catch
Fixed

dpctl/_sycl_queue.pyx
vlad-perevezentsev: I have added it, thanks

dpctl/tests/test_sycl_queue_memset.py
vlad-perevezentsev: Fixed, thanks

dpctl/tests/test_sycl_queue_memset.py
vlad-perevezentsev: I have added a helper function `_skip_if_usm_unsupported` that will skip the test in such a case
Thanks

ndgrigorian (approved):
LGTM, thank you @vlad-perevezentsev

PR #2362, opened 2026-08-13
[MAINT] Use `ubuntu-latest` in conda-package workflow

ndgrigorian:
This PR proposes moving conda-package workflow to use Ubuntu 24.04 via `ubuntu-latest`

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -186,11 +186,11 @@ jobs:
           - python: '3.14'
             python_spec: '3.14.* *_cp314'
             experimental: false
-            runner: ubuntu-22.04
+            runner: ubuntu-latest
           - python: '3.14'
             python_spec: '3.14.* *_cp314t'
             experimental: false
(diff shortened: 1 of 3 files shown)


.github/workflows/conda-package.yml
```diff
@@ -186,11 +186,11 @@ jobs:
           - python: '3.14'
             python_spec: '3.14.* *_cp314'
             experimental: false
-            runner: ubuntu-22.04
+            runner: ubuntu-24.04
```
vlad-perevezentsev: Would not it be better to use `ubuntu-latest` here instead of pinning `ubuntu-24.04`?
All other jobs here (and in the other workflows) use ubuntu-latest and it already resolves to 24.04

ndgrigorian: It's a fair point, we could do that instead

vlad-perevezentsev (approved):
LGTM
Thank you @ndgrigorian

PR #2365, opened 2026-08-18
Add `dpctl.SyclQueue.fill()` method

vlad-perevezentsev:
This PR proposes adding `dpctl.SyclQueue.fill` and `dpctl.SyclQueue.fill_async` methods as a Python wrapper over `sycl::queue::fill` backed by the existing `DPCTLQueue_Fill8/16/32/64/128` and new `DPCTLQueue_Fill8/16/32/64/128WithEvents` C-API functions and adding a new `test_sycl_queue_fill.py`

### DIFF
--- a/dpctl/_sycl_queue.pyx
+++ b/dpctl/_sycl_queue.pyx
@@ -40,6 +40,11 @@ from ._backend cimport (  # noqa: E211
     DPCTLQueue_CopyDataWithEvents,
     DPCTLQueue_Create,
     DPCTLQueue_Delete,
+    DPCTLQueue_Fill8,
+    DPCTLQueue_Fill16,
+    DPCTLQueue_Fill32,
+    DPCTLQueue_Fill64,
+    DPCTLQueue_Fill128,
     DPCTLQueue_GetBackend,
     DPCTLQueue_GetContext,
     DPCTLQueue_GetDevice,
@@ -86,10 +91,13 @@ from cpython.buffer cimport (
     PyObject_GetBuffer,
 )
 from cpython.ref cimport Py_INCREF, PyObject
+from libc.stdint cimport uint8_t, uint16_t, uint32_t, uint64_t
 from libc.stdlib cimport free, malloc
 
 import collections.abc
 import logging
+import struct
+import sys
 
 
 cdef extern from "_host_task_util.hpp":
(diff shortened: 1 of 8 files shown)


dpctl/_sycl_queue.pyx
```diff
@@ -571,6 +583,127 @@ cdef DPCTLSyclEventRef _copy_memcpy_impl(
...
+        raise ValueError(
+            f"Value {value!r} cannot be represented as dtype '{dtype}': {e}"
+        )
+
+
+cdef DPCTLSyclEventRef _fill_impl(
+     SyclQueue q,
+     object dst,
+     object value,
+     size_t count,
```
ndgrigorian, line 634: We need a check for OOB count, I think simply raising would be fine

dpctl/_sycl_queue.pyx
```diff
@@ -571,6 +583,127 @@ cdef DPCTLSyclEventRef _copy_memcpy_impl(
...
+            "i1, u1, i2, u2, i4, u4, i8, u8, f4, f8, c8, c16"
+        )
+    fmt = _fill_dtype_formats[dtype]
+    try:
+        if dtype == "c8" or dtype == "c16":
+            # complex: pack real then imag
+            cval = complex(value)
+            return struct.pack("=" + fmt, cval.real, cval.imag)
+        return struct.pack("=" + fmt, value)
+    except (struct.error, TypeError, ValueError) as e:
```
ndgrigorian: `struct.pack("=f", 1e40)` raises an `OverflowError`, not covered here

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
```diff
@@ -1063,3 +1203,41 @@ DPCTLQueue_Fill128(__dpctl_keep const DPCTLSyclQueueRef QRef,
...
+                dep_events.reserve(DepEventsCount);
+                for (size_t i = 0; i < DepEventsCount; ++i) {
+                    event *ei = unwrap<event>(DepEvents[i]);
+                    if (ei)
+                        dep_events.push_back(*ei);
+                }
+            }
+            complexNumber Val;
+            Val.real = Value[0];
+            Val.imag = Value[1];
```
ndgrigorian, line 1230: we need a check here and in the original `DPCTLQueue_Fill128` for nullptr `Value` here, as the try block won't prevent the segfault from a nullptr deref

libsyclinterface/tests/test_sycl_queue_interface.cpp
```diff
@@ -795,6 +815,210 @@ TEST_P(TestDPCTLQueueMemberFunctions, CheckFill128)
     delete[] host_arr;
 }
 
+TEST_P(TestDPCTLQueueMemberFunctions, CheckFill8WithEvents)
```
ndgrigorian, line 822: need test where `Value` is nullptr for Fill128 variants

libsyclinterface/source/dpctl_sycl_queue_interface.cpp
vlad-perevezentsev: Good catch
Updated, thanks

libsyclinterface/tests/test_sycl_queue_interface.cpp
vlad-perevezentsev: Done, thanks

dpctl/_sycl_queue.pyx
vlad-perevezentsev: Done in https://github.com/IntelPython/dpctl/pull/2365/commits/10929556c7acb039f82537266005820ae4d473ea
thanks

ndgrigorian (approved):
LGTM

PR #2367, opened 2026-08-19
[MAINT] Add conda-recipe for conda-forge only dpctl builds

ndgrigorian:
This PR proposes adding a workflow for conda-forge builds of the project, which test that the Conda packages for the compiler dependencies (and NumPy) brought in purely from conda-forge work with dpctl


conda-recipe-cf/build.sh
```diff
@@ -0,0 +1,38 @@
...
+export CMAKE_GENERATOR=Ninja
+# Make CMake verbose
+export VERBOSE=1
+
+# set CMAKE to use less threads to avoid OOM
+export CMAKE_BUILD_PARALLEL_LEVEL=${CPU_COUNT}
+
+CMAKE_ARGS="${CMAKE_ARGS} -DDPCTL_LEVEL_ZERO_INCLUDE_DIR=${PREFIX}/include/level_zero -DDPCTL_WITH_REDIST=ON"
+
+# -wnx flags mean: --wheel --no-isolation --skip-dependency-check
```
antonwolfy: It's probably a follow-up question, but why do we need to enable wheels?

conda-recipe-cf/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+source:
+  path: ..
+
+build:
+  number: {{ GIT_DESCRIBE_NUMBER }}
+  # let the public CI build without IPO, which needs more resources than the
+  # GitHub Actions runners provide
+  script_env:
+    - OVERRIDE_INTEL_IPO  # [win]
```
antonwolfy: We are not using that in the feedstock. And I believe we probably no need that for dpctl build anymore, since it's lightweighted now.

conda-recipe-cf/bld.bat
```diff
@@ -0,0 +1,46 @@
...
+:: wheel file was renamed
+for /f %%f in ('dir /b /S .\dist') do (
+    %PYTHON% -m pip install %%f ^
+      --no-build-isolation ^
+      --no-deps ^
+      --only-binary :all: ^
+      --no-index ^
+      --prefix %PREFIX% ^
+      -vv
+    if %ERRORLEVEL% neq 0 exit 1
```
antonwolfy: Here the check is effectively dead — it always sees the pre-loop errorlevel. To actually catch a failing `pip install` inside the block we need:
```suggestion
    if errorlevel 1 exit 1
```
Probably needs to be backported to the feedstock also.

antonwolfy: While at top level, both work. `if %ERRORLEVEL% neq 0` is actually more precise than `if errorlevel 1` because the latter is a `>=` test and also matches unusual negative codes differently. This is why the `run_test.bat` scripts (all top-level statements) are fine as written.

conda-recipe-cf/conda_build_config.yaml
```diff
@@ -0,0 +1,18 @@
+numpy:
+    - '2'
+c_compiler:                   # [linux]
+    - gcc                     # [linux]
+cxx_compiler:                 # [linux]
+    - gxx                     # [linux]
+cxx_compiler_version:         # [linux]
+    - '14'                    # [linux]
```
antonwolfy: to match the feedstock
```suggestion
    - '15'                    # [linux]
```

conda-recipe-cf/bld.bat
ndgrigorian: in that case I will consider it resolved

conda-recipe-cf/meta.yaml
ndgrigorian: should start running tests in the feedstock too...

conda-recipe-cf/bld.bat
antonwolfy: As I got, we should use `if errorlevel 1 exit 1` inside the loop as above

conda-recipe-cf/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+  name: dpctl
+  version: {{ GIT_DESCRIBE_TAG }}
+
+source:
+  path: ..
+
+build:
+  number: {{ GIT_DESCRIBE_NUMBER }}
+  # let the public CI build without IPO, which needs more resources than the
+  # GitHub Actions runners provide
```
antonwolfy: The comment is obsolete now
```suggestion
```

antonwolfy (dismissed):
Thank you @ndgrigorian

.github/workflows/conda-package-cf.yml
```diff
@@ -0,0 +1,372 @@
...
+          echo "CONDA_BLD=$CONDA/conda-bld/win-64/" | tr "\\\\" '/' >> "$GITHUB_ENV"
+      - name: Show Conda info
+        run: |
+          conda info --all
+      - name: List base environment packages
+        run: |
+          conda list -n base
+      - name: Build conda package
+        env:
+          OVERRIDE_INTEL_IPO: 1   # IPO requires more resources that GH actions VM provides
```
antonwolfy: Can be dropped

conda-recipe-cf/bld.bat
```diff
@@ -0,0 +1,46 @@
+REM A workaround for activate-dpcpp.bat issue to be addressed in 2021.4
+set "LIB=%BUILD_PREFIX%\Library\lib;%BUILD_PREFIX%\compiler\lib;%LIB%"
+set "INCLUDE=%BUILD_PREFIX%\include;%INCLUDE%"
+
+"%PYTHON%" setup.py clean --all
+
+REM useful for building in resources constrained VMs (public CI)
+set "CMAKE_ARGS=%CMAKE_ARGS% -DCMAKE_INTERPROCEDURAL_OPTIMIZATION:BOOL=FALSE"
```
antonwolfy, line 8: And I believe that can be dropped also, but we can do both in the follow-up

antonwolfy (approved):

PR #2374, opened 2026-09-02
ci: harden GitHub Actions workflows and add zizmor scan

antonwolfy:
Adds a CI job that runs the [zizmor](https://docs.zizmor.sh) static analyzer over the workflow files under `.github/`.

`zizmor` audits GitHub Actions workflows for supply-chain and privilege-escalation weaknesses — unpinned action references, credential persistence through the checkout token, template injection via `${{ ... }}` expansion in `run:` blocks, and overly broad `GITHUB_TOKEN` permissions.

Alongside the new scan, this applies the corresponding hardening to the existing workflows so they pass the audit:

- add `persist-credentials: false` to every `actions/checkout` step that lacked it
- narrow top-level `permissions: read-all` to `permissions: contents: read`
- move `github.event.inputs.tag` into an `env:` var in `backfill-docs.yml` to avoid template injection in a `run:` block
- fix a malformed version-pin comment for `mshick/add-pr-comment` (`# v3.12.0.8.3.12.0` -> `# v3.12.0`)
- add a 7-day `cooldown` to the dependabot update entries

This is a CI/configuration-only change; no library code, tests, or documentation are affected.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,8 @@ updates:
     schedule:
       interval: "weekly"
       day: "saturday"
+    cooldown:
+      default-days: 7
     labels: ["autoupdate"]
     groups:
       github-actions:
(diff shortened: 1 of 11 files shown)


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

